PR #92, opened 2026-08-31
Add effort selector to model picker

Sniezka1927:
## Summary
- add a model-specific effort selector to the conversation header model popup
- keep model and effort changes in the popup while persisting them immediately
- optimistically update provider/model/effort state and roll back failed saves
- preserve the existing Agent Settings reasoning control

## Verification
- bun run test -- src/renderer/src/components/ProviderModelPicker.test.tsx src/renderer/src/App.test.tsx
- bun run typecheck
- bun run lint


Sniezka1927:
Screenshot 2026-08-31 at 15 57 19

PR #128, opened 2026-09-01
Browser Automation V2: native CDP control

NorbertBodziony:
## Main merge — `75ab981f`

Merged `main` at `fa0e9086` without conflicts. Checked the shared provider, agent-service, renderer stylesheet, and preview changes. No extra source changes were needed.

Local verification passed: 66 tests across provider, queue, Claude client, and browser upload suites; the full Electron browser suite including persistence; lint (existing mobile test warnings), typecheck, and UI checks. CI is checked on the pushed merge commit. GPT Luna resumed the existing local test agent and passed open, mobile-to-fill reset (mode fill, scale 1), double-click activation, and close after the merge.

## Latest review fixes — `295af1e5`, `4b04dc74`, `9c193550`

- Keep assigned upload copies across document navigation, since a surviving parent can hold the earlier File objects. Unassigned reservations still expire with their source document. Existing input and byte quotas apply; tab close and service stop remove assigned copies. Both existing retention tests failed with ENOENT before the fix and passed after it.
- Dispatch both press/release pairs for double-click, with click counts 1 and 2. The new case requires the first trusted click to select an item before double-click activation. It failed before the fix and passed after it. The fixture has a separate tab.
- Removed a timing-dependent width assertion from the fill-reset test after CI returned the previous width during resize. Reset success, fill mode, and scale 1 remain required.
- Reset the custom scale when returning from mobile to fill mode. The new Electron case failed on the old code with the physical viewport limit error, then passed with the fix. The full Electron suite passed; one earlier run failed the existing unrelated window-focus check.
- Bound `set_environment`, including engine rollback, through the existing timeout and queue drain. A stopped main renderer now returns a timeout and permits the tab to close.
- Removed the observer counter installed in the page world. It could not measure observers in the automation world. The DOM-quiet timeout and recovery checks remain.
- Added a stopped-main-renderer case to the existing Electron suite. The full `bun run test:browser` passed. Removing the bound made the case fail with `V2 environment change did not bound an unresponsive renderer: null`; the bound was then restored.
- Local checks passed: seven upload tests, lint (eight existing mobile test warnings), typecheck, and UI checks. CI and automated review are checked on the pushed commit.
- Changed surfaces: desktop browser backend and Electron test harness. No renderer, mobile, public web, Signal service, contract, or database changes in this follow-up.
- Live model check: GPT Luna through Codex CLI 0.146.0, using isolated `OpenBot Dev 5348`. Open, mobile/dark, desktop/system, mobile-to-fill, and close all succeeded. Fill returned mode `fill` and deviceScaleFactor `1` after mobile returned `3`. Luna also stopped a test page, received `Browser environment change timed out.`, then closed that tab successfully.
- Final live checks with Luna: double-click selected and activated the fixture; a retained upload was readable after iframe navigation through `new Blob([window.savedFile]).text()`, returning `shared upload remains readable`; tab close succeeded. Calling the removed iframe realm's original `File.text()` timed out even with the backing file present, so the parent-realm read was used to verify the retained bytes.
- Final CI: all checks passed on `9c193550`; NorbiAI reported no actionable findings.
- Follow-up model and harness: GPT-6 in Codex desktop on macOS, Bun 1.4.0.

## Summary

Rebuilds embedded browser automation around native Electron/CDP:

- Semantic accessibility-tree snapshots with revision-bound refs, trusted CDP input, event-driven settling, iframe/OOPIF and shadow-DOM piercing.
- 23 provider-neutral tools for Codex, Claude and Grok, replacing the previous 7, with Team API v1 and legacy `act` compatibility preserved.
- Adaptive snapshots, page diagnostics, environment emulation, page evaluation, file upload, drag, recording, and a richer browser activity UI.
- Persistent sessions, ownership isolation, downloads, takeover flows and the v1 → v2 state migration all kept.

## This update: merged `main` and ported onto the current architecture

`main` had moved 69 commits since the merge base and the branch was `CONFLICTING`. The moves were
the ones that break this branch, so this was a port rather than a conflict resolution:

| What moved on `main` | Where the PR's change went |
| --- | --- |
| `AgentService` split into a facade + 21 controllers (`agent-service.ts` 6056 → 1244 lines) | The +306 lines of upload staging became a new controller, `src/backend/agent/browser-uploads.ts`, owning `mkdtemp`/`0o700` staging, per-tab and global byte/input quotas, and cleanup on tab close, document change and service stop |
| Takeover moved into `src/backend/agent/attention-registry.ts` | The takeover lifecycle re-homed there, behind a narrow `AttentionBrowserHost` interface (`listTabs`/`beginTakeover`/`endTakeover`) rather than the whole host |
| `bot` → `agent` rename (#239) | `ownerBotId` → `ownerAgentId` throughout |
| `console.*` → `@openbot/logging` | The one remaining `console.*` in `browser-host.ts` now uses `createOpenBotLogger`/`toLogValue` |
| Renderer flattened to one dir per domain (#241) | Panel edits applied to `features/conversation/BrowserPanel.tsx`, CSS into `features/conversation/conversation.css`, import depths fixed |
| `agent-service.test.ts` deleted (#167) into per-controller suites | The 385-line monolithic test case became focused cases in `src/backend/agent/browser-uploads.test.ts` |
| `browser-state.ts` grew `reownStoredBrowserTab`/`storedBrowserTab`/`isPersistableBrowserUrl` | The hand-rolled `isStoredBrowserTab` in `browser-host.ts` was dropped; v2 `environment` validation lives in `browser-state.ts` beside its siblings |

The dispatch split by reason-to-change: `BrowserUploads` owns staging and quotas, `AttentionRegistry`
owns the takeover lifecycle, and the facade only dispatches.

`main` has been merged again since, for the eight commits up to `21b07b43`. That merge was clean, but
the retheme in `#287` collapsed the type scale to six steps and dropped the 10px
`--openbot-text-2xs` the three new browser toolbar chips were sized with, which
`scripts/design-tokens.test.ts` catches from both the desktop renderer and the public web surface.
The chips moved to `--openbot-text-sm`, the smallest step the scale now has.

## Review findings fixed

1. **`evaluate` described itself as "sandboxed" while running in the page's main world.**
   `Runtime.evaluate` is called with no `contextId`, unlike every internal helper, which goes through
   `Page.createIsolatedWorld`. Isolating it would not change what is reachable — an isolated world
   shares the DOM and `document.cookie` — so the capability is the normal one for a browser
   automation tool, but the description shipped to all three providers was wrong. It now reads
   "the main frame's own page context, with the same access to page scripts, DOM and cookies as the
   page itself."
2. **`upload_files` widened the attachment sandbox through an anonymous positional boolean.** It is
   now a named `AttachmentSourceScope { allowAnyReadablePath }` option on `attachment-gateway.ts`.
   See the security note below.
3. **`request_takeover` was published to all three providers with no `BrowserHost` handler** — it
   only worked because the facade intercepts the namespace first, and any other caller hit
   `Unknown browser tool`. It now has an explicit case returning a clear error naming the agent
   service as its owner.
4. **`evaluate` returned `Promise<unknown>`** to its caller despite having already proven the value
   JSON-serializable and under 64 KB. It now returns a named `BrowserJsonValue` from
   `packages/contracts`.
5. **Console diagnostics were bounded but never redacted.** A page logging
   `Authorization: Bearer ...` put that token straight into the snapshot that `textResult` hands to
   the provider. Redaction now happens in `BrowserDiagnostics` on the way *into* the ring rather
   than at each reader, so the ring never holds the secret: that covers the renderer's error count
   and the Team API projection as well as the tool result. `target` and `detail` on the action ring
   got the same treatment — a target description quotes the element's accessible name and an error
   detail quotes the page's own exception message. Found by the automated review, fixed in
   `3631e8c6`.
6. **`evaluate` could deadlock the tab's queue.** It defaults to `awaitPromise: true`, and CDP's
   execution timeout does not bound waiting for a promise the page never settles. The host returned
   its own timeout but `drained` still awaited the pending command, so the queue never advanced and
   takeover, close and the shutdown drain all blocked behind it. `#runEvaluation` now cancels the
   pending command on timeout and skips the drain once cancellation is confirmed, which is the
   pattern `#runAction` already used. Found by the automated review, fixed in `3631e8c6`.
7. **The diagnostic chip could not count.** The chip renders whenever the count exceeds zero, so a
   tab with exactly one console error read "1 browser diagnostic errors" in both its tooltip and its
   accessible name. Found while driving a real agent at a fixture page that logs one error — see
   Level D. Fixed in `62297cda`, with the singular case added as a story.
8. **A page exception carried the page's own text to the provider unredacted.** Every browser tool
   failure reaches a provider through one `catch` in `handleDynamicTool`, and a page that throws
   `new Error("password=hunter2")` put that string in the tool result — the same class of leak as
   fix 5, on the path fix 5 does not cover. That one `catch` now applies the same `redactText`,
   which covers all 23 tools at once. Found by the automated review, fixed in `ad35c80e`.
9. **A takeover starting during upload staging could still write files into the page.** Staging runs
   outside the tab queue and takes as long as the files are large, so the facade's pre-flight
   takeover check can be stale by the time the input is assigned — and `BrowserHost` checks tab
   ownership, not takeover, so nothing downstream caught it. `BrowserUploads` now reads takeover
   again inside the queued operation at `onUploadTargetResolved`, the last point before assignment,
   and invalidates the reservation so the staged copies are removed rather than left for the next
   document change. Found by the automated review, fixed in `ad35c80e`.
10. **`type` with `submit: true` inserted the text and never submitted it.** Enter was pressed as a
    second action against a re-resolved target, and typing changes a contenteditable's visible text,
    so the ref fingerprinted the element against its *pre-typing* text and threw "The target changed
    after the snapshot". Submitting now happens inside the lease that already resolved the node,
    behind an options object matching `click()`'s existing idiom rather than a fourth positional
    boolean. Found by the automated review, fixed in `ad35c80e`.

11. **An upload error carried a page-controlled secret to the provider.** Fix 8 redacts what
    `handleDynamicTool` *returns*, but everything `BrowserUploads` does around that call — resolving
    the target, opening the sources, the quotas, staging — throws past it to the facade, which
    forwards `String(error)` to the provider unchanged. An ambiguous semantic target names its
    candidates by accessible name, so two inputs labelled `Upload password=hunter2` put the password
    in the error. `uploadFiles` now wraps its whole body and rethrows a redacted `Error`, which keeps
    browser redaction inside the browser subsystem and covers target resolution, source opening,
    quotas and staging in one place. Found by the automated review, fixed in `f39a7cbe`.
12. **`press("Enter")` never submitted and `press("Space")` inserted nothing.** `dispatchShortcut`
    sent `rawKeyDown` and `keyUp` and gated the `char` event on `key.length === 1`, which no named
    key satisfies. Chromium decides implicit form submission and text insertion from the character
    event, not the key event, so a plain `<form>` with no script never submitted and a textarea never
    gained a line. `normalizeKey` now carries the character a key produces and `dispatchShortcut`
    dispatches it as `char`. `Tab` stays characterless on purpose — the browser moves focus on the
    key event, and a character dispatched afterwards would land in whatever gained focus — and a
    modified shortcut gets none either, because `Ctrl+S` is a command rather than an `s` in the
    document. Found by the automated review, fixed in `f39a7cbe`.
13. **Document enumeration stopped at a shadow boundary and freed live uploads.**
    `documentIdsExpression` was the one traversal in `browser-cdp.ts` still using `querySelectorAll`,
    which does not pierce shadow roots, so an upload assigned to an input in an iframe nested inside
    a shadow root looked closed to the caller — and a document that looks closed frees the staged
    files the still-open input is holding. It now uses the same `createTreeWalker` walk that target
    discovery uses, because it has to reach the same documents. Found by the automated review, fixed
    in `f39a7cbe`.

14. **A truncated document scan was reported as complete, and freed live uploads.** The tree walk
    fix 13 introduced stops at the 10,000-node budget — a cap the `querySelectorAll` it replaced never
    had — so an upload frame past that point was never visited, and a document missing from the list
    is indistinguishable from a closed one. `retainDocuments` then deleted the staged files of an
    input that is still open. The expression now reports completeness, and an unvisited document is
    presumed open: a staging directory kept until the tab closes costs a temp directory, while freeing
    one whose input is live hands the page a path that no longer exists. Found by the automated review
    — a fair catch against my own fix 13 — fixed in `06585ad3`.
15. **`select_option` could not clear an optional selection.** The schema and the engine both accept
    `values: [""]`, but `requiredStringArray` rejected every empty string before the engine ran, and
    `<option value="">` is how a page spells "no selection". The empty check is now explicit per list:
    allowed for option values, still rejected for upload paths, where an empty path is never a file.
    Found by the automated review, fixed in `06585ad3`.
16. **Two `select_option` calls on the same select within a second left the value unchanged.**
    Uncovered while testing fix 15, not by review. Chromium keeps a typeahead buffer per select for
    about a second, so the second call appends to the characters the first one typed: selecting `b`
    and then an option labelled `Any mode` searched for `ba`, matched nothing, and silently left the
    selection where it was. No phase had covered it because no phase selected twice. A focus
    round-trip before typing clears the buffer; waiting the timer out is the only alternative and
    costs a second on every call. I measured all three paths before choosing — an immediate retry
    fails, focusing another element first succeeds, a 1400 ms wait succeeds. Fixed in `06585ad3`.

    The engine's non-typeahead fallback (`Home` then `ArrowDown`) turns out to be inert on macOS,
    because a closed select opens its native popup on `ArrowDown` rather than moving the selection,
    and no CDP key event reaches that popup. Typeahead is therefore the only channel on this platform,
    which means an option carrying **no label at all** cannot be selected here. That is a platform
    limit rather than something to fix, so it is now named in the error instead of being reported as
    a bare index mismatch. It is left as-is on Linux and Windows, where the fallback works.

17. **A snapshot the page never answered blocked the tab forever.** `#readSnapshot` registered the
    snapshot's completion as a queue drain, and Electron's `sendCommand` carries no timeout of its
    own, so a frame whose renderer never replies leaves that promise pending for good. The caller
    still got its timeout error, which is what made this invisible: everything else on the tab —
    navigation, takeover, tab destruction, host shutdown — queued behind a drain that would never
    settle. Snapshots now cancel the way evaluations already did (fix 6), and release the drain once
    the detach is confirmed; when the recorder holds the debugger there is nothing to detach and the
    wait correctly stands. Found by the automated review, fixed in `05559272`.
18. **`Shift+Enter` inserted nothing and reported success.** The character event was suppressed for
    *every* modifier, on the reasoning that `Ctrl+S` is a command rather than an `s` in the document.
    Shift is not a command modifier — `Shift+Enter` is how a composer spells "line break, do not
    submit" — so the shortcut reached the page as a keydown that changed nothing while the tool
    returned success. Shift-only combinations now carry their character, and `event.key` agrees with
    it (`Shift+a` reports `A`, not an `a` with a flag beside it). A shifted glyph nobody can name from
    the layout is deliberately left alone: `Shift+1` is `!` on a US keyboard and something else on
    half a dozen others, so it stays a key event rather than a guessed character. Found by the
    automated review, fixed in `05559272`.
19. **The staged-file size check could not see a file that grew.** The copy reads to
    `expectedBytes - 1`, the size measured before staging, so a source that grew underneath it
    produced a staged file of exactly the expected length and the equality check passed — it could
    only ever catch a file that shrank. The descriptor the copy read from is now re-stat'd after the
    copy — the same open handle, so it is the same file even if the path was replaced — and either
    direction is rejected. Found by the automated review, fixed in `05559272`.
20. **`wait_for` and the settle after a dispatched action had the same hole as fix 17.** Both check
    their deadline *between* CDP commands, so a frame whose renderer answers none of them is never
    reached by it: `wait_for` with a `text` condition evaluates in every frame, the post-action settle
    waits for DOM quiet in every frame, and either one hangs on that frame with the tool's own timeout
    running inside it. The tab queue then never advances past the drain, which is navigation,
    takeover, close and host shutdown as well as the next tool call. Both now go through the
    wall-clock bound the snapshot already had, extracted as `#boundEngineOperation`. Found by the
    automated review, fixed in `0ad76b5c`.
21. **The character event carried no modifier mask.** CDP defaults the omitted mask to zero, so the
    `Shift+Enter` that fix 18 started sending arrived at the page as a `keypress` with
    `shiftKey === false` — and a composer that decides between "send" and "line break" in its keypress
    handler sends the message, whatever the textarea ends up containing. The `char` event now carries
    the same mask as the `keydown` and `keyUp` around it. Found by the automated review, fixed in
    `0ad76b5c`.
22. **Cancelling was too eager, and a lease detached the debugger out from under the next
    operation.** Not from review — found by fix 20 turning the suite red in two places that had
    nothing to do with the change. Detaching the debugger is the only cancellation primitive there is
    and it takes the whole session down, so cancelling the moment a deadline passed cancelled
    renderers that were about to answer: a 5 ms wait is *expected* to time out, and the existing
    phases that use one started failing with `target closed while handling command`. Underneath that
    was an older bug the eager cancel exposed. `#lease` detached whenever it was the lease that had
    attached, which is wrong as soon as two overlap — an operation past its deadline goes on running
    while the next one starts, and whichever finished first took the other's debugger with it,
    surfacing as `target closed while handling command` and then `No target available` on a page that
    was perfectly healthy. Leases are counted now and the last one out detaches; a timeout waits for
    the operation to unwind on its own before detaching anything, and waits again afterwards, so the
    queue advances into an attached debugger rather than one being torn down. Fixed in `0ad76b5c`.
23. **Three queued entry points still sent frame commands nothing bounded.** Fix 20 bounded the tool
    paths; three others queue the same unbounded frame walk on the same tab. The upload preflight is
    the one that stings: `resolveUploadTarget` proves a target unique across *every* frame, and it is
    queued ahead of the bounded upload action, so a frame that answers nothing means the tab never
    reaches the operation the timeout was meant to protect. The other two are the document
    enumeration a `did-frame-navigate` starts — which has no caller and no deadline at all — and the
    settle in the legacy `act`, which is the renderer's own path, so wedging it leaves the panel's tab
    dead for the rest of the session. All three go through `#boundEngineOperation` now. Enumeration
    also gained a distinction its failure path was missing: the empty document set it reports tells
    upload staging that the documents holding its files are gone and it deletes them, which is the
    right answer when the walk ran and found nothing and the wrong one when the frames were never
    asked — so a timeout preserves the staged files rather than freeing them. Legacy `act` still
    *fails* on a settle timeout, as it always did; only the hang is gone. Found by the automated
    review, fixed in `90fae5c7`.
24. **Bounding the legacy `act` covered its settling and not its dispatch.** Fix 23 bounded the
    settle that follows the action; the click, type, press and scroll that come first were still
    awaited directly. They carry the engine's own deadline, but that deadline is checked *between*
    CDP commands, and re-resolving a snapshot ref fingerprints the element in the frame that owns it
    — so a renderer that answers none of those commands never reaches the check, and navigation,
    takeover, close and shutdown stay blocked behind the tab queue. The dispatch now goes through
    `#boundEngineOperation` too. Found by the automated review, fixed in `98db920a`.
25. **Reassigning a file input deleted files the page could still read.** Staging deleted an input's
    previous directory the moment the input was given a new selection. But setting `input.files`
    again does not invalidate the `File` objects a page already took from it, and collecting them
    across successive selections is exactly how an "add another file" flow works — so picking a
    second attachment could break reading or uploading the first. An input now retains every
    directory it was given until the document that could read them is gone, which is the point the
    tab-close, navigation and shutdown paths already free them at. Nothing is subtracted on
    reassignment any more, so the earlier copies keep counting against the per-tab and global byte
    quotas, and those quotas are what bounds the accumulation. Found by the automated review, fixed
    in `5441ec83`.

## Security note: the upload sandbox is deliberately wider

`upload_files` skips both the symlink pre-check and the
`isWithin(workspaceRoot) || isWithin(sharedRoot)` guard, so an agent can stage any file OpenBot can
read into a browser file input. Agents already run with `danger-full-access` and could copy such a
file into the workspace first, so this is convenience rather than new capability — but it is a real
widening and it is now spelled out rather than implied: the scope is a named option, it is passed
only from `browser-uploads.ts`'s single `UPLOAD_SCOPE`, and the second `lstat(resolved)` symlink
rejection, the `O_NOFOLLOW` open and the dev/ino re-validation after resolution all still run.
Verified end-to-end in Level D: a file in `/tmp`, outside the workspace, staged successfully.

**Team API v1 is unaffected.** `projectV1Object` is a positive allowlist over
`V1_BROWSER_TAB_KEYS`, so the new optional fields (`environment`, `recording`,
`diagnosticErrorCount`, `detailAction`) are stripped rather than leaked, and no protocol version is
needed. `src/main/team-api-server.browser.test.ts` still passes.

## Report-only findings

- The recorder's `getDisplayMedia` fallback is safe only while `setDisplayMediaRequestHandler` pins
  `{video: contents.mainFrame}`.
- `BrowserSnapshot.url`/`title` are deliberately unredacted while `element.frame.url` goes through a
  second, near-duplicate helper — `diagnosticUrl` vs `redactedMetadataUrl` are worth unifying.
- `debugger.detach()` as a cancellation primitive interacts with user-opened DevTools.
- The `paths` bound is duplicated: schema `1..10` against a literal in `uploadFiles`.
- The prompt-injection surface grew (cross-origin iframe text, third-party console output) against
  one sentence of prompt mitigation.
- Pre-existing, untouched: submitting the address bar calls `browser.open`, so a new URL opens a new
  tab rather than navigating the selected one, and `focus: true` did not move selection.
  `features/conversation/stores/browser-store.ts` is unchanged by this PR.
- Worth noting as a **win**: unifying Claude's tool list onto `BROWSER_TOOL_DEFINITIONS` fixes
  existing drift. On `main`, `BROWSER_DYNAMIC_TOOLS` publishes 7 tools including `request_takeover`,
  but the hand-written Claude MCP list in `claude-client.ts` publishes only 6 and omits it — so a
  Claude agent on `main` cannot ask for takeover at all. One list now feeds all three providers,
  which is what `browser-tools.test.ts` asserts.

## Surfaces touched

Desktop renderer (`features/conversation`, `preview/`, one story file), backend
(`browser-*`, `agent/*`, `agent-service.ts`, `claude-client.ts`, the test harness), IPC contracts
(`ipc-browser.ts` — types only, no new channel, so the channel-list/manifest/mirror dance does not
apply) and its second implementation `preview/mock-openbot.ts`, persisted state
(`browser-tabs.json` v1 → v2), the Team API v1 projection (verified unchanged), the Electron smoke
script, and `docs/ARCHITECTURE.md`. **Not touched:** mobile, public web, the Signal service, SQLite
migrations.

`browser-tabs.json` is versioned in-file, sits outside `openbot.db` and the migration runner, and
nothing copies it first, so it is now documented under "State ownership" in `docs/ARCHITECTURE.md`,
including that a tab whose environment fails validation is still returned without it.

## Tests

`lint`, `typecheck` (all 11 projects) and `check:ui` are clean. The only lint warnings repo-wide are
5 pre-existing `no-module-mocking` in `apps/mobile/src/features/auth/api/mobile-auth.test.ts`. No
`biome-ignore`, `@ts-expect-error`, `any` or assertion past a checker was added.

New and changed cases:

- `src/backend/agent/browser-uploads.test.ts` (new file — the controller is a boundary that did not
  exist before): staging from any readable path, the per-tab input quota, `retainDocuments`,
  `dispose`, and reservation release.
- `src/backend/agent/attention-registry.test.ts`: the duplicate-tab refusal, and a recorder proving
  `endTakeover` is actually called. **This closed a real coverage gap** — nothing asserted the
  reverse state, and a tab the browser never leaves takeover on is a state the user enters and
  cannot exit.
- `src/backend/browser-state.test.ts`: a v1 row upgrades with `environment` undefined, a valid v2
  environment round-trips, and a 2000×2000 @ scale 4 viewport is dropped while the tab is still
  returned.
- `src/backend/browser-diagnostics.test.ts`: a bearer token, a JSON `password`, an `api_key=` in an
  action detail and an email in an action target are all redacted before the ring retains them.
  Exact text, because a security message is one of the cases where exact text is the contract.
- `scripts/browser-smoke-electron.ts`: a `new Promise(() => {})` evaluation phase ahead of the
  existing `queueRecovered` check — the deadlock is only observable with a real debugger session, so
  it belongs in the Electron suite rather than a unit test. Two more phases came with fixes 8 and
  10: a `submit: true` typed into the contenteditable *through a snapshot ref*, asserting the page's
  own `onkeydown` saw a trusted Enter, and a page exception carrying `password=hunter2`, asserting
  the provider gets `[redacted]` instead.
- `src/backend/agent/browser-uploads.test.ts`: a sixth case for fix 9 — a takeover that begins while
  files are staging. The controller is constructed directly rather than driven through the facade,
  because the race window opens after the facade's pre-flight check and closes at assignment, and
  only the controller's own boundary can be held at that instant deterministically. It also asserts
  the staged copies are gone, since a refusal that leaves them behind is a file the next document
  change would hand to a page the user is holding.
- `scripts/browser-smoke-electron.ts`: fixes 12 and 13 share one phase, on a new `/keys` page with
  a `/frame-files` route behind it — `press("Space")` into a field, then `submit: true` asserting the
  page's own `onsubmit` saw a trusted event, then an upload into an iframe nested inside a shadow
  root followed by a navigation elsewhere in the same tab, asserting the shadow-nested document is
  still enumerated. Its own page rather than more markup on `/v2`: that page's height and width are
  asserted against a 220×560 panel, and a block form plus a default-sized iframe put a scrollbar in
  it. That is how I found out — adding them to `/v2` turned the existing narrow-panel phase red.
- `src/backend/agent/browser-uploads.test.ts`: a seventh case for fix 11 — an ambiguous target whose
  message quotes a page-chosen accessible name, asserting the provider's JSON-RPC error reads
  `[redacted]` and not `hunter2`.

- `scripts/browser-smoke-electron.ts`: fixes 14-16 share the select and keys phases. The `Mode`
  fixture gained `<option value="">Any mode</option>`, selected immediately after `b` — one assertion
  covering the validation fix and the typeahead buffer at once, since clearing is exactly the
  second-selection case that used to fail — and the keys phase now grows its document past the
  10,000-node budget with `evaluate` before navigating a frame, so the upload document behind the
  shadow-root iframe is enumerated across a truncated scan.

- `scripts/browser-smoke-electron.ts`: fixes 17 and 18 share the keys phase. Fix 18 types `line` into
  a textarea and presses `Shift+Enter`, asserting the value gained a newline. Fix 17 needed a page
  that answers nothing, so a new `/blocking-frame` fixture appends a cross-origin frame that
  announces itself and then spins forever; `wait_for` with a `url` condition is the one wait that
  needs no CDP command, so the wait settles and the snapshot behind it is what hits the blocked
  frame. The assertion is that `close` on that tab returns at all. Two things the fixture had to get
  right, both found by running it: a loop entered inline never lets the announcement out, so the spin
  starts from a task; and the frame needs a site of its own (`spin.localhost`, not the `localhost`
  the other cross-origin fixtures use) because same site means the same renderer, and wedging it took
  the `/v2` page's own frame down with it for the rest of the run.
- `scripts/browser-smoke-electron.ts`: fixes 20 and 21 reuse the two fixtures fix 17 and fix 18 built.
  The keys phase now reads a `keypress` logger on the textarea and asserts `keypress:Enter:true`,
  which is the distinction the value assertion alone cannot make — a newline lands in the textarea
  either way. The `/blocking-frame` phase adds a `wait_for` with a `text` condition, which unlike the
  `url` condition of fix 17 is evaluated in every frame, so the wait itself is what hits the frame
  that answers nothing; the assertion is that the call returns at all.
- **The settle half of fix 20 has no test.** Every path that wedges a frame ends in a detach, which
  clears the auto-attached session list, so the settle that follows only reaches the main frame and
  returns on its own — a click against the blocked fixture stayed green with the bound reverted, and
  an assertion that cannot go red is a costume. The bound stays because the hazard is real (`settle`
  awaits `Promise.all` over one `Runtime.evaluate` per frame, none of them bounded), but it is
  carried by code review rather than by the suite, and this says so rather than implying coverage
  that does not exist.
- **Fix 19 has no test.** The window it closes opens between the source's `stat` and the copy, and
  nothing can hold a file open at that instant without a sleep — a test that needs a timeout to pass
  is one this repo's rules reject outright. The existing `browser-uploads.test.ts` was run to confirm
  the extra `stat` does not reject an ordinary upload.

- `scripts/browser-smoke-electron.ts`: two of fix 23's three paths reuse the `/blocking-frame`
  fixture fix 17 built, so neither needed a new page. The upload preflight is driven with a `css`
  target, which has to be proven unique in every frame, so the preflight itself is what hits the
  frame that answers nothing — ahead of the action it precedes. The legacy `act` is driven twice: a
  `scroll` for the settle, which reaches the wedged frame only after dispatching against a
  responsive main frame, and a `click` on a ref *inside* the wedged frame for the dispatch. The ref
  is what made fix 24's assertion need a fixture change: `act` accepts no target but a ref, and a
  ref only exists for an element some snapshot saw, so the spinning frame now waits to be told to
  spin and carries a button — the tab is snapshotted while the frame still answers, then wedged on
  command. That assertion runs first in the phase, because the unwind every later one triggers
  detaches the session the ref resolves through. Each assertion is that the call returns at all, and
  the phase's existing `close` race then shows the queue they were holding is free.
- **The enumeration half of fix 23 has no test, and this one is not a cost decision.** The empty set
  is only reported for a *main*-frame navigation, and a main-frame navigation replaces the document
  the wedged subframe lives in — so reaching that branch means landing on a new page whose own
  spinning frame has already attached a CDP session by the time enumeration runs, which is a race no
  fixture here can close. An assertion that goes green when the race falls the other way is worse
  than none, so the guard is carried by code review, and the existing truncated-scan phase covers the
  neighbouring case where the walk *does* run.
- The suite's per-process hard timeout went from 60 s to 120 s. It is a backstop for a phase that
  hangs, not a performance budget, and four phases now spend about ten seconds each proving that a
  frame which answers nothing is given up on; the whole suite runs in about 46 s of wall clock on
  this machine, against 35 s before this commit.

- Kept: `browser-tools.test.ts` and the existing browser assertions in
  `agent-service.providers.test.ts`.

**I watched each new and changed test fail before restoring the code.** Eight distinct breaks — the
viewport guard, the input quota, `retainDocuments`, `dispose`, `releaseReservation`, `UPLOAD_SCOPE`,
`hasBrowserTakeoverForAgent`, the duplicate-tab clause and `endTakeover` — each broke exactly the
case that covers it, and each failure message names the consequence rather than the tree. Two
assertions were rewritten during that pass because their first failure message was uninformative.

The two review fixes were watched fail the same way, in two halves each. Dropping the `message`
redaction leaves `refresh failed: Authorization: Bearer sk-live-...` in the ring; dropping only the
`target`/`detail` half leaves `ada@example.com` in the action target. Dropping the evaluate
cancellation makes the Electron suite hang at the V2 phase until its hard timeout, instead
of reaching the `queueRecovered` check — that hang *is* the finding, reproduced.

Fix 25 changed an existing test rather than adding one: `frees the previous copy when the same input
is given different files` asserted the behaviour the finding is about, so it became `keeps the
previous copy readable when the same input is given different files` — and it still ends by
navigating the document away and watching both copies go, so the retention has a bound. Watched it
fail with the immediate deletion restored: `promise rejected "Error: ENOENT: no such file or
directory" instead of resolving`, on the first file's staged path.

Fix 24 was watched fail the same way: awaiting the dispatch directly again, with the bound removed,
gives `V2 legacy click never returned from a frame that answers nothing: null` — the null being the
20-second race the assertion holds it against, so the failure *is* the hang.

Fixes 20 and 21 were watched fail one at a time as well. Dropping `modifiers` from the character
event gives `V2 Shift+Enter reached the page unshifted: keypress:Enter:false` — the review's finding
reproduced exactly. Reverting the `wait_for` bound gives `V2 wait condition never returned from a
frame that answers nothing`. Fix 22 needed no new assertion: reverting the lease counter turns the
suite's *existing* bounded-action phase red with `No target available`, which is how it was found.

Fixes 8, 9 and 10 were watched fail one at a time, since the Electron suite stops at its first
error. Reverting the tool's `submit` to a second `press` gives `V2 type did not submit through the
node it typed into: Error: Stale browser reference. The target changed after the snapshot`;
reverting only the `redactText` on the tool error gives `V2 page exception carried a page secret to
the provider: Error: page said password=hunter2`; reverting the takeover recheck gives
`expected true to be false` on the refusal. Reverting the engine's `submit` block outright also
breaks the suite's *existing* legacy iframe submit phase before either new phase is reached, so the
bug's blast radius was wider than the phase added for it.

Fixes 11, 12 and 13 were watched fail the same way. The redaction case goes red with
`expected 'Error: Target is ambiguous (at least …' not to contain 'hunter2'`; dropping `\r` and
`" "` from the alias table gives `V2 press did not insert a space: … "value":"open"`; dropping the
shadow-root branch from the walk gives
`V2 document enumeration lost an upload document inside a shadow-root iframe.` Both smoke reverts
were re-run after the phase moved onto its own page, so the evidence is against the code as merged.

Fixes 14-16 were watched fail the same way, one revert per run against the final code:
`values contains an invalid string.` for the rejected empty option;
`Native select interaction did not produce the requested selection (expected 2, got 1)` with the
focus round-trip removed; and
`V2 document enumeration reported a truncated scan as complete and lost an upload document.` with the
completeness guard removed. Fix 16 was found by a failing test rather than by reading — the assertion
for fix 15 went red for a reason that had nothing to do with empty values, which is what sent me
looking at the buffer.

Fixes 17 and 18 were watched fail the same way, one revert per run against the final code. Restoring
`keepQueueBlocked(completion)` gives
`V2 snapshot timeout left the tab queue waiting on a CDP command it never cancelled.` — the tab's
`close` never returns, which is the finding reproduced. Restoring the `modifiers === 0` character
gate gives `V2 Shift+Enter inserted no newline: … ("line")`, with the tool's own action history
recording the press as `outcome: "success"` beside it.

Fix 23's two assertions were watched fail one at a time, one revert per run against the final code.
Reverting the preflight bound gives
`V2 upload preflight never returned from a frame that answers nothing: null` — the `null` is the race
timer winning, which is the finding reproduced. Reverting the legacy `act` settle bound gives
`V2 legacy act never returned from a frame that answers nothing: null`.

The three new chips are covered by `src/renderer/stories/BrowserPanel.stories.tsx` (5 stories:
idle, mobile viewport, all three chips at once, a narrow panel, and the single-error label), since
colour and layout belong in a story rather than a test.

## Verification: end-to-end on a real machine

Four levels, all passing. Everything ran against an isolated dev instance `5190`
(`OpenBot Dev 5190`); the shared `OpenBot Dev` profile was never touched.

**A. Offline Electron suite — `bun run test:browser`: exit 0, every phase green**, re-run after fixes 8-10 and again after 11-13. Snapshot/actions, V2
semantics, adaptive image mode, iframes, upload, waits, emulation, Chromium page/request identity,
child-tab ownership, screenshot, shared cookies, cache revalidation, download, runtime tool argument
schemas, agent tab isolation, agent control lifecycle, persisted tabs, plus all four cross-process
persistence phases (write/read/clear/verify-cleared).

**B. Live-network probe — `bun run test:browser:google-live`: exit 0, 18 phases,** including
"Google identifier step passed without signin/rejected".

**C. Isolated dev app driven through `bun run dev:automation`.** Instance registered; opened the
computer panel; navigated the address bar; `pages` listed both embedded views as CDP targets; read
the embedded page itself through `--page=` and captured it. The viewport chip tracked the real panel
size (`1200×800` → `423×747`).

**D. A real Codex agent (CLI 0.146.0, model Luna) issuing `openbot_browser` calls** against a local
fixture page. All eleven requested steps succeeded:

| Tool | Observed result |
| --- | --- |
| `open` | fixture tab opened |
| `snapshot` | returned `FIXTURE_READY`; revisions incremented 9 → 10 → 11 → 13 across calls |
| `set_environment` | mobile preset → `390×844` @3, matching the toolbar chip; later back to desktop → `1440×900` @1 |
| `recording_start`/`recording_stop` | a real 73,005-byte WebM, 36.3 s, valid EBML magic `1a45dfa3`, mode `0600` |
| `select_option` | `cherry` |
| `set_checked` | `checked (isTrusted=true)` |
| `upload_files` | `/tmp/pr128-upload.txt` → `pr128-upload.txt:75b`, staged from outside the workspace |
| `drag` | `dropped handle (isTrusted=true)` |
| `evaluate` | returned the four joined page outputs |
| diagnostics | the fixture logged `token=not-a-secret`; the agent received `token=[redacted]` |
| `wait_for` | timed out as `Error: Browser wait condition timed out.` and released control |
| `request_takeover` | the card rendered with a live page preview and `Cancel` / `I'm done`; `I'm done` returned `{"status":"completed","next":"Take a fresh snapshot and continue the task."}` and the agent resumed |

`isTrusted=true` on the checkbox and drop handlers is the trusted-input claim holding against a real
page. One claim in an earlier revision of this body was wrong and is corrected here: the fixture's
`token=not-a-secret` did reach me as `token=[redacted]`, but that redaction was the provider CLI's
own, not this code's — nothing redacted the diagnostics ring until fix 5 above. The automated review
caught what this level could not distinguish; `browser-diagnostics.test.ts` now asserts it at the
boundary that owns it. The "controlled by agent" indicator was caught mid-call: the tab's accessible name became
`PR128 Fixture, controlled by Chief`, with the tab and panel outlined and a dot on the toolbar
button.

Screenshots — before/after the panel opens, the three chips, the embedded page, the takeover card,
and the control indicator — are on the machine under `.openbot-build/dev-automation/pr128/`.

Not run, by policy: `bun run check`, `check:desktop`, `test`, `build-storybook` — CI owns those, and
all four are green on this head.

### What was *not* verified by hand

- **10 of the 23 tools never went through a live model**: `list_tabs`, `status`, `navigate`, `type`,
  `press`, `hover`, `scroll`, `act`, `screenshot`, `close_tab`. All of them are driven by Level A and
  by `browser-tools.test.ts`; none was driven by a provider issuing the call.
- **Only the takeover *complete* path ran live.** `Cancel` is covered by `attention-registry.test.ts`
  only.
- **The 10-input-per-tab upload quota** is unit-tested, never hit live.
- **Only the Codex provider** ran end-to-end. Claude and Grok rest on the schema-drift assertions in
  `browser-tools.test.ts` and on `claude-client.test.ts`.
- **The 5 Storybook stories were typechecked and built by CI, but never visually rendered by me.**
- **No pre-existing v1 `browser-tabs.json` was upgraded on a live profile.** The v1→v2 path is
  covered by `browser-state.test.ts` and by Level A's persisted-tabs phase.
- Level A's input phase failed three times out of five on this machine with stray characters in the
  typed value (`runs locallyription`), because the suite focuses real Electron windows and any
  keystroke made during the run lands in the page. Not a code failure; worth knowing before reading
  a red local run.

## Model and harness

Ported, reviewed and verified with Claude Opus 5 (`claude-opus-5`) in Claude Code, on macOS
(darwin 25.6.0), Electron 44, Bun 1.4.0.


NorbertBodziony:
Both `[NEW][P1]` findings were correct and are fixed in 3631e8c6.

**3. Redact console diagnostics before returning snapshots** — fixed, and wider than reported. Redaction now happens in `BrowserDiagnostics` on the way *into* the ring rather than at each reader, so the ring never holds the secret at all: that covers the snapshot `textResult`, the renderer's diagnostic error count and the Team API projection, not just the provider path. `target` and `detail` on the action ring got the same treatment, since a target description quotes the element's accessible name and an error detail quotes the page's own exception message.

**4. Cancel asynchronous evaluation when its timeout expires** — fixed, using the pattern `#runAction` already had. `#runEvaluation` now calls `cancelPendingCommands()` when the timeout fires and skips the `operationCompletion` drain once cancellation is confirmed. `scripts/browser-smoke-electron.ts` grew a `new Promise(() => {})` phase ahead of the existing `queueRecovered` check; without the fix the suite hangs there until its 60-second hard timeout, which is how the deadlock was reproduced before fixing it.

---

The two `[REMAINS][P1]` findings are accurate about what the code does. Neither is a capability the agent does not already hold, and both were the developer's explicit decision when this port was planned. Concretely:

**1. Browser uploads (`UPLOAD_SCOPE`, `allowAnyReadablePath: true`).** Agents run with `sandbox: "danger-full-access"` (`src/backend/agent/thread-lifecycle.ts:111` and `:137`), so an agent asked to upload `~/.aws/credentials` can `cp` it into its own workspace and reach the identical bytes through the confined path. Removing the scope adds a step, not a boundary — and `src/main/AGENTS.md:59` states this repo's own position: the process boundary is what is left, because "agents already run with `danger-full-access`". What the scope drops is exactly two checks, the containment guard and the symlink pre-check; the TOCTOU half all still runs — regular-file check, `O_NOFOLLOW` on the fully resolved path, and dev/ino re-validation after the open (`src/backend/agent/attachment-gateway.ts:225-242`). The alternative the finding asks for is a tool that cannot upload a file the user names, which is the whole use.

**2. `evaluate` in the authenticated main world.** Isolating it would change nothing that matters: an isolated world shares the DOM and `document.cookie`, so `Page.createIsolatedWorld` is not a mitigation here — which is why the branch does not pretend it is one. And the storage the finding names is reachable without the tool: the session is `persist:openbot-browser` (`src/backend/browser-host.ts:154`), whose Cookies and Local Storage databases are ordinary files under the app's userData, readable by the same `danger-full-access` shell. What *was* wrong is that the description shipped to all three providers called the evaluation sandboxed; that is fixed — it now says it runs in the main frame's own page context with the same access to page scripts, DOM and cookies as the page itself.

So the honest position on 1 and 2: the risk statements hold, the proposed remedies do not reduce the attacker's reach while they do remove the feature, and the trade-off is recorded in the PR body under "Security note: the upload sandbox is deliberately wider". If a gate is wanted anyway it should be a user-facing approval prompt for both tools, not silent containment — that is a product decision, not this port.

/norbiai review

NorbertBodziony:
Both new findings are fixed in `ad35c80e`, plus one more found while fixing them.

**P1 — a page exception carried the page's own text to the provider unredacted.** Fixed at the one place every browser tool failure reaches a provider: the `catch` in `BrowserHost.handleDynamicTool` now applies the same `redactText` the diagnostics ring applies, which covers all 23 tools in one line rather than patching `exceptionDescription` in the engine.

**P1 — a takeover starting during upload staging could still write files into the page.** Fixed in `BrowserUploads` at `onUploadTargetResolved`, which is the only hook that runs *inside* the queued operation and before the input is assigned; the reservation is invalidated so the staged copies are removed rather than left for the next document change. The facade's pre-flight check stays where it is — it is the fast path — but it can be stale by the time large files finish staging, and `BrowserHost` checks tab ownership, not takeover.

**Found while fixing those: `type` with `submit: true` inserted the text and never submitted it.** Enter was pressed as a second action against a re-resolved target, so for a contenteditable the ref fingerprinted the element against its *pre-typing* text and threw "The target changed after the snapshot". Submitting now happens inside the lease that resolved the node, behind an options object matching `click()`'s existing idiom.

Watched each fail first, one at a time, since the Electron suite stops at its first error:

- reverting the tool's submit → `V2 type did not submit through the node it typed into: Error: Stale browser reference. The target changed after the snapshot`
- reverting only the `redactText` → `V2 page exception carried a page secret to the provider: Error: page said password=hunter2`
- reverting the takeover recheck → `expected true to be false` on the refusal

Reverting the engine's `submit` block outright also breaks the suite's *existing* legacy iframe submit phase before either new phase is reached, so that bug's blast radius was wider than the phase added for it.

New coverage: two Electron smoke phases (a `submit: true` through a snapshot ref asserting the page's own `onkeydown` saw a trusted Enter, and a page exception carrying a secret asserting the provider gets `[redacted]`) and a sixth case in `browser-uploads.test.ts` that also asserts the staged copies are gone, since a refusal that leaves them behind is a file the next document change would hand to a page the user is holding.

Green on this head: `bun run lint` (5 pre-existing mobile `no-module-mocking` warnings only), `bun run typecheck` (0 errors), `browser-uploads.test.ts` (6 passed), and the full Electron suite including all four cross-process persistence phases.

NorbertBodziony:
All three findings from the latest review pass are fixed in `f39a7cbe`. Each was verified against the code before I touched anything, and each was watched fail first.

**P1 — upload errors carried a page-controlled secret.** Confirmed: `browser-cdp.ts:1027` builds `Target is ambiguous (at least 2 matches). Candidates: ${sample}` from `candidate.name.slice(0, 80)`, and only `BrowserHost.handleDynamicTool` redacts — everything `BrowserUploads` does *around* that call throws past it to the facade, which sends `String(error)` on. Fixed by wrapping the whole of `uploadFiles` and rethrowing a redacted `Error`, so target resolution, source opening, the quotas and staging are all covered by one boundary that stays inside the browser subsystem. New unit case; red before the fix with `expected 'Error: Target is ambiguous (at least …' not to contain 'hunter2'`.

**P2 — `press` sent no character event.** Confirmed: the `char` event was gated on `key.length === 1`, which no named key satisfies, and `normalizeKey` had no character to give it. Chromium decides implicit form submission and text insertion from the character event, so `press("Enter")` on a scriptless `<form>` did nothing and `press("Space")` inserted nothing. `normalizeKey` now carries the character; `Tab` deliberately stays characterless, since the browser moves focus on the key event and a later character would land in whatever gained focus, and a modified shortcut gets none because `Ctrl+S` is a command rather than an `s` in the document.

**P2 — document enumeration stopped at shadow boundaries.** Confirmed: `documentIdsExpression` was the only traversal left in the file using `querySelectorAll` — the other six use the `createTreeWalker` walk that pierces shadow roots and descends into frames. The consequence is worse than a miss: a document that looks closed frees the staged files a still-open input is holding. It now uses the same walk.

Both P2s share one Electron phase on a new `/keys` page. It got its own page rather than more markup on `/v2` because that page's dimensions are asserted against a 220×560 panel and a block form plus a default-sized iframe put a scrollbar in it — which turned the existing narrow-panel phase red, so the fixture moved instead of the assertion. Fail-side evidence, one revert at a time: `V2 press did not insert a space: … "value":"open"` and `V2 document enumeration lost an upload document inside a shadow-root iframe.` Both reverts were re-run after the move, so the evidence is against the code as merged.

`lint`, `typecheck` (all 11 projects) and `check:ui` clean; the 5 remaining lint warnings are pre-existing `no-module-mocking` in `apps/mobile`. Full Electron suite green end to end.

NorbertBodziony:
Both findings are fixed in `06585ad3`, and the second one turned out to be the smaller half of a bigger bug.

**P2 — truncated document scan reported as complete.** Correct, and a fair catch against my own fix 13: the tree walk I introduced stops at the 10,000-node budget, which the `querySelectorAll` it replaced never had, so an upload frame past that point is never visited — and a document missing from the list is indistinguishable from a closed one, so `retainDocuments` frees the staged files of an input that is still open. The expression now reports completeness and an unvisited document is presumed open. That direction is deliberate: keeping a staging directory until the tab closes costs a temp directory, while freeing one whose input is live hands the page a path that no longer exists. Covered by growing the keys fixture past the budget before navigating a frame; red without the guard with `V2 document enumeration reported a truncated scan as complete and lost an upload document.`

**P2 — `select_option` rejected empty option values.** Correct: the schema and the engine both accept `values: [""]`, and only `requiredStringArray` refused it. The empty check is now explicit per list — allowed for option values, still rejected for upload paths, where an empty path is never a file.

**Found by that fix: two `select_option` calls on the same select within a second left the value unchanged.** The new assertion went red for a reason that had nothing to do with empty values — `expected 2, got 1` — and a probe with a *non-empty* value on the same option failed identically, which ruled out the validation path. Chromium keeps a typeahead buffer per select for about a second, so the second call appends to the characters the first one typed: after selecting `b`, typing `a` for `Any mode` searched for `ba`, matched nothing, and left the selection alone. No phase had caught it because no phase selected twice on one select. I measured the three candidate fixes before choosing: immediate retry fails, focusing another element first succeeds, a 1400 ms wait succeeds. The engine now does a focus round-trip before typing — instant, and it clears a buffer left by the page or a prior keystroke too, which the wait would not.

One platform limit fell out of that investigation and is report-only. The engine's non-typeahead fallback (`Home` then `ArrowDown`) is inert on macOS: a closed select opens its native popup on `ArrowDown` instead of moving the selection, and no CDP key event reaches that popup. Typeahead is therefore the only channel here, so an option with **no label at all** cannot be selected on macOS however the validation is written. The fallback still works on Linux and Windows, so rather than change behaviour I made the error say it instead of reporting a bare index mismatch.

Full Electron suite green (exit 0, first attempt); `lint`, `typecheck` across all 11 projects and `check:ui` clean.

NorbertBodziony:
All three findings verified against the code and fixed in `05559272`.

**[P1] `browser-host.ts:1166` — the snapshot drain.** Confirmed, and it is worse than a hung
snapshot: Electron's `debugger.sendCommand` carries no timeout, so a frame that never answers the
walk leaves `completion` pending for good, and every later operation on that tab queues behind the
drain — navigation, takeover, tab destruction, host shutdown. The caller's timeout error is what hid
it. `#readSnapshot` now cancels the way `#runEvaluation` does and releases the drain once the detach
is confirmed; when the recorder owns the debugger `cancelPendingCommands()` returns false and the
wait correctly stands, since there is still an outstanding command.

**[P2] `browser-cdp.ts:2040` — `Shift+Enter`.** Confirmed. Shift-only combinations now carry their
character and `event.key` agrees with it (`Shift+a` reports `A`). `Ctrl+S` still gets none. One
deliberate limit: `Shift+1` is `!` on a US layout and something else on half a dozen others, so a
shifted glyph that is not an alias key or a letter stays a key event rather than a guessed character.

**[P2] `browser-uploads.ts:173` — the size check.** Confirmed: the read stops at the size measured
before staging, so growth produced a staged file of exactly the expected length and only shrinkage
was detectable. The open descriptor the copy read from is re-stat'd afterwards and either direction
is rejected.

Watched 17 and 18 fail, one revert per Electron-suite run against the final code: restoring
`keepQueueBlocked(completion)` gives
`V2 snapshot timeout left the tab queue waiting on a CDP command it never cancelled.` (the tab's
`close` never returns); restoring the `modifiers === 0` gate gives
`V2 Shift+Enter inserted no newline: … ("line")` with the press recorded as `outcome: "success"`
beside it. 19 has no test, and the PR body says why: its window opens between the source `stat` and
the copy, and holding a file open at that instant needs a sleep.

`lint`, `typecheck` and the full Electron suite are green.

NorbertBodziony:
Both findings confirmed and fixed in `0ad76b5c`.

**[P1] Bound CDP condition waits and post-action settling** — right on both paths. `wait_for` with a
`text` condition evaluates in every frame and the post-action settle waits for DOM quiet in every
frame, and each checks its deadline only between commands, so the spinning-frame fixture hangs them
with their own timeout running inside the wedged renderer. Both now go through the wall-clock bound
the snapshot already had, extracted as `#boundEngineOperation`, and the `evaluate` settle was given
the same treatment so all three paths behave alike. The wait half is covered: the `/blocking-frame`
phase now runs a `text` condition against that frame, and reverting the bound gives
`V2 wait condition never returned from a frame that answers nothing`. The settle half is not, and
the PR body says so — every way of wedging a frame ends in a detach, which clears the auto-attached
session list, so the settle that follows only reaches the main frame and returns on its own. A click
against the blocked fixture stayed green with the bound reverted, so that assertion was removed
rather than kept as a costume.

**[P2] Include modifiers in the character event** — confirmed; the mask is now on the `char` event.
You were also right that the textarea check missed it: the newline lands either way. The keys phase
now reads a `keypress` logger and asserts `keypress:Enter:true`, and dropping `modifiers` gives
`keypress:Enter:false`.

One thing the first finding turned up that neither of us had: cancelling on *any* timeout is too
eager. Detaching is the only cancellation primitive and it takes the whole session down, so a 5 ms
wait — which is *expected* to time out and has its own existing phase — started cancelling a renderer
that was about to answer, and the suite went red with `target closed while handling command`.
Underneath that was an older bug: `#lease` detached whenever it was the lease that had attached,
which is wrong as soon as two overlap, so an operation past its deadline took the *next* operation's
debugger with it (`target closed while handling command`, then `No target available`, on a healthy
page). Leases are counted now, the last one out detaches, and a timeout waits for the operation to
unwind on its own before detaching anything and again afterwards, so the queue advances into an
attached debugger rather than one being torn down.

Full Electron suite green twice on this head; `lint` clean (5 pre-existing `apps/mobile` warnings),
wide `typecheck` clean.

NorbertBodziony:
Confirmed and fixed in `90fae5c7`.

**[P1] Bound the remaining queued CDP entry points** — all three named sites verified and bounded
through `#boundEngineOperation`, and the preflight is the one that mattered most: `resolveUploadTarget`
is queued *ahead* of the bounded upload action, so on the spinning-frame fixture the tab never reached
the operation the timeout was there to protect. It now takes the tool's own `timeoutMs`, so the
preflight and the action share the caller's budget rather than the preflight having none.

Document enumeration also took the second half of the finding — "preserving staged files when
document enumeration cannot establish completeness". The empty document set its failure path reports
is what tells upload staging the documents holding its files are gone, and it deletes them: correct
when the walk ran and found nothing, wrong when the frames were never asked. A timeout is now
distinguished from a failed walk and leaves the staged files alone. The enumeration itself moved onto
`#enqueue`, which is the same queue chaining it hand-rolled, so the bound and the drain come with it.

Legacy `act` deliberately kept its old outcome: a settle timeout still fails the action. Copying
`#runAction`'s "report success and snapshot anyway" would have changed what the renderer's own browser
panel does on a slow page, which is wider than the finding — and failing there also means nothing else
holds the session when the unwind's detach lands.

Two of the three are covered by assertions in the Electron suite, both reusing the `/blocking-frame`
fixture, and both watched fail one revert at a time:
`V2 upload preflight never returned from a frame that answers nothing: null` and
`V2 legacy act never returned from a frame that answers nothing: null`.

The enumeration guard has no assertion, and the PR body says so with the reason: the empty set is
only reported for a *main*-frame navigation, and a main-frame navigation replaces the document the
wedged subframe lives in — so reaching that branch requires landing on a new page whose own spinning
frame has already attached a CDP session by the time enumeration runs. That is a race no fixture here
can close, and an assertion that goes green when the race falls the other way is worse than none.

The suite's per-process hard timeout went 60 s → 120 s. Four phases now spend ~10 s each proving a
frame that answers nothing is given up on; the backstop exists for a phase that hangs, not as a
performance budget. Full suite: 46 s wall clock, green. `lint` and `typecheck` clean.

NorbertBodziony:
**P1 — restore the timeout around legacy input dispatch: fixed in `98db920a`.**

The finding was right that fix 23 stopped half-way. `act`'s dispatch was awaited directly, and the
`deadline` those engine calls carry is checked between CDP commands, so a renderer that answers none
of them never reaches it — re-resolving the ref fingerprints the element in the frame that owns it,
and that is where a wedged frame holds the click, with navigation, takeover, close and shutdown stuck
behind the tab queue. The whole `switch` is now one operation through `#boundEngineOperation`, with
the remaining deadline as its timeout and the same cancellation-and-drain unwind the other paths use.

You also called out that the blocked-frame test only scrolled a responsive main frame. That is fixed
rather than noted: `act` accepts no target but a ref, and a ref only exists for an element some
snapshot saw, so the spinning fixture now waits to be told to spin and carries a button. The tab is
snapshotted while the frame still answers, the frame is wedged on command, and the ref that snapshot
handed back is clicked. The assertion runs first in the phase, because the unwind every later one
triggers detaches the session the ref resolves through.

Watched it fail: with the bound removed and the dispatch awaited directly again, the phase goes red
with `V2 legacy click never returned from a frame that answers nothing: null` — `null` is the
20-second race the assertion holds the call against, so the failure is the hang itself. Full suite
green afterwards in 58 s.

Separately, `main` was merged again (eight commits, up to `21b07b43`). Clean merge, but the retheme
in #287 collapsed the type scale and dropped the 10px `--openbot-text-2xs` the three new browser
toolbar chips used; they now use `--openbot-text-sm`, and `scripts/design-tokens.test.ts` passes.

NorbertBodziony:
**P2 — retain previously selected files until their document closes: fixed in `5441ec83`.**

Confirmed. The staging map held one directory per input and deleted the old one on reassignment,
which is only correct if `input.files` is the sole reference to those files — and it is not. A page
that took the earlier `File` objects, which is what any "add another file" flow does before it
submits, keeps reading the directory that was just removed.

An input now holds a list of the directories it was given, appended to rather than replaced, and the
list is freed on the paths that already existed: the document changing, the tab closing, and
`dispose`. Those are the points at which nothing can read the files again, which is the invariant the
controller was written around; only the reassignment shortcut was wrong.

Quota accounting followed, as you suggested. `#reserve` no longer subtracts the replaced input's
bytes, so every retained copy counts against the per-tab and the global byte limits — that is what
bounds the accumulation now, instead of the reassignment freeing it. The inputs-per-tab limit is
unchanged, since it is keyed by input.

Watched it fail: `browser-uploads.test.ts` already had `frees the previous copy when the same input
is given different files`, asserting exactly the behaviour the finding is about, so it was rewritten
rather than joined by a new case — it now reads both copies after the second upload and then
navigates the document away and watches both go. With the immediate deletion restored, it goes red
with `promise rejected "Error: ENOENT: no such file or directory" instead of resolving` on the first
file's staged path. Suite green after; the Electron smoke suite is green too.

PR #155, opened 2026-09-02
fix: apply the auto-download setting and end the macOS update hang

### LINKED ISSUE #152: Fix automatic update download and Preparing update hang on macOS
## Problem

With **Automatically download updates** enabled, OpenBot detects that an update is available but does not download it automatically. The user must click **Download update**, after which the updater remains stuck on **Preparing update...**.

Reported on **macOS** with **OpenBot v0.4.0**. It is currently unknown whether Windows is affected.

## Steps to reproduce

1. Run OpenBot v0.4.0 on macOS.
2. Open **Settings → General**.
3. Enable **Automatically download updates**.
4. When an update is available, observe that it is detected but not downloaded automatically.
5. Click **Download update**.
6. Observe that the action remains stuck on **Preparing update...**.

## Expected behavior

- When automatic downloads are enabled, an available update is downloaded automatically.
- A manually started download progresses beyond the preparation phase and provides visible progress or a clear error.
- The updater recovers cleanly from transient failures and does not remain indefinitely in a loading state.

## Actual behavior

- Automatic update download does not start even though the setting is enabled.
- Manual download gets stuck on **Preparing update...**.

## Acceptance criteria

- Automatic update downloads work on macOS.
- Manual update downloads complete successfully or show an actionable error with retry support.
- The updater cannot remain indefinitely stuck in `Preparing update...`; add timeout/cancellation handling and useful diagnostics.
- Verify behavior for both automatic and manual flows, including restart/resume where supported.
- Check whether the same issue affects Windows and document the result.
- Add regression coverage for update detection, automatic download, preparation, failure, retry, and cancellation states.
- Attach the provided screenshots as visual references.

### Automatic download enabled, update available

Settings showing automatic update downloads enabled and Download update available

### Manual download stuck on Preparing update

Download update stuck on Preparing update

NorbertBodziony:
Fixes #152.

## Root causes

Two independent defects, both in app code. The release artifacts are fine — `latest-mac.yml`, the arm64 zip and its blockmap are all published and valid, and the `NorbertBodziony/openbot` → `nightly-labs/openbot` redirect resolves correctly.

**The auto-download setting was never applied.** `UpdateService` hardcoded `autoDownload = false`, and `autoDownloadUpdates` was a renderer-only signal that `updateGeneralSettings` never forwarded over IPC and never persisted — so it reset to `true` on every launch and did nothing. **This half affected Windows too**, since the hardcoded `false` was platform-independent.

**The macOS `preparing` phase could never exit.** The darwin branch waited for Electron's *native* `autoUpdater` `update-downloaded` event. But `MacUpdater.updateDownloaded()` only calls `nativeUpdater.checkForUpdates()` — the call that makes Squirrel fetch and stage the ZIP, and therefore the thing that eventually emits that event — **when `autoInstallOnAppQuit` is true**. We keep it `false` so nothing installs without `prepareForUpdateInstall()` running. So Squirrel was never asked, the event never arrived, and the phase was terminal; the renderer treats `preparing` as busy, which disabled the button. **Windows was never affected by this half** — it called `#markReady` directly.

`MacUpdater.quitAndInstall()` already stages on demand when Squirrel hasn't, so `ready` is valid the moment the download completes and the `preparing` wait was pure liability.

## The fix

- The preference is persisted in `update-preference-store.ts` (reusing the `analytics-preference-store` atomic-write pattern), read at startup, exposed over three new IPC channels, and hydrated into the toggle. Switching it on while an update is already `available` starts that download — the issue's exact repro steps.
- `update-downloaded` → `ready` on all platforms. `preparing` and `prepare_failed` are **removed from the contract**, so the state that hung cannot be reached again.
- Auto-download is driven from `UpdateService` rather than by setting electron-updater's own `autoDownload`, so the automatic and manual flows share one download path, one cancellation token, and one set of progress/watchdog/error semantics.
- Every busy phase is bounded by a single timer re-armed on each status write, so download progress refreshes the stall deadline for free. A stalled check, download, or restart now reports an actionable error and retries in place.
- Cancellation needs no new dependency: `checkForUpdates()` already returns the `CancellationToken` it hands to the download, so the service captures and reuses it.

### A second dead end this also closes

`BaseUpdater.quitAndInstall` calls `install()` and, when that returns `false` (missing installer path, or a throwing installer), **returns normally without quitting**. That left `#installStarted` stuck at `true` forever, so the restart action could never become available again. The error handler now releases the latch, with a test covering it.

## Keeping it from coming back

More tests in the existing style would not have helped: the old test `"waits for native macOS staging before it exposes restart"` **passed while the product hung forever**, because its hand-written fake emitted an event the real library never sends. A fake encodes the author's belief about the dependency, so a wrong belief gets certified.

**A busy phase without a deadline is now a compile error.** The enabling condition was duplicated phase lists — the renderer said `preparing` was busy while the service had no bounded exit, and nothing connected those facts. The lists now live in `packages/contracts` as the single source of truth (`UpdatePhase` is derived from the array), and the service's timeout table is typed `Record<UpdateBusyPhase, number>`. Verified by reintroducing `preparing` as a busy phase:

```
src/main/update-service.ts(72,7): error TS2741: Property 'preparing' is missing in type
'{ checking: number; downloading: number; installing: number; }'
```

`update-status.ts`, `AccountUpdateIsland.tsx` and `SettingsModal.tsx` now call the shared `isUpdateBusyPhase`/`isUpdateActivePhase` instead of keeping their own copies. Backing that is a driven test that enters each busy phase for real and proves the service leaves it — its driver table is also typed over `UpdateBusyPhase`. Confirmed to catch a regression by stubbing out the download watchdog.

**The upstream assumptions are pinned.** `electron-updater-assumptions.test.ts` reads the *installed package* rather than a fake and asserts the four load-bearing facts the fix depends on, pinning version `6.8.9`. It is deliberately source-level: driving the real `MacUpdater` needs Electron plus an HTTP proxy and a downloaded archive, and the alternative — a fake — is exactly what hid this bug. A dependency bump that changes this logic now fails loudly.

## Verification

Lint and typecheck clean. **+49 passing tests.**

Full suite: 31 failed / 1375 passed, against a clean-HEAD baseline of 32 failed / 1326 passed — the same three pre-existing failing files (`agent-service`, `openbot-database` with "cannot start a transaction within a transaction", and flaky queued-message tests in `App.test.tsx`). Nothing new broken.

Two caveats:

- **End-to-end cannot be verified locally.** The service is constructed with `enabled: false` unless `app.isPackaged` **and** `resources/app-update.yml` exist, so `checkForUpdates()` short-circuits in dev. Confirming the real macOS flow needs a signed build updating from a prior release — an available update auto-downloading with visible progress, reaching "Restart to update" without ever showing "Preparing update…", and restarting into the new version; then the same with the toggle off. Worth repeating on Windows to close out that acceptance criterion.
- This suite is flaky enough under load that one intermediate run showed 10 failing files, all of which passed in isolation. A single red CI run here is worth re-running before trusting it.

## Note for users already affected

Anyone stuck on "Preparing update…" on 0.4.0–0.4.2 has **no in-app recovery**, because the UI disables the button while busy. They need to quit and reinstall once from the DMG; every release from this fix forward updates normally. Called out in `CHANGELOG.md` and `docs/RELEASING.md`.


NorbertBodziony:
Addressed the NorbiAI review in d1ff251.

**[P1] Install timeout disabled before the installer starts — valid, fixed.** Confirmed: `beforeInstall` is `prepareForUpdateInstall` → `prepareForShutdown` → `updateService.stop()` (`index.ts:2250`), and `stop()` cleared the phase timer before `quitAndInstall` ran. The watchdog was dead in production, so the dead end it was added to close was still reachable — most realistically on macOS, where `quitAndInstall` returns immediately when the ZIP is not staged and only quits once Squirrel finishes, so a silent stall there latched the app in `installing` with no escape. `stop()` now preserves an in-flight install deadline and `installUpdate` re-arms it after the handover. The existing test missed this because its fake `beforeInstall` never stopped the service; the new test models the real one, and I verified it fails with `expected { phase: 'installing' } to match object { phase: 'error' }` against the unfixed code.

**[P3] Rapid toggle changes can persist the wrong preference — valid, fixed.** Writes are now serialized in invocation order, so the last toggle wins on disk and replies come back ordered. Test covers four overlapping writes. (Note `analytics-preference-store` has the same pattern; left alone as out of scope here.)

**[P2] Download cancellation exposed but unreachable — valid observation, taking the removal option.** We do not want user-facing cancellation, so the IPC channel, preload method, `UpdateDesktopApi` member, `UpdateService.cancelDownload()`, preview mock entry and the documentation claim are all removed. Cancellation stays internal, where the stall watchdog needs it to stop a dead transfer, and the two tests that covered it now drive it through that watchdog instead — which is the real scenario rather than a synthetic one.

Lint, typecheck and the affected suites are clean.

NorbertBodziony:
Second review round addressed in a80fe2b. Both findings were valid.

**[P2] A timed-out install can still restart the application later.** `installUpdate` awaited `beforeInstall` and then called `quitAndInstall` unconditionally, so shutdown preparation that outlived the 60s deadline went on to restart the app *after* the UI had reported the attempt as failed — and because the watchdog releases the latch, a retry could run two attempts concurrently. An install generation is now taken before the await and checked after it (and in the catch), so a superseded attempt returns without touching the installer. The watchdog bumps the generation when it fires.

**[P2] A permanently stalled check disables future automatic checks.** Confirmed: the next periodic check was scheduled only in that call's `finally`, which never runs for the exact hang the check watchdog exists to catch — so the app silently stopped checking for updates until restart. The timeout branch now owns the next schedule, and the `finally` only schedules while its generation is still current, so a late completion cannot push out the newer timer.

Both are covered by new tests, each confirmed to fail against the unfixed code:
- `does not restart the app once the install deadline has reported a failure` → `expected "vi.fn()" to not be called at all, but actually been called 1 times`
- `keeps checking on schedule after a check never settles` → `expected "vi.fn()" to be called 2 times, but got 1 times`

Lint, typecheck and the updater suite (27 tests) are clean.

NorbertBodziony:
Third review round addressed in 05051e9.

**[P2] Late updater events can revive timed-out operations — valid, and fixed structurally rather than by another guard.** This was the real shape of the problem: status had two writers, and only one of them was generation-guarded. The awaited results were; the event listeners were not.

Rather than gate five listeners, the check lifecycle now has exactly one owner. Every check flows through `checkForUpdates()`, whose result is already generation-guarded, so the `checking-for-update`, `update-available` and `update-not-available` listeners are **removed** rather than gated — a second unguarded writer for a transition that is already owned is the bug shape, not just this instance of it. That also let the internal-check suppression flag go.

The remaining listeners are scoped to the attempt the service still believes in: a download and an install each record their live generation, the watchdogs retire it when they abandon an attempt, and progress, completion and error events are dropped unless they belong to a live one.

Three of the four new tests fail against a80fe2b:
- `ignores a progress event that arrives after the stall watchdog gave up`
- `ignores a completed download that lands after the stall watchdog gave up`
- `ignores a late error from an operation the user has moved on from`

The fourth (`does not let an abandoned check result overwrite a newer download`) passes against the old code too — the post-await generation guard already covered it. Kept as a guard on that invariant, not claimed as a fix.

**[P3] Dependency-source tests duplicate the version pin — accepting the substance, keeping a tripwire.** You are right that with an exact pin the compiled source cannot change without a deliberate bump, so the regexes could never fail independently of the version they duplicated, and they would have added formatting churn to any upgrade. Those 7 assertions are gone and the file is 93 lines down to 33.

I kept one assertion on the *resolved* version, because deleting the file outright removes the only mechanism that would notice upstream drift: the behavioural tests run against a fake, and a fake cannot notice when the real library stops behaving as written against — that is precisely how #152 stayed green. The four load-bearing behaviours are now documented on `UpdateAdapter` in `update-service.ts`, where they are read alongside the code, and the test's failure message names exactly what to re-verify before updating the pin.

Lint, typecheck clean; the updater suite is 31 tests.

NorbertBodziony:
Fourth review round addressed in 71cf5e4. Findings 1-4 were valid; I verified the mechanism behind 1 and 3 in the installed package before changing anything.

**[P2] Timed-out checks cannot actually retry — valid, and it shared a root cause with finding 3.** Confirmed at `AppUpdater.js:257-260`: `checkForUpdates()` returns the outstanding promise when one is in flight, and `downloadUpdate()` does the same via `downloadPromise`. So my whole "retire the attempt and retry" model never produced a *fresh* operation — a hung check re-entered `checking` and timed out again every interval while never issuing a request. Both operations now refuse to start while the previous attempt is unsettled, and the check loop keeps rescheduling so a refusal is never the last check of the session.

**[P2] Late updater events can revive timed-out operations — REMAINS was correct.** `#isDownloadLive()` only proved *some* download was live, as you said. Rather than bind events to an attempt-specific adapter, refusing a retry until the abandoned attempt settles removes the window entirely: there is no longer a moment where a new attempt is live while an old transfer is still emitting. That is the first of your two suggested fixes, and it falls out of the finding-1 fix for free.

**[P2] Install timeout permits concurrent teardown — valid, and the right conclusion was to stop offering retry at all.** `beforeInstall` is shutdown preparation and is not transactional, so there is no state to safely retry from once it has begun. The latch now stays set, install failures report that OpenBot has to be reopened, and the renderer no longer offers an install retry. This partly reverses the latch release from round two — that fix was right about the dead end but wrong to make install retryable; the watchdog reporting an actionable error is what removes the dead end.

**[P2] Preference hydration can overwrite a newer toggle — valid, fixed.** A local change marks itself and a startup read landing afterwards is dropped. Worth noting my first attempt at this test passed without the fix because it asserted before the hydration continuation ran; I caught that, added an explicit flush, and confirmed it now fails without the guard.

Every fix is covered by a test confirmed to fail against 05051e9. The stalled-check test also caught a bug in this change itself: refusing a retry without rescheduling would have ended periodic checks for the session.

**[P3] Assumption test — still declining, and flagging it for a human.** Your objection is right that it asserts a constant `package.json` already pins. My reason for keeping it is that deleting it leaves nothing that notices upstream drift: the behavioural tests run against a fake, and a fake cannot notice when the real library stops behaving as written against — which is exactly how #152 stayed green while the product hung. That is a judgement call about a 33-line tripwire rather than a defect, so I have left it and asked the repo owner to arbitrate rather than keep going back and forth.

NorbertBodziony:
Fifth review round addressed in 10e5a41.

**[P2] Failed token refresh still starts an untracked download — valid, fixed.** Exactly as described: a spent token forced a refresh, the refresh failed, and the attempt went ahead with `downloadUpdate(undefined)`. The stall watchdog then had nothing to cancel, the transfer ran untracked with its events discarded as unattributable, and an unsettled one blocked every later retry. It now reports `download_failed` rather than starting. Test confirmed to fail against 71cf5e4.

**[P2] Install timeout returns to a torn-down application — valid on the substance; fixed differently from the suggestion.** Two changes:

The deadline moves from 60s to **300s**. On macOS `quitAndInstall` asks Squirrel to stage the ZIP and only quits once that finishes, which is disk-bound on an artifact of a few hundred megabytes — so 60s was reporting failures over installs that were still working. That was a tuning bug on my side.

Once shutdown preparation has begun, the service now refuses checks and downloads for the rest of the process's life. This is the reachable half of your finding: a failed install leaves an `error` phase that the ordinary phase guards let a check through, so the renderer really could run a check in a dead app.

I did **not** implement force-exit on timeout. Given macOS staging can legitimately outlast any deadline we pick, force-quitting risks aborting a working install part-way — a worse outcome than a stale window with an actionable message. If you still want a forced exit I would rather it be driven by a signal that staging has actually failed, not by elapsed time.

**[P2] Timed-out checks cannot actually retry — declining, because the premise does not hold.** A permanently hung promise is not reachable through the network path: `builder-util-runtime`'s `HttpExecutor.addErrorAndTimeoutHandlers` defaults to a 60s socket timeout that calls `request.abort()` and rejects (`httpExecutor.js:141`, `:278-284`), which clears `checkForUpdatesPromise`. So the call settles and the next scheduled check issues a real request.

You did surface a real tuning bug though: my deadline was also 60s, so it *raced* the library's. It is now 90s, deliberately above it, so the library times out first and reports the actual error while mine only covers a stall outside the request itself. An abortable check adapter would mean reimplementing provider resolution and HTTP, which is not proportionate to a failure mode the dependency already bounds.

**[P3] Assumption test — unchanged, deferred to the repo owner.** I have made my case and you have made yours; it is a judgement call about a 33-line tripwire, not a defect, so I am leaving it for a human to settle rather than continue the exchange.

PR #164, opened 2026-09-02
Show detailed live agent activity

Sniezka1927:
## Summary
- show the latest streamed commentary in the active agent status row
- retain generic activity text until commentary is available
- clamp live status details to two lines and preserve completed Thinking history

## Tests
- fnm exec --using=22 bun run test -- src/renderer/src/App.test.tsx -t 'agent activity row|preserves omitted attention only'
- bun run typecheck:renderer
- Biome checks
- git diff --check


Sniezka1927:
The P3 scope finding about Grok error-detail handling should be withdrawn. Commit 7796c00 added that code specifically to resolve NorbiAI’s earlier P1 finding that detailed Grok errors exposed the Authorization bearer secret. Reverting it here would reintroduce that validated security issue. The new accessibility P2 is fixed in e479ea0: the changing activity label is now content of an atomic polite status region, with assertions for provider-detail updates. /norbiai review

PR #169, opened 2026-09-02
docs: define the test types we want and the seams that require them

NorbertBodziony:
## What changed

The test value policy in `AGENTS.md` is a list of prohibitions with no positive map, so tests accumulate where rendering is easy and stay thin where a defect is expensive. Today 40% of the suite sits in the renderer, while the renderer-to-main trust boundary, the IPC contract surface, the migration engine, and the provider clients have no tests at all.

This splits the guidance by whether an agent needs it **to decide** or only **while writing**:

- **`AGENTS.md`** gains five decision rules that apply to every change, because this file is loaded into every session: the default is no test; check whether `tsc`, Biome/GritQL, `check:ui` or Storybook already enforces the rule; pure logic goes in the vitest `node` project even when the module lives in the renderer; the list of seams where a test is mandatory, with an imperative pointer to the reference; and `App.test.tsx` is closed to new cases.
- **`docs/TESTING.md`** is new and holds the reference an agent opens on demand: three gating questions, the mechanisms that are cheaper than a test, **seventeen named test types** with what each asserts and its pattern in this repository, a table mapping each mandatory seam to the types it requires, what not to test, and a review checklist.

Deliberate constraints:

- **No rule is stated in both files.** The prohibitions stay in `AGENTS.md`; `docs/TESTING.md` points at them rather than restating them. The seam list appears in both, but as names in one and detail in the other — a hierarchy, not a duplicate.
- **No line numbers or file sizes** in the document, so it does not rot into confidently wrong references.
- **No enforcement added.** No check script, no CI gate, no changes to any test. Documentation only.
- The additions to `AGENTS.md` are appended after the existing bullets, leaving the current policy lines untouched.

Follow-ups this surfaced but does not address: `test:sites` (security tests for the router) and `remote:check` (including ES256 ticket verification) are defined in `package.json` but run by neither `bun run check` nor CI, and `typecheck:sites`, `typecheck:team-client` and `mobile:typecheck` never run in CI either.

## Verification

- [x] `bun run lint` — clean, 759 files
- [x] `bun run typecheck` — clean, all 8 projects
- [ ] `bun run check` — **fails locally with 2 flaky tests, unrelated to this change.** `src/backend/agent-service.test.ts` fails with `Codex request timed out: initialize` and one `App.test.tsx` case exceeds its 5s timeout, both only under the full parallel run. Each passes when run in isolation (91/91 and 1/1). This branch differs from `main` by two Markdown files only, so it cannot affect test behaviour.
- [x] Relevant manual smoke test completed, or not applicable — not applicable, documentation only

Also verified: every file path cited in the document exists, and the document answers unambiguously for five sample tasks (an IPC payload change, a new `*-store.ts`, a new D1 migration, a button-variant fix, a new provider stream parser) whether a test is required, which type, and in which vitest project.

## Risk and security impact

None. Documentation only; no code, configuration, permissions, or CI behaviour changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,3 +52,8 @@
 - Do not test the same rule at the component and application levels. Prefer one test at the lowest stable behavior boundary. Keep only one application-level happy path when it protects a critical integration.
 - A UI test must fail when behavior or accessibility breaks. It must not fail only because markup, wording, or styling changes.
 - For a bug fix, add a regression test only when it reproduces the failure at a stable boundary and can prevent the same failure from returning.
+- The default answer is no test. A test is an exception that must name the consequence it protects.
+- Before writing a test, check whether `tsc`, Biome and its GritQL rules, `bun run check:ui`, or Storybook already enforces the rule. If one of them does, do not write the test.
+- Put pure logic in the vitest `node` project even when the module lives in the renderer.
+- A test is mandatory when a change touches the renderer-to-main trust boundary, the IPC contract, database schema or migrations, persisted state, secrets, the provider process boundary, the Team API wire protocol, or the updater. Read `docs/TESTING.md` before writing it.
+- `src/renderer/src/App.test.tsx` is closed to new cases. New coverage goes to a module or to a component test at the lowest stable boundary.
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -1,0 +1,207 @@
+# Testing guide
+
+This document decides **whether** a test is required, **which type** it should be,
+and **where** it belongs. Read it before writing a test at any seam listed in
+section 4.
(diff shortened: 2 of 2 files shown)


NorbertBodziony:
Superseded by #170, which deletes the machinery this PR deepens.

#170 removes `docs/TESTING.md`, `scripts/test-value-check.ts`, and `scripts/policy-sync-check.ts`, and moves the test policy into `AGENTS.md` as six rules with Biome owning the mechanical half. Defining more test types and a fuller seam table pulls the opposite direction, so this is closing rather than merging.

The part worth keeping — the seams that make a test mandatory — survives as rule 6 in the new `## Tests` section.

PR #175, opened 2026-09-03
refactor(renderer): decompose the App god controller into domain contexts and a keyed server scope

NorbertBodziony:
Breaks up the `App.tsx` god controller into domain contexts on the opencode
`packages/app` model, and moves per-server state out of global
`Record<serverId, …>` maps into a keyed provider subtree.

`App.tsx` was 4194 lines, 92% of them one function — `createAppController`, with
92 `createSignal` calls, 35 non-reactive instance fields and a flat ~190-key
return object behind one context. It is now **14 lines**.

Model and harness: Claude Opus 5 in Claude Code.

## What changed

| | before | after |
|---|---|---|
| `App.tsx` | 4194 lines | 14 |
| `AppView.tsx` | 855 lines | 130 |
| `WorkspaceShell` | 430 lines, destructuring 133 controller keys, 18 `use*()` | 87 lines, 6 contexts |
| state | 1 context, 92 signals, ~190 keys | one context module per domain |
| per-server state | `Record<serverId, …>` + a 20-setter manual teardown | a `<Show keyed>` subtree that disposes on switch |
| `activeServerId() !== serverId` after an await | 65 hand-written comparisons | `createScopeGuard()` |

The nesting order in `app-providers.tsx` **is** the dependency graph: a domain may
read one it is nested inside, never one nested under it. `noImportCycles` at error
severity (landed in `6736d2d`) makes that a lint failure rather than a convention.

Three cross-cutting readers could not become contexts — each reads five to nine
signals across three or more domains — so `AgentEventBridge`, `DynamicIslandBridge`
and `AppBootstrap` are `null`-returning components at the bottom of the tree.
Nothing imports them, so the edge points one way.

`createSimpleContext` is ported from opencode and adapted to Solid 2:
`<Ctx value={…}>`, no `@ts-expect-error`, no `typeof`, no `Record<string, any>` —
all three would fail lint or the approvability rules here.

## Correctness this fixes

The per-server load sequence existed **twice** — once in the bootstrap block, once
in `selectServer` — and the copies had already drifted: the bootstrap caught errors
per promise, `selectServer` used one `Promise.all` with no `catch`, so a single
rejection killed all eight loads there. First mount and server switch are now the
same provider mount, so the sequence exists once.

`activeServerId() !== serverId` got A → B → A wrong: the id matches again even
though the owner that issued the request was disposed in between.
`createScopeGuard()` asks whether the owner is still alive. The per-request
counters beside it are kept — they order concurrent writes *within* one owner,
which disposal cannot express.

`completedTurnByBot` grew unbounded for the life of the process on `main`. It is
now bounded by the scope's lifetime.

## Tests

Wave 1 (the migration) added **no** tests: the net is the 160 existing application
tests. Every step was checked by breaking what it covered and watching the failure
name the consequence. Six new pure modules and their tests came out of the review
pass, each at a boundary that did not exist before, plus four application-level
cases for the four real regressions the review found — the superseded island
action, the two microphone lifetimes, and the typing indicator's destination.
Each of those was watched failing too; the messages are quoted under
"Review follow-up".

### Watch it fail — migration steps

`<` → `<=` on the read-state revision guard: 2 red · drop the `deferred` branch: 1 red ·
drop `update.getStatus()`: 2 red · drop the platform `blur` listener: 8 red ·
drop `auth.getState()`: 156 red · `completedCodeChallenge = false`: 1 red ·
`saveSetup` stops closing the review flow: 1 red · drop `.then(setSetupState)`: 151 red ·
drop `localStorage.setItem` for the panel width: 1 red ·
drop the `serverLoadRequest` publish: 1 red · drop `servers.onPresence`: 7 red ·
drop `servers.onDirectMessage`: 4 red · `applyStoredBots` stops opening first-Bot setup: 1 red ·
`activeQueue` → undefined: 4 red · `markAgentMessagesRead` guard inverted: 26 red ·
`resetForServer` no-op: 1 red · `beginBrowserLoad` guard inverted: 2 red ·
`toggleSidebarSection` no-op: 1 red · `selectBot` drops `agentChatsToMarkRead.add`: 9 red ·
drop the `autoDownloadUpdatesChanged` guard: 1 red · `openAppSettings` no-op: 4 red ·
`connectProvider` drops `applyAgentStatus`: 3 red · `openServerSettings` no-op: 2 red ·
`hideRemoteDesktopWorkspace` stops hiding: 2 red · drop the active-server-first sort: 1 red ·
`computeSidebarAgentStates` returns `{}`: 2 red · `createAgent` stops adding to `botList`: 1 red ·
the event bridge ignores every event: 58 red · `turn-completed` stops clearing the active turn: 1 red ·
`selectServer` stops calling `resetAgentsForServer()`: 4 red ·
the projection drops `replaceServer(input)`: 4 red ·
initial load drops `.then(applyConversationReads)`: 2 red ·
the focus handler stops calling `setRecentReplies({})`: 1 red ·
`<AppProviders>` moved inside the remountable `<Show>`: 1 red ·
`WorkspaceShell` stops reading `conversationReads`: 14 red.

### Watch it fail — the keyed scope

- `turns.tsx` seed `seed?.queues ?? {}` → `{}` → 1 red: `Unable to find role="group" and name "Queued message 1: Next work"`.
- `components/Sidebar.tsx` cleanup drops `resizeObserver.disconnect()` → 1 red: `expected 7 to be 5`.
- `conversation.tsx` loses the new `autoReadAgentMessages` guard → 1 red: `expected "vi.fn()" to be called 2 times, but got 3 times`.
- `ui-errors.tsx` keys on `agentConversationKey("", botId)` → 8 red.
- `ScopedProviders` builds its own `createStableConversationState` instead of receiving the hoisted one → 6 red.
- `AgentReadTrackingProvider` moved from above `ServerScopeBoundary` into `ScopedProviders` → 1 red: `does not carry a failed automatic read to the same bot on another server`.
- `createScopeGuard` → `() => Boolean(owner) || true` → 2 red: `stops opening a Dynamic Island message when the active server changes`, `discards a chat-open reload that resolves during a server switch`.
- `dynamic-island-bridge.tsx` drops `|| !loaded()` → **GREEN, unprotected**.
- `agent-read-tracking.tsx` was **not** mutated directly and here is why: it holds only Maps and Sets with no logic, so its behaviour is enforced by its callers and that mutation is already on record above (`selectBot` drops `agentChatsToMarkRead.add` → 9 red). Its *placement* was mutated instead.

### Behaviours the existing tests did not protect

Found by mutating real code and running all 162 application tests. Each stayed
green, so each is a hole this refactor could have fallen into silently:

`expandSidebar`'s auto-compact clear · `host.onEvent` · the cross-server guards in
`markDirectMessagesRead` · `resetForServer` clearing direct threads and
conversations · `resetForServer` clearing `botList` · `resetForServer` re-seeding
`activeTurns` · `mutateSidebarLayout`'s `setSidebarLayout(layout)` · the
runtime-snapshot revision guard `<` → `<=` · `refreshServerSettings`'s
stale-request guard · `createServerInvite`'s refresh · `startRemoteDesktopConnection`'s
stale-request guard · the retain-effect prune · the queue half of `working` and the
whole `working` branch in the sidebar badge · `deleteBot` moving the selection ·
`createAgent` closing the form · `duplicateBot`'s capability and in-flight guards ·
a failed switch restoring `dynamicIslandLoadedServerId` · the projection's
`serverId !== activeServerId()` guard · a submitted prompt coming back on the next
snapshot · the live-message dedupe · takeovers overlaying prompts · approvals
surviving `attentionComplete` · the whole next-active-tab selection after a close.

The last six are now covered by the new modules below. The rest are named here
rather than tested: covering them all is a separate topic, and this PR is already
one.

### New tests — six pure modules

Each names a boundary that did not exist before, and each of their cases was
watched failing. The full mutation lists are in the individual commit bodies;
the headline for each:

| module | cases | measured gap before |
|---|---|---|
| `conversation-merge.ts` | 7 mutations | page/window merge arithmetic had no test of its own |
| `queue-reconciliation.ts` | 12 mutations | three memos inside a 3500-line component |
| `sidebar-agent-states.ts` | 5 mutations | deleting the whole `working` branch stayed green |
| `browser-tab-reconciliation.ts` | 8 mutations | next-active-tab → `remaining[0]?.id ?? null` left **all 20** `App.browser` tests green |
| `agent-runtime-snapshot.ts` | 15 mutations | four of five snapshot behaviours unprotected |
| `conversation-read-state.ts` | net of 31 existing tests | — |

Two mutations stayed green and are equivalent mutants rather than gaps, and both
are argued in their commit bodies: taking the first recognised loaded index instead
of the last in a latest-anchored window (a contiguous suffix has the same bounds
either way), and mutating only the left side of the queue position comparison (an
inconsistent comparator, not a behaviour — applied to both sides it reddens).

### Test premises changed, explicitly

Three, each keeping every assertion:

1. Both `Harness`es: "a controller object handed to a provider" → "a provider subtree above a remountable view".
2. `setupTests.ts`: `TestResizeObserver.instances` now tracks observers that are *observing* rather than observers that were *constructed*. The browser keeps an observer alive only while it has a target, so counting constructions reports a leak for any library that releases with `unobserve` (`@tanstack/virtual-core` is one) and misses nothing in return. Proven by the `Sidebar.tsx` mutation above.
3. `App.servers.test.tsx::loads capability-gated state when a provisional handshake becomes ready`: asserts the reload rather than `servers.select("remote-1")`, because a completed handshake now remounts the scope instead of re-selecting an already-active server.

Nothing was deleted, skipped or weakened.

### Two tests the plan asked for, and why they are not here

- **`server-scope.test.tsx`** — provably unnecessary. Its three claims are already
  measured elsewhere: seeding is protected by the `turns.tsx` seed mutation (1 red,
  `restores the queue of a server the user comes back to`); the disposal guard by
  the `createScopeGuard` mutation (2 red); subscription disposal by
  `App.servers.test.tsx:913 does not accumulate event subscriptions across server
  switches`, which asserts `subscriberCounts()` and `TestResizeObserver.instances.size`.
  All three consequences are cross-server assertions, which the plan's own policy
  keeps at app level, and a new *file* needs a boundary that does not exist yet.
- **`agent-event-pipeline.ts`** — measured to be the wrong shape. The 21-case switch
  fans out to nine domain contexts, calls IPC, plays a sound and mutates
  non-reactive Maps; a `(event, state) → patches` reducer would need one combined
  state object spanning five contexts, which is the god controller this PR removes.
  All the *arithmetic* the switch held has moved out instead —
  `mergeConversationPage`, `queueAfterTurnCompleted`, `reconcileQueuesWithRuntimeWork`
  and the three snapshot folds. The routing stays.
- **`runtime-settings-order.ts`** — skipped for the same reason.
  `saveRuntimeSettings` (`ConversationView.tsx:956-1024`) is promise-tail chaining,
  generation counters and signal setters. That is lifecycle, so it stays on its module.

## The IPC call-trace diff

The 160 application tests assert rendering results, not *which* IPC calls happen in
*what order* — which is exactly what splitting one controller into providers
changes. So each `vi.fn` in `app-test-harness.ts` was wrapped locally with a
recorder, and the per-test method-name sequence was dumped on `main` and on this
branch and diffed. The instrumentation is **not committed** — it is a diffing tool,
not a test.

162 tests here, 160 on `main`, 160 shared. **No IPC method appears on one branch and
not the other**, and no test lost its Dynamic Island projection. Every difference
falls into four groups:

1. **Re-subscription on scope remount** (27 tests): `agent.onEvent`, `servers.onPresence`,
   `dynamicIsland.onAction`, `voice.onModelStatus`, `browser.onPictureInPictureEvent`,
   `agent.onAttachmentImport`, `browser.setVisible`, `browser.closePictureInPicture`.
   The keyed boundary disposes and rebuilds, so a switch re-subscribes. That the old
   subscriptions actually go away is what the new
   `does not accumulate event subscriptions across server switches` test asserts.
2. **`dynamicIsland.publishPresentation` coalescing** (35 tests, net −13, both
   directions): the projection publish is scheduled and coalesced, and the schedule
   points differ. No test dropped to zero.
3. **The merged load sequence** (2 tests): `connects to a remote host after account
   sign-in` performs one *fewer* full per-server load set. This is the duplicated
   bootstrap/`selectServer` sequence collapsing into one — the diff the plan
   predicted.
4. **`servers.select` for an already-active server** (2 tests, −3 calls):
   `keeps a newer online event when retry returns an older summary` and
   `loads capability-gated state when a provisional handshake becomes ready`. In
   both, `main` re-issued `servers.select` against the server main already reports
   as `active` purely to trigger a reload; the keyed remount reloads without it. In
   the first test `servers.select` is mocked to reject, so `main` also ran the
   `servers.list()` recovery and a second `select` — both disappear with the
   redundant call that provoked them.

## Behaviour changes

- Direct-message page caches no longer survive a server switch.
- `showComposerActions`, `markingRead`, `dropActive`, `rightPanels`, `settings*`, `browserAddress*`, previews, popovers and chat search now reset on a server switch. Previously they reset only when `Conversation` unmounted.
- Panel widths, composer drafts, voice state, `uiErrors` and unsettled read markers now **survive** a switch. A voice message transcribed after the user has moved on still fails against the server it was dictated on, and its error has to be there when they come back.
- A `Conversation` unmount no longer sets `voiceDisposed`.
- A completed provisional handshake reloads the workspace by remounting the keyed scope instead of re-issuing `servers.select` for an already-active server.
- `markLatestVisibleAgentMessageRead` no longer duplicates a read an optimistic read already owns.
- An A → B → A round trip discards in-flight work from the first A scope. Previously it was accepted, because the id matched again.
- `InitialSetup` and `DirectConversation` now share one lazy wrapper each.
- `TestResizeObserver.instances` (test-only) tracks observing rather than constructed observers.

## Surfaces

**Touched**: desktop renderer only (`src/renderer/src`) — `App.tsx`, `AppView.tsx`,
the new context and pure-logic modules, `app-test-harness.ts`, `setupTests.ts` and
the 8 `App.*.test.tsx` files. Plus one entry in `docs/ARCHITECTURE.md` "Change
rules" saying where renderer state belongs.

**Reverse states** are the surface here, since after the keyed scope each is a
disposal path: server A → B → **A**, sign-in → sign-out → **sign-in**, browser
suspend → **resume**, PiP enter → **exit**, Remote Control connect → hide →
**resume** → disconnect, tab and dialog open → **close**.

**Provably untouched**: `packages/contracts` — not one IPC call shape changed, and
that was an acceptance criterion; `apps/mobile`, `apps/auth-api`, `packages/team-client`
(separate workspaces, no import path into `src/renderer`); the three palettes (no new
`--openbot-*` token, no CSS change); migrations and the latest schema; the `README.md`
command table; `PRIVACY.md`; the Team API protocol and its adapters.

## Verification

- `vitest run src/renderer` → **66 files, 640 tests, all passing**
- `biome check src/renderer/src` → **273 files, no findings**. Run over the whole directory deliberately: `noImportCycles` only sees a cycle when one invocation holds modules from both ends of every edge, so passing two edited files makes that gate blind.
- `tsc --noEmit -p tsconfig.web.json` → clean, as are the other seven `typecheck:*` projects.
- No `biome-ignore`, `@ts-expect-error`, `@ts-ignore`, `biome.json` override, or widening to `any`/`unknown`. Auto-approvable.

**Two manual passes could not be run** and are owed before merge. Several agents
share one dev stack and one profile on this machine, so a second `bun run dev` is
not safe to start:

1. `bun run dev`: A → B → A with the built-in browser open, PiP active, a hidden Remote Control session, a composer draft and an open settings dialog. Each is a disposal path and none is fully covered by a test.
2. `bun run storybook` → `App / Playground`: `preview/mock-openbot.ts` is a second implementation of the IPC contract and CI only typechecks it. If a keyed provider calls a method the stub does not handle, the app hangs there and CI stays silent.

Everything else — `check:desktop` on macOS, the full test job with three workers,
the Surfaces and Storybook build jobs — belongs to CI.

## Review follow-up

Seven rounds of automated review raised thirteen findings. Eleven are fixed
here, one is a duplicate, and one I answered instead of fixing. The seventh round
raised none.

The one I did not fix is *completed turns resurrected by the seed*, raised three
times. The backend clears `activeTurnId` and persists it
(`agent-service.ts:3693`, `openbot-database.ts:921`) *before* it emits
`turn-completed` at `:3736`, so a post-switch read cannot name a finished turn.
The renderer side is the stronger half of the answer: `dynamic-island.tsx:78`
subscribes `onScopedEvent` **above** the keyed boundary, and
`dynamic-island-coordinator.ts:164-176` nulls `runtime.activeTurns[botId]` on
`turn-completed` for *any* server, active or background. The coordinator's
snapshot therefore cannot carry a finished turn into a fresh scope, which is why
`completedTurnByBot` starting empty is correct rather than lossy.

The same argument was offered for *answered prompts resurrected by the seed* and
it was wrong there, for a reason worth naming: `replaceServer` deletes its
`resolvedPrompts` marker as soon as the scope stops reporting the prompt as
pending, which is exactly when the scope is torn down. That one is a real
regression and it is fixed below.

**A superseded island action could erase the newer action** (`7f269de`). A real
regression this PR introduced: the `pendingIslandAction` handoff exists only
because the keyed remount replaces the bridge, so on `main` a superseded selection
could only ever drop its own action. Actions for B then C arriving while the bridge
was still on A meant C's `selectServer` bumped the generation, B's returned `false`,
and B's unconditional clear deleted C before C's scope consumed it. The bridge now
clears only when the pending handoff is still the one that request published.

One new case in `App.test.tsx` covers it, watched failing against the unfixed
bridge — `keeps a newer Dynamic Island action when the older one's selection is
superseded`: `expected [] to include 'office-message'`, the Office conversation
never opened around its message. Both existing supersession cases stay green,
because with nothing newer pending the action is still its own and is still cleared.

**Conversation teardown could open the microphone after navigation** (`7a8f4d9`).
A real regression. On `main` the controller was created inside `Conversation`
(`main:Conversation.tsx:252`), so unmount set `voiceDisposed = true` and killed the
in-flight `prepareModel()` / `getUserMedia()` chain. Moving the controller above the
per-server scope — which is what keeps drafts and an in-flight voice send alive
across a switch — left that chain running after the view is gone, and it went on to
open the microphone with no recording UI on screen.

The guard sits on the hardware-acquisition step alone, after the failure branch
returns: a model *failure* still belongs to the conversation that asked for it,
and `App.voice.test.tsx:77` (unchanged from `main`) pins that. Two new cases,
both watched failing — `does not open the microphone for a conversation the user
has left`: `expected "vi.fn()" to not be called at all, but actually been called 1
times`; and `stops offering a recording the user walked away from`, with only
`setVoicePhase("transcribing")` mutated out: `Expected the element to have
attribute: aria-disabled="true"`.

**The typing indicator was released to the wrong server** (`e60ede2`). Main moves
its active server the moment `servers.select()` resolves, while the keyed boundary
disposes the scope only after `setServers`, so the clear reached the server being
entered. `stopComposerTyping()` and `setDirectTyping(false)` now run before
`servers.select()`, while the source server is still active, and
`stopComposerTyping` moved onto the controller so the view, app teardown and the
switch share one implementation.

Worth being precise, because an earlier comment of mine on this PR got it wrong:
on `main` a switch sent no `typing: false` at all — `main:AppView.tsx:487` wraps
`<Conversation>` in `<Show when={!blockedRemoteServer() && !botSetupOpen() &&
!activeDirectMember()}>` and none of those flip on a switch. So the stale indicator
on the source server is inherited, and only the misdirected call was new. Both are
gone. One new case in `App.servers.test.tsx`, watched failing: `expected
[ 'typing on', 'select', 'typing off' ] to deeply equal [ 'typing on', 'typing off', 'select' ]`.

**A pending voice setup blocked unrelated conversations** (`ee45bc7`). The same
hoist as above kept `voicePhase` alive with the drafts, and that one does not
belong to the app: a model download or permission prompt started in one
conversation left the next one showing "Downloading voice model" on a disabled
microphone until a promise settled that may never settle. Leaving a conversation
now releases `preparing` and `requesting`; `transcribing` stays, because unlike
those two it names work already running that finishes into a draft the user can
still reach.

The phase was also the abandoned chain's only staleness check, and it cannot be
that once a later attempt can sit in the phase an earlier one was abandoned in.
`resources.voiceRequestGeneration` replaces it, which keeps the two halves
separable: an abandoned chain still reports its failure to the conversation that
asked for it, but can no longer move a phase that now belongs to someone else.

Both halves watched failing, against one existing case extended and one left
alone — release removed: `does not open the microphone for a conversation the user
has left` red with `expected document not to contain element, found <button
aria-label="Downloading voice model">`; generation counter swapped back for the
phase check: `shows a deferred voice setup error in the original conversation` red
with `Unable to find an element with the text: Local voice setup failed`.

**The context readiness gate nothing used** (`5f3a1ff`). `createSimpleContext` was
ported with a `ready` option that withholds a provider's children. Twenty domains
later not one passes it: the app gates on data in one place, the `<Show>` ladder in
`AppView.tsx`, which was already there and reads better than the same decision
spread across twenty providers. The option, the branch, its `createMemo`/`Show`
imports, the documentation on writing a gate, and the gated test are gone; the
provider is now a context wrapper. That test is deleted because the option no
longer exists, not to make anything pass — which is the one line of the plan this
PR does not deliver, and the reason is in the code rather than in the effort.

**A queued settings save could be applied to the next server** (`ceb2013`).
`runtimeSettingsSaveTails` and `runtimeSettingsAttempts` are keyed by bot id on
the same hoisted controller, and `agent.updateBot` is routed by the server main
has selected rather than by the bot id in its payload
(`preload/index.ts:78-83`), so a save still queued behind an earlier one when the
user left was applied to whichever server they arrived at. It is dropped now:
the conversation that asked for it is gone, along with the pickers it would roll
back and the composer it would report to. Both keys name the server as well, so
a pending attempt from one server can no longer overlay another's header.

Unlike the three above, this one is not a regression: `main:AppView.tsx:487`
keeps `<Conversation>` mounted across a compatible switch, so the controller and
both maps survived there too. It is fixed here because those maps *are* the
layer split this PR is about — every other resource on the controller
(`voiceServerId`, `voiceSubmitRequest`, `browserOpenRequests`,
`importTargetBots`) already carries its server, and these two were the only ones
that did not. Read from `main`, not run against it.

One new case in `App.settings.test.tsx`, watched failing against the unfixed
save — `does not send a queued settings save to the server the user switched
to`, which reports the offending call by name: a second `updateBot` for
`claude-sonnet-5` arriving after the switch, where only the `claude-opus-5` save
belongs.

**Contributor guidance cited a deleted module** (`42e429e`). `CONTRIBUTING.md`
pointed at `app-controller-context.tsx`, which this PR removes. It now names
`components/conversation-controller-context.tsx` and `team-typing.ts`.

**An answered prompt came back after a server switch** (`975b4f2`). A real
regression. Answering a prompt is proved only by the snapshot that follows it,
and `submittedPromptRequests` is what keeps `pendingPrompts` empty in between
(`agent-runtime-snapshot.ts:35`). On `main` that map was global and nothing ever
cleared it, so the gap could outlast a switch; scoping it to `turns.tsx` ended it
at the switch instead, and a `runtime-snapshot` arriving after the user returned
offered the answered prompt again. Both markers now live in a new
`answered-prompts.tsx` above the keyed boundary, keyed by server id and then bot
id, and are re-exported from `useTurns()` under their own names, so no consumer
changes. Keying by server rather than by bot id alone is deliberate: bot ids are
`bot-${uuid}` (`backend/bot-store.ts:182`) and unique in practice, but
`isValidBotId` (`:770`) admits any slug, so this does not have to depend on it.

One new case in `App.servers.test.tsx`, watched failing against the unfixed
`turns.tsx` — `does not offer an answered prompt again after leaving its server
and coming back`: `expected document not to contain element, found <input`, the
answered prompt's own answer field back on screen after `local → remote-1 →
local`.

**…and came back a second way, through the seed** (`58b6f75`). The review was
right that the fix above closed only one of the two entrances, and the
measurement is worth recording: a probe on the `turns.tsx` seed prints
`SEEDPROBE local {"chief":{"type":"prompt",…}}` on the return leg. The Dynamic
Island coordinator's memory *is* the renderer's own projection played back
(`dynamic-island-bridge.tsx:66-89`), and answering a prompt does not clear
`pendingPrompts` — `submitPromptAnswers` only records the marker, and
`QuestionPromptBubble` hides the questions in local state. A remount throws that
state away, so the seeded prompt asked again, on its first page, with the
answers gone. `seededAttentionPrompts` (`agent-runtime-snapshot.ts:62`) filters
the seed against both markers, next to the reconciliation that does the same
against main's snapshots; browser takeovers are left alone because
`respondToBrowserTakeover` clears `pendingPrompts` itself.

The same case now waits on the composer, which `ConversationView.tsx:3471` hides
for as long as anything is being asked, so it covers the interval *before* a
snapshot arrives instead of asserting only after one — which is exactly the
interval the review said the first version missed. Watched failing against the
unfiltered seed: same message, `expected document not to contain element, found
<input`, this time with no snapshot involved at all.

**A send in flight disabled the next server's composer** (`975b4f2`).
`attachmentBusy`, `composerError`, `submitting` and `selectionSending` sat in the
stable half of the conversation controller, above the boundary, which made the
layer-2 doc comment untrue: unlike everything else there, they carry no key at
all. Each describes the composer on screen right now, so a send left in flight on
the server being left disabled the arriving server's composer for as long as it
stayed unresolved, and showed that server's failure underneath it. They move into
the scoped half; `conversationErrors` is keyed and stays.

One new case in `App.servers.test.tsx`, watched failing — `does not leave the
next server's composer disabled by a send in flight`: `Received element is not
enabled`.

**Island actions acted on a half-loaded workspace** (`975b4f2`). The bridge waits
out `loaded()` for an action that has to switch servers, but an action for the
server that was *already* active ran `selectBot` against an empty bot list and
read an empty conversation. It now joins the same handoff, which the readiness
effect drains once the scope has filled; `publishedAction` stays untouched,
because nothing disposes this bridge in the meantime and it is the one that
consumes its own entry.

## The plan's second wave, measured: there is no PR left to open

The plan split the work in two — refactor under the existing tests, then re-home
them, one PR per family. The re-homing turned out to have happened already, in
the wave-1 PR that made each extraction, and the residue is kept at app level by
the plan's own policy rather than by neglect. The count, over the ~169 app-level
cases:

- **Pushed down, at the point of extraction**: six pure modules with their own
  node tests (`conversation-merge`, `conversation-read-state`,
  `queue-reconciliation`, `browser-tab-reconciliation`, `sidebar-agent-states`,
  `agent-runtime-snapshot`), listed above.
- **Stays at app level, policy 3** — rule 7's boundaries plus every cross-server
  assertion. On `App.read-state.test.tsx`, the file the plan expected to shrink
  from 31 to ~12, this is nearly all of it: nine cases turn on window focus,
  which no reducer can see; fourteen are the `markConversationRead` persistence
  path, its retries and its rollbacks; two (`:397`, `:534`) the plan itself
  promotes to keyed-scope acceptance criteria.
- **Deleted: none.** Policy 4 wants a citation per deletion, and the one
  candidate does not survive its own citation. `App.read-state.test.tsx:23`
  (`merges a refreshed remote conversation page without dropping loaded
  messages`) looks duplicated by `conversation-merge.test.ts:53`, but the pure
  test covers the merge; the app case covers which *mode* the `conversation-page`
  event merges in. Pick "replace" there and loaded messages vanish with every
  push from main, and the pure test stays green.
- **Double coverage introduced by wave 1: none.** The one app-level assertion
  that reads like a `computeSidebarAgentStates` duplicate
  (`App.servers.test.tsx:483`, `Chief is working`) uses the indicator as an
  observable for compact-snapshot attention merging, which is a different claim.

So the honest report is that wave 2 is complete and empty, not skipped. Splitting
the remaining app tests further would mean either testing routing against a
fixture that mirrors it, or asserting the same consequence twice — the two things
rule 7 and the value policy exist to stop.

## Follow-up, not this PR

`vitest.config.ts:13-17` mutes `[STRICT_READ_UNTRACKED]`, the prerelease diagnostic
that catches lost reactivity. With the filter disabled locally it fires **zero**
times across all 66 renderer files, so it is currently muting nothing and could be
removed. Different topic.


NorbertBodziony:
Thanks — worked all three findings. Two fixed here, one is pre-existing and belongs in its own PR.

**2. [P2] A superseded island action can erase the newer action — confirmed, fixed in 7f269de.**

Real regression, and one this PR introduces: the `pendingIslandAction` handoff exists only because
the keyed remount replaces the bridge, so on `main` a superseded selection could only drop its own
action. Actions for B then C arriving while the bridge is still on A meant C's `selectServer` bumped
the generation, B's returned `false`, and B's unconditional clear deleted C before C's scope
consumed it. The bridge now clears only when the pending handoff is still the one that request
published.

Covered by a new case in `App.test.tsx`, watched failing against the unfixed bridge:

```
× keeps a newer Dynamic Island action when the older one's selection is superseded
AssertionError: expected [] to include 'office-message'
```

— the Office conversation is never opened around its message. Both existing supersession cases stay
green: with nothing newer pending, the action is still its own and is still cleared.

**3. [P3] Contributor guidance cites a nonexistent module — confirmed, fixed in 42e429e.**

`app-controller-context.tsx` was deleted by this PR. `CONTRIBUTING.md` now names
`components/conversation-controller-context.tsx` and `team-typing.ts`, both of which exist on this
branch and illustrate the same rule.

**1. [P1] Typing cleanup after the server switch — pre-existing, not a regression, separate topic.**

The ordering is identical on `main`. `git show main:src/renderer/src/App.tsx` around 3181–3190:
`setServers(nextServers)` and `setActiveBotId("")` both run *after* `await
window.openbot.servers.select(serverId)` resolves, so the `Conversation` unmount and its
`setTyping(false)` already fired after main had switched the active server there too. This PR moves
the code but not the timing, and per "one topic per PR" the fix wants its own change.

Verification for the two fixes: `vitest run src/renderer` — 66 files, 637 tests, all passing;
`biome check src/renderer/src` clean; `tsc --noEmit -p tsconfig.web.json` clean.

NorbertBodziony:
Both findings from the last review are fixed, and one thing I told you earlier was wrong.

**1. [P1] Microphone after navigation — real, fixed in `7a8f4d9`.**

Confirmed by archaeology rather than reasoning: on `main` the controller was created *inside*
`Conversation` (`main:Conversation.tsx:252`), so unmount set `voiceDisposed = true` and killed the
in-flight `prepareModel()` / `getUserMedia()` chain. This PR moved the controller above the
per-server scope so drafts and an in-flight voice send survive, which left that chain alive after
the view is gone.

The guard goes on the hardware-acquisition step alone, after the failure branch returns. A model
*failure* still belongs to the conversation that asked for it whether or not that conversation is
still open, and `App.voice.test.tsx:77` ("shows a deferred voice setup error in the original
conversation", unchanged from `main`) pins exactly that — resetting the phase in the unmount
cleanup would have broken it.

Two tests, both watched failing:
- *does not open the microphone for a conversation the user has left* →
  `expected "vi.fn()" to not be called at all, but actually been called 1 times`
- *stops offering a recording the user walked away from*, with only `setVoicePhase("transcribing")`
  mutated out → `Expected the element to have attribute: aria-disabled="true"`

**2. [P2] Typing cleanup targets the newly selected server — real, fixed in `e60ede2`.**

Your ordering diagnosis is right and that is how it is fixed: `stopComposerTyping()` and
`setDirectTyping(false)` now run in `server-selection.tsx` before `window.openbot.servers.select()`,
while the source server is still the active one. `stopComposerTyping` moved onto the conversation
controller so the three owners that need it — the view, the controller at app teardown, and the
switch — share one implementation.

New test in `App.servers.test.tsx`, watched failing: removing the two calls gives
`expected [ 'typing on', 'select', 'typing off' ] to deeply equal [ 'typing on', 'typing off', 'select' ]`.

**Correction to my earlier comment on this finding.** I said the timing was "identical on `main`".
The reasoning behind that was wrong. On `main`, `AppView.tsx:487` wraps `<Conversation>` in
`<Show when={!blockedRemoteServer() && !botSetupOpen() && !activeDirectMember()}>`, and none of
those flip on a server switch — so `Conversation` never unmounted and `main` sent **no**
`typing: false` at all on a switch.

That changes the shape of the finding, not its verdict:
- the *stale indicator on the source server* is inherited from `main`, not caused by this PR;
- the *misdirected call to the destination server* is new here (though `notifyTeamTyping` sends
  `botId: null` on a false, so it was a no-op at the far end).

Both are gone now, so the switch is strictly better behaved than `main` rather than merely restored.

NorbertBodziony:
Third review triaged. Findings 3 and 4 are fixed (`ee45bc7`, `5f3a1ff`). Findings 1 and 2 I could not reach — evidence below rather than a fix, because the remedy both ask for is the cross-server memory this PR set out to delete.

### 1 — completed turns resurrected after a round trip: not reachable

The backend clears the turn before it announces it.

`src/backend/agent-service.ts:3693` sets `snapshot.activeTurnId = null`. `:3727` then calls `#emitConversation(snapshot, "turn.completed", …)`, which persists that snapshot through `src/backend/openbot-database.ts:921` (`UPDATE projection_threads SET active_turn_id = ?`, with the now-null value). Only at `:3736` does the `turn-completed` event reach the renderer. And the read path merges live over persisted only when live still has a turn — `agent-service.ts:1260`, `live?.activeTurnId ? mergeConversationSnapshots(persisted, live) : persisted` — so both sides of that expression are already clear by the time the renderer learns the turn finished.

So a conversation read issued *after* the switch cannot name the completed turn, and a read issued *before* it never lands: every async application path in `conversation.tsx` sits behind `scopeIsCurrent()` (`:163, :189, :265, :275, :318, :652, :679, :686, :695`), and that guard is exactly what the keyed boundary flips. `completedTurnByBot` protects one race — a read already in flight when the completion event arrives — and that race is within one scope, which is now exactly the map's lifetime.

Corroboration that the backend is already trusted here without the map: `agent-event-bridge.tsx:233` replaces `activeTurns` wholesale from the runtime snapshot, unguarded, and does so on `main` too. If the backend could report a finished turn as active, that line would already be wrong.

### 2 — submitted prompts reappearing: not reachable, for a different reason

The seed is `pendingPrompts`, and the coordinator keeps `pendingPrompts` correct across the switch on its own.

The two markers are internal to the turns/conversation/bridge triangle — nothing renders them, and no reader exists outside those three files. They only ever suppress a prompt that is in `pendingPrompts`, so the question is whether the seed can restore an answered one. It cannot:

- `respondToPrompt` emits `agent-input-resolved` synchronously (`agent-service.ts:1500`), before its own reply reaches the renderer.
- If A is still the active server when that arrives, A's bridge clears `pendingPrompts` (`agent-event-bridge.tsx:187-193`) inside a `flush`, so `DynamicIslandBridge` publishes the cleared value before disposal.
- If A is already in the background, the event arrives on `onScopedEvent`, whose subscription lives above the keyed boundary and is never torn down (`dynamic-island.tsx:79`); the coordinator clears `runtime.pendingPrompts[botId]` and records `resolvedPrompts` (`dynamic-island-coordinator.ts:182-189`), which `replaceServer` then uses to suppress the re-seed (`:53-57`).

There is no third case: main moves its active server when `servers.select()` resolves, which is strictly before `setServers` disposes the scope.

The duplicate-*submission* guard is also not the markers'. The backend holds it: `agent-service.ts:1481`, `if (!pending) throw new Error("This prompt is no longer active.")`. A second answer fails closed and the renderer reports "Answer failed".

If either of these is wrong I would rather see the counter-case than guess — but the remedy on offer is per-server memory above the keyed boundary, and `completedTurnByBot` grew for the life of the process on `main`, so I want the scenario before I put it back.

### 3 — pending voice setup blocking unrelated conversations: fixed, `ee45bc7`

Correct, and a real regression. Hoisting the conversation controller above the keyed scope kept composer drafts and an in-flight transcription alive across a switch; it kept `voicePhase` too, and that one does not belong to the app.

Leaving a conversation now releases `preparing` and `requesting`. `transcribing` stays, because unlike those two it names work already running that finishes into a draft the user can still reach.

The generation token was load-bearing rather than belt-and-braces: the phase was also the abandoned chain's only staleness check, so releasing it silently broke the deferred-error path. `resources.voiceRequestGeneration` separates the two — an abandoned chain still reports its failure to the conversation that asked for it, but can no longer move a phase that now belongs to someone else.

Watched both halves fail:

- release removed from the cleanup → *does not open the microphone for a conversation the user has left* red with `expected document not to contain element, found <button aria-label="Downloading voice model">`
- generation counter swapped back for the phase check → *shows a deferred voice setup error in the original conversation* red with `Unable to find an element with the text: Local voice setup failed`

### 4 — the unused readiness API: fixed, `5f3a1ff`

Agreed — it came from the plan, not from anything the code turned out to need. Twenty domains later nothing passes `ready`, because the app gates on data in one place, the `<Show>` ladder in `AppView.tsx`, which was already there and reads better than the same decision spread across twenty providers. The option, the branch, its `createMemo`/`Show` imports, the documentation on how to write a gate, and the gated test are gone; the provider is now a context wrapper. The gated test is deleted because the option no longer exists, not to make anything pass.

### Verification on `5f3a1ff`

`biome check src/renderer/src` clean over 273 files; `tsc --noEmit -p tsconfig.web.json` and `-p tsconfig.storybook.json` clean; `vitest run src/renderer` 66 files / 639 tests green — 640 before, minus the gated case.

NorbertBodziony:
Third round: two resolved, one new, two repeats.

**New, P1 — fixed in `ceb2013`.** The finding is right about the mechanism and I had missed it. `agent.updateBot` is routed by the server main has selected (`preload/index.ts:78-83`), not by the bot id in its payload, so a settings save still queued behind an earlier one when the user leaves is applied to whichever server they arrive at. It is dropped now, and both maps are keyed by server and bot rather than bot alone.

One correction to the finding's framing: this is not caused by the hoist. `main:AppView.tsx:487` wraps `<Conversation>` in `<Show when={!blockedRemoteServer() && !botSetupOpen() && !activeDirectMember()}>`, and none of those flip on a compatible switch, so on `main` the controller and both maps survived a switch too. It is fixed here anyway, because those maps are exactly what the layer split is about: every other resource on the controller (`voiceServerId`, `voiceSubmitRequest`, `browserOpenRequests`, `importTargetBots`) already carries its server and these two were the only ones that did not. That reading is from `main`'s source, not from running the case against it.

**The two repeats.** Findings 2 and 3 are unchanged from the previous round and so is my answer, with the file-and-line evidence in https://github.com/nightly-labs/openbot/pull/175#issuecomment-5520534664. Briefly:

- *Completed turns.* `agent-service.ts:3693` clears `activeTurnId` in memory and `#emitConversation` at `:3727` persists that through `openbot-database.ts:921`, both strictly before `turn-completed` is emitted at `:3736`. `readConversationPage` falls back to the persisted row precisely when live has no turn (`:1260`), so a post-switch read cannot name the finished turn, and a pre-switch in-flight read is dropped by `scopeIsCurrent()` in `conversation.tsx`. `agent-event-bridge.tsx:233` already trusts the backend unguarded, on `main` as well as here.
- *Submitted prompts.* `presentedPromptResolutions` and `submittedPromptRequests` have no view consumers; they only suppress a prompt that is present in the coordinator's `pendingPrompts`, and that set is kept correct across a switch by `agent-event-bridge.tsx:187-193` while the bridge is still mounted, or by `onScopedEvent` (`dynamic-island.tsx:79` → `dynamic-island-coordinator.ts:182-189`), whose subscription sits above the keyed boundary and is never torn down. The duplicate-submission guard itself is server-side (`agent-service.ts:1481`).

Preserving either set across the boundary would reintroduce the cross-server memory this PR exists to remove, so I would rather see a counter-case than add it on suspicion. If one of the two paths above can be shown not to run, I will fix it.

NorbertBodziony:
Round four, four findings: three valid and fixed in `975b4f2`, one I am answering again with better evidence than I gave the first two times.

**Submitted prompts reappear after a switch — valid, and a regression this PR introduced.** You are right and my earlier answer to the equivalent point was wrong. On `main` `submittedPromptRequests` was global and nothing cleared it, so the window between answering a prompt and the snapshot that proves it was answered could outlast a server switch. Scoping it to `turns.tsx` ended that window at the switch, and a `runtime-snapshot` arriving after the user came back reached `reconcileAttentionPrompts` with an empty marker map (`agent-event-bridge.tsx:242`), so the answered prompt was offered again.

The coordinator cannot stand in for it: `replaceServer` suppresses via `resolvedPrompts` but deletes the marker as soon as the scope stops reporting the prompt as pending (`dynamic-island-coordinator.ts:53-57`) — exactly when the scope is being torn down.

Both markers now live in a new `answered-prompts.tsx` above the keyed boundary, keyed by server id and then bot id, re-exported from `useTurns()` under their own names so no consumer changes. Keying by server as well as bot is deliberate rather than defensive: ids are `bot-${uuid}` (`backend/bot-store.ts:182`), but `isValidBotId` (`:770`) admits any slug, so this does not have to rely on global uniqueness. New case in `App.servers.test.tsx`, watched failing against the unfixed `turns.tsx`: `expected document not to contain element, found <input`.

**Transient composer state hoisted above the boundary — valid, fixed.** `attachmentBusy`, `composerError`, `submitting` and `selectionSending` move into `createServerConversationState`. Unlike everything else in the stable half they carry no key at all, so a send left in flight on the server being left disabled the arriving server's composer and showed that server's failure under it. `conversationErrors` is keyed and stays where it was. Watched failing: `Received element is not enabled`.

**Island actions bypass the load gate — valid, fixed.** An action naming the server that is already active now defers to the same `pendingIslandAction` handoff the cross-server branch uses, instead of running `selectBot` against an empty bot list. The readiness effect drains it once `loaded()` turns true; `publishedAction` stays untouched, because nothing disposes this bridge in the meantime and it is the one that consumes its own entry.

**Completed turns resurrected by the seed — not real.** Two things have to be true for this, and neither is. Main clears and persists `activeTurnId` (`agent-service.ts:3693`, `openbot-database.ts:921`) *before* it emits `turn-completed` (`:3736`), so a post-switch read cannot name a finished turn. And in the renderer, `dynamic-island.tsx:78` subscribes `onScopedEvent` **above** the keyed boundary, while `dynamic-island-coordinator.ts:164-176` nulls `runtime.activeTurns[event.botId]` on `turn-completed` for *any* server, active or background. The coordinator's snapshot is the only thing `turns.tsx` seeds from, so it cannot hand a finished turn to a fresh scope, and `completedTurnByBot` — whose whole job is to suppress a turn id the snapshot still reports as active — starting empty is correct.

Verification on the fixes: `vitest run src/renderer` → 66 files, 642 tests green; `biome check src/renderer/src` and `tsc -p tsconfig.web.json` clean.

NorbertBodziony:
Round five: the one remaining finding was right, and it was right about the part I got wrong — fixed in `58b6f75`.

**[P2] Submitted prompts reappear after switching away and back.** Valid. Making `submittedPromptRequests` outlive the scope closed the snapshot entrance and left the seed open, and the seed is the one that does not need main to say anything.

I measured it before fixing it, because the previous round's answer deserved a check: a temporary probe on the seed prints, on the return leg,

```
SEEDPROBE local {"chief":{"type":"prompt","requestId":"prompt-across-servers",…}}
```

The reason is the one you named. `submitPromptAnswers` (`turns.tsx:137`) records the marker and leaves `pendingPrompts` untouched — what hides the questions is `QuestionPromptBubble`'s own local state, which the keyed remount throws away — and the Dynamic Island coordinator's memory is that same `pendingPrompts` played back through the projection effect (`dynamic-island-bridge.tsx:66-89`). Its `resolvedPrompts` markers only come from `agent-input-resolved` and island actions, so an answer given in the UI never reaches them.

`seededAttentionPrompts` (`agent-runtime-snapshot.ts:62`) now filters the seed against both markers, next to `reconcileAttentionPrompts`, which does the same against main's snapshots — one home for prompt suppression, one entrance each. Browser takeovers are deliberately not filtered: `respondToBrowserTakeover` clears `pendingPrompts` on success, so the projection above has already dropped them.

Your point about the test was the sharper half. The case now waits on the composer, which `ConversationView.tsx:3471` hides for as long as anything is being asked, so a settled prompt-free workspace is a positive observable and the assertion lands in the interval before any snapshot. Watched failing against the unfiltered seed: `expected document not to contain element, found <input`, with no snapshot emitted at all.

`biome check src/renderer/src` clean over 274 files, `tsc -p tsconfig.web.json` clean, `vitest run src/renderer` 66 files / 642 tests green.

Issue #174, opened 2026-09-02
Test and structure improvements from a comparison with anomalyco/opencode


NorbertBodziony:
Follow-up to #171, same exercise against a different repository: [`anomalyco/opencode`](https://github.com/anomalyco/opencode) — a Bun monorepo of 34 workspace packages, ~515k production LOC, shipping roughly every two days.

Measured on branch `dev`, HEAD `b046973` (v1.18.26). The clone is truncated at 229 commits (2026-08-14 → 2026-09-02), so the commit-habit figures carry that caveat. Everything else is a full-tree measurement.

## Baseline

Same methodology as #171 — tracked files only (`git ls-files`), `.test.` suffix marks a test file — so the openbot column reproduces the numbers published there.

|  | opencode | openbot |
| --- | ---: | ---: |
| tracked files | 6,611 | 1,007 |
| prod LOC / files | 515,145 / 2,648 | 129,734 / 543 |
| test LOC / files | 165,697 / 675 | 50,233 / 192 |
| test cases | 5,952 | 1,408 |
| **prod LOC per test case** | **87** | **92** |
| test files using module mocks | 33 | **0** |
| accessible role / name / focus assertions | **0** | **1,477** |
| snapshot assertions | 25 | 0 |
| prod median / p95 file LOC | 75 / 859 | 73 / 898 |
| prod files > 1,500 LOC | 24 (0.9%) | 15 (2.8%) |
| largest hand-written prod file | 3,465 | **6,056** |
| root `package.json` scripts | 13 | 96 |
| `AGENTS.md` files | 18 (1,261 lines) | 1 (+1 mobile) |
| GitHub workflows | 26 (2,547 lines) | 4 (1,694 lines) |
| composite actions | 2 | 0 |

The headline is the fifth row: **test density is effectively identical.** Their suite is 4× larger because their codebase is 4× larger. Their raw test count is not evidence of a stronger practice, and several rows below suggest the opposite. What they are genuinely better at is *structure* — how the same amount of code is arranged, and how much of that arrangement is machine-enforced rather than written down.

---

### 1. A fresh install and a migrated install can silently produce different schemas

`src/backend/openbot-database-schema.ts:278`:

```ts
const LATEST_SCHEMA_SQL = BASELINE_V8_SCHEMA_SQL;
```

and `createLatestDatabase` at line 418:

```ts
db.exec(LATEST_SCHEMA_SQL);
[...]

PR #185, opened 2026-09-03
test: guard the two database build paths and the IPC mirror, and put shared versions in one catalog

NorbertBodziony:
Closes the three items from #174 that earned their place now: two boundaries AGENTS.md already
calls mandatory but nothing checked, and one version-drift class no test could catch. Kept as one
PR at the developer's request.

Model: Claude Opus 5, via Claude Code in a git worktree. No renderer UI, so no before/after
screenshots apply.

## 1. Schema parity between the two build paths

`src/backend/openbot-database-schema-parity.test.ts`

`migrateOpenBotDatabase` has two paths. A new install execs `LATEST_SCHEMA_SQL` and stamps
migrations 9/10/11 as applied **without running them** — correct today only because all three are
the data-only `refreshProviderSessionsForDynamicTools`. An existing database runs the migrations.
Nothing but the comment at the top of the file kept the two in agreement, so the first DDL
migration ≥ 12 not hand-mirrored into `LATEST_SCHEMA_SQL` would ship new installs a database
missing a column while upgraded installs get it — on users' machines, with no backup.

The test builds both and compares columns, indexes, uniqueness, foreign keys, the normalized DDL
(which catches what the PRAGMAs cannot express, notably the partial index on `routine_runs`), and
the recorded migration versions.

No production file changed. The module still exports only `migrateOpenBotDatabase` — the "upgraded"
fixture is a database containing *only* an empty `schema_migrations` table, which makes
`hasExistingSchema` true while `latestAppliedVersion` returns 7, so every migration runs from the
v8 baseline upwards.

Worth recording why not the obvious fixture: building the migrated database by deleting
`schema_migrations` rows above 8 from a fresh one is wrong. A *correct* future migration 12 would
then hit "duplicate column", because `LATEST_SCHEMA_SQL` already added it — the test would go red
on correct code.

## 2. IPC channel coverage

`src/main/ipc-channel-coverage.test.ts`

189 channels in `packages/contracts/src/ipc-channels.ts` are mirrored by hand in the main process
and in the preload, and only `tsc` links them — incompletely: a channel the preload invokes with no
main handler type-checks and rejects at runtime, and a handler nobody calls is dead trust-boundary
surface. The test proves both sides cover every declared channel, disjointly across request and
event mode, and that the two sides agree on which is which. Today: 171 request + 18 event = 189.

Neither side can be imported — both call into Electron at module scope and export nothing — so the
test reads source text, the same shape as `scripts/anti-slop-rules.test.ts`. That is safe because a
channel name is never written as a literal on either side, and the first assertion is what keeps it
safe: every `IPC_CHANNELS.` reference must be attributed to a known call, so a new helper or an
unrecognised call shape fails loudly instead of silently shrinking the compared sets. The scan also
blanks comments and string literals, so a commented-out registration does not read as a
registration.

One correction to what the issue implies: the scan has to cover **all** of `src/main` recursively,
not just `index.ts` and `ipc/`. `dynamic-island-window.ts` holds 8 event channels, and without them
the two sides do not match.

## 3. One catalog for shared dependency versions

`typescript` was `5.9.3` in three workspaces and `7.0.2` in the rest; `zod` was `4.4.3` for desktop
and `remote/api` but `4.5.4` for mobile. `bun.lock` confirmed both as real duplicate installs.

Root `package.json` gains a `workspaces.catalog` with the 24 dependencies declared in two or more
manifests, and every one of them becomes `"catalog:"` at the call site. That rule deliberately
excludes everything Expo owns — the `expo*`, `react`, `react-native` and friends pins only ever
appear in `apps/mobile`, and a `catalog:` specifier is opaque to `expo install` / `expo-doctor`.
`@openbot/*` deps keep `workspace:*`. `zod` unifies **upward** to `4.5.4`, per the developer's
decision.

New root `bunfig.toml` sets `[install] exact = true` so `bun add` stops writing carets. It is read
from the cwd, so it applies to installs run from the repo root; `apps/mobile`'s caret style stays
Expo's to manage.

**`solid-js` is deliberately not in the catalog.** The first push proved why: `storybook-solidjs-vite`
resolves the Solid major from the *declared* version string to choose its renderer entry, so
`"catalog:"` failed the Storybook build with `Could not detect Solid version: solid-js is not
installed.` A catalog specifier only means something to bun — anything that reads the version itself
sees a value it cannot parse. It keeps its literal `2.0.0-rc.0` in all three manifests.

`scripts/dependency-catalog.test.ts` keeps the catalog from decaying into prose, with five
assertions: no catalogued dependency may be declared with a literal version anywhere; **no external
dependency two manifests share may be missing from the catalog**; no catalog entry may have fewer
than two users; the dependencies a build tool reads must stay *out* of the catalog, with the reason
carried in the failure message; and those same dependencies must still be pinned to one version
everywhere, since the catalog is not there to do it for them. It expands the workspace globs from the
root manifest, so a workspace added later is covered without editing the test.

The second assertion is the one that came out of review — the others all walk the catalog's own keys,
so on their own they say nothing about the *next* shared dependency to arrive. Two workspaces could
both `bun add yaml` with literal versions and stay green, recreating exactly the drift this PR
removes. It exempts `workspace:` specifiers and the names below.

**Lockfile:** `bun.lock` changed no resolved version other than the two intended collapses —
`zod@4.4.3 → 4.5.4` with the nested `@openbot/mobile/zod` copy removed, and both
`typescript@5.9.3` entries gone. Verified by extracting every resolution from the lockfile before
and after and diffing; that diff is exactly four lines. `uniwind`'s transitive `tailwindcss@4.3.2`
and `@expo/cli`'s `zod@3.25.76` are untouched.

**Adjacent one-liner, not the topic:** root `lucide-solid` moves from `^1.33.0` to `1.33.0`. It is
patched at `1.33.0` and `patchedDependencies` keys on `name@exactversion`, so the caret meant a
patch release would silently un-patch it. `bun.lock` already resolved 1.33.0, so this changes no
resolution. Single-consumer, so it does not go in the catalog. `@kobalte/core` is pinned in the
catalog at exactly `2.0.0-alpha.0` for the same reason.

## Watch it fail

Run for every assertion, per AGENTS.md test rule 2.

**Schema parity** — appended a migration 12 running
`ALTER TABLE projection_agents ADD COLUMN nickname TEXT` plus an index and a partial index, without
touching `LATEST_SCHEMA_SQL`. Tests 1 and 2 went red naming `nickname`, `agents_nickname`,
`WHERE nickname IS NOT NULL` and the changed `projection_agents` DDL. Test 3 correctly stayed green:
versions are the part `createLatestDatabase` fakes *consistently*. Reverted.

**IPC coverage** — deleted one `handleTrusted(IPC_CHANNELS.getSetupState, …)`. Two assertions went
red with `getSetupState` named in the diff, not "expected 171, got 170". Then renamed that call to
an unknown helper: the completeness guard reported
`src/main/index.ts: registerViaSomeNewHelper(IPC_CHANNELS.getSetupState)`. Then commented the
registration out instead of deleting it, confirming the comment stripper is load-bearing — it read
as unregistered. Reverted each time.

**Dependency catalog** — put a literal `"vitest": "4.1.10"` back in `packages/contracts` and added a
single-user catalog entry: both assertions went red naming the manifest, the field and the
dependency. Then added an uncatalogued `nanoid` to two manifests at two different versions: the new
reverse-direction assertion went red with
`nanoid: 5.1.1 in packages/contracts/package.json, 5.1.0 in packages/team-client/package.json`. Then put `solid-js` into the catalog and gave `apps/auth-api` a different `2.0.0-rc.1`:
the exclusion assertion went red carrying the whole Storybook reason into the message, and the
pinning assertion reported `2.0.0-rc.0 in package.json, packages/brand/package.json vs 2.0.0-rc.1 in
apps/auth-api/package.json`. Reverted each time.

## Surfaces touched

Desktop main and preload, `src/backend`, `packages/contracts` (read only — no contract change), and
every workspace manifest: mobile, public web (`apps/auth-api`), `site-router`, `team-client`,
`brand`, `contracts`, `remote/api`. No renderer UI, no palette token, no migration, no Team API wire
protocol change, so no `PRIVACY.md` or `docs/ARCHITECTURE.md` change either. The `README.md` command
table is unaffected — no script was added or renamed.

## Verification

Narrow, per AGENTS.md — CI owns the full suite.

- `bunx vitest run` on each of the three new files — 14 tests green
- `bun run test:backend` — 27 files, 258 green, 2 skipped
- `bun run typecheck:sites`, `typecheck:team-client`, `--cwd remote/api typecheck`, `typecheck:api`,
  `desktop:typecheck`, `mobile:typecheck` — all clean. The three workspaces that moved to
  TypeScript 7.0.2 were confirmed clean under it before the change too.
- `bun run --cwd remote/api test` — 19 green, covering the `zod` bump there
- `biome check` on the touched paths only — clean

## Two things worth a look

`packages/brand` now carries `catalog:` in its `peerDependencies`. Bun resolves it and the package
is private, but if it is ever published that has to become a literal range again.

The `solid-js` exclusion is the general shape of a risk this PR takes on: any tool that reads a
declared version string rather than asking bun will choke on `catalog:`. Storybook was the one CI
caught. `expo-doctor` is the other candidate — it is not in CI, and `bun run mobile:doctor` is worth
a look before the next mobile release, though nothing Expo owns was catalogued precisely for this
reason.


NorbertBodziony:
**Finding 1 (P2, duplicate main handlers) — accepted and fixed** in `e543fb91`.

Correct: `handleTrusted` wraps `ipcMain.handle`, which throws on a second registration for the
same channel, so a channel registered in two files crashes the app on every launch while the set
comparisons stay green. Added a sixth assertion that counts handler registrations per channel and
names every site of a duplicate. Set semantics are kept for sends and subscriptions, where
multiplicity is legitimate, as the finding says.

Watch-it-fail: duplicated `handleTrusted(IPC_CHANNELS.getSetupState, …)` in `src/main/index.ts`.
Only the new assertion went red, with
`getSetupState: src/main/index.ts handleTrusted, src/main/index.ts handleTrusted`. Reverted.

**Finding 2 (P3, duplicated database test boundary) — I think this one is wrong.**

The two files test different modules. `openbot-database.test.ts` imports `OpenBotDatabase` from
`./openbot-database` and never references the schema module — `grep -c openbot-database-schema
src/backend/openbot-database.test.ts` returns 0. `openbot-database-schema-parity.test.ts` imports
`migrateOpenBotDatabase` from `./openbot-database-schema`, which has no test file today. So the
boundary it covers does not exist yet in the suite, which is the condition AGENTS.md sets for a new
test file, and the file is named after the module it tests. Folding a test for
`openbot-database-schema.ts` into `openbot-database.test.ts` would be the surprising placement, not
the other way round.

The consequence is different too. `openbot-database.test.ts` asserts what the database does once it
is open. This asserts that the two *construction paths* agree — `createLatestDatabase` for a new
install versus running migrations 8→11 for an existing one — which is a property of neither path
alone and has no assertion anywhere today.

On the shared infrastructure: what actually repeats is four lines of `mkdtemp` and an `afterEach`
that removes the roots. The 231 lines are almost entirely the schema-inspection helpers
(`PRAGMA table_info` / `index_list` / `index_xinfo` / `foreign_key_list` and the DDL normalizer),
which have exactly one consumer. Merging would take `openbot-database.test.ts` from 1220 to ~1451
lines to deduplicate an `mkdtemp` call, against a repository that split its test monoliths in
c66e995a.

/norbiai review

NorbertBodziony:
**Finding 1 (P2, "IPC scan never validates the actual channel argument") — the premise is wrong.**

The finding rests on the claim that `withoutCommentsAndLiterals` "blanks both string contents and
quote delimiters". It blanks the contents only; the delimiters survive at their original offsets.
`src/main/ipc-channel-coverage.test.ts:281-285`:

```ts
if (character === '"' || character === "'" || character === "`") {
  index += 1;                                                        // steps past the opening quote
  blankUntil((position) => characters[position] === character, true); // blanks the inside
  index += 1;                                                        // steps past the closing quote
  continue;
}
```

`index` is advanced past the opening quote *before* `blankUntil` runs, and `blankUntil` stops *at*
the closing quote, so neither delimiter is ever assigned `" "`. That is deliberate: the whole
function keeps every character at its original offset, and it is what makes `/^["'`]/` work.

I ran the finding's own counter-example rather than argue it. Adding
`handleTrusted("existing-wire-value", () => readSetupState(setupFile));` beside the legitimate
`getSetupState` registration in `src/main/index.ts` does **not** leave every assertion passing:

```
× names every channel through IPC_CHANNELS rather than a string literal
  - []
  + [ "src/main/index.ts: handleTrusted takes a string channel" ]
```

The same check was already demonstrated on three shapes before this commit landed, including the
second-argument position and the preload side:

- `handleTrusted("undeclared:channel", …)` → `src/main/index.ts: handleTrusted takes a string channel`
- `sendToRenderer(mainWindow, "agent-event", …)` → `src/main/index.ts: sendToRenderer takes a string channel`
- `ipcRenderer.on("dynamic-island-preference", handler)` → `src/preload/index.ts: ipcRenderer.on takes a string channel`

On the suggested replacement — "require the channel-position argument to be a direct `IPC_CHANNELS`
member" — that is stricter than the code allows, and would be a false positive. The preload's
forwarding helpers pass a variable: `src/preload/index.ts:93` is
`return ipcRenderer.invoke(channel, request).then(decoder)`. Rejecting every non-`IPC_CHANNELS`
argument would flag that legitimate line. The helper itself (`invokeAgent`, `invokeAgentForServer`)
is in the callee list, so its own call sites are what get checked, which is the correct level.

Comparing "resolved wire values" is also not available to a static scan, and the duplicate-registration
assertion added in `e543fb91` already covers the crash-on-startup consequence the finding names, for
every channel the scan can see.

/norbiai review

PR #191, opened 2026-09-03
Set up Bun through one composite action, and split main-process IPC by domain

NorbertBodziony:
Closes the remaining CI and structure items from #174.

## `ci: set up Bun and Node through one composite action`

Ten call sites across `ci.yml` and `release.yml` each carried their own `oven-sh/setup-bun`, `actions/setup-node` and `bun install` lines. `.github/actions/setup-bun` is now the one copy: `install` is `frozen` / `ignore-scripts` / `none`, `node` takes an expression so a job can set Node up on one platform only, and both are validated in the action because a typo would otherwise silently drop `--ignore-scripts`. Node always comes from `.nvmrc` — there is no version input, since an overridable one is the same drift one level down. Net −98/+80.

`remote-desktop-runtime.yml` is deliberately **not** converted. Its toolchain is part of a versioned build recipe: the workflow makes a change to itself require a `remoteDesktop.recipeVersion` bump, so routing it through a shared action would move the Bun and Node versions somewhere that check cannot see, and a later edit to the action would change the recipe silently. The issue counted four workflows; three is the right number, and the fourth is now documented as an exemption rather than an oversight.

## `refactor(main): register IPC per domain, and move guidance next to what it guards`

`src/main/index.ts` was 2,697 lines with 120 `handleTrusted` registrations inline, interleaved with window and lifecycle code. It is now 1,627, and `registerIpcHandlers` is a dispatcher — sixteen `register*IpcHandlers` calls against fifteen new files under `src/main/ipc/`, following the shape `register-team-handlers.ts` established for one domain and stopped at. Registrations, input parsers and route helpers moved unchanged.

The largest is `register-agent-handlers.ts` at 431 lines: bots, the conversation and the queue, which are one surface. Memories, routines and attachments came out into their own registrars — each a closed sub-surface with its own parsers and no shared state with the rest.

`registerIpcHandlers` now takes a named `IpcHandlerDependencies` object instead of 21 positional parameters. Three were `setupFile`, `analyticsPreferenceFile` and `updatePreferenceFile` — same type, so swapping two compiles cleanly and writes one preference into another's file.

Root `AGENTS.md` went 279 lines to 202. Renderer UI, migrations and Team API compatibility moved to `src/renderer/`, `src/backend/` and `packages/contracts/`, which load when you open a file under them. The D1 deploy-race rule went to a new `apps/auth-api/AGENTS.md` — `src/backend` is not an ancestor of `apps/auth-api/migrations/` and would not have loaded for the change it governs. `src/main/AGENTS.md` is new and carries the rule the fifteen module headers cite.

Each of the five gains a **waiting table**, which is what Tests rule 4 has been missing: the rule says a test needing a timeout is wrong but never said what to use instead, so a rejected `setTimeout` becomes a longer `vi.waitFor` timeout rather than a real barrier.

## Tests

**No test was added.** Four properties that look like they need one are already mechanical, and Tests rule 2 was run on each — every one was watched failing:

| Property | What catches it | Observed failure |
| --- | --- | --- |
| A moved block left behind in both files | `ipc-channel-coverage.test.ts` | duplicated one `agentListMemories` registration → `1 failed \| 10 passed`, naming both paths |
| A bare `ipcMain` outside the wrapper | same file | pre-existing assertion, unchanged |
| A registrar never called from `index.ts` | `tsc` | `TS6133` — the call site is the import's only use |
| A method missing from / extra in `mock-openbot.ts` | `tsc` | `TS2741` / `TS2353`, both annotated `: OpenBotDesktopApi` |

The first is what makes this refactor safe: 47 agent channels moved across four files and the coverage test proves each is still registered exactly once through the trusted wrappers.

The fourth is #174 item 6 — the issue itself files generating that mirror as a separate decision, and under Tests rule 3 `tsc` covering both directions is the end of it. What the type checker cannot cover is the *behaviour*, and that is written down in `packages/contracts/AGENTS.md`.

## Surfaces touched

Main process and the IPC registration surface only. **No contract changed** — no channel added, removed or renamed, no parser altered — so `packages/contracts`, `src/preload`, `src/renderer` and `mock-openbot.ts` are untouched. Mobile, public web, the three palettes, migrations and `PRIVACY.md` are not involved. Documentation is the AGENTS.md set; the `README.md` command table and `docs/ARCHITECTURE.md` needed no change.

## Verification

`bun run typecheck:node` clean. `bun run test:desktop -- src/main` → 43 files, 339 tests passed. `biome check` clean on every touched path. Repo-wide checks left to CI.

### DIFF
--- a/.github/actions/setup-bun/action.yml
+++ b/.github/actions/setup-bun/action.yml
@@ -1,0 +1,52 @@
+name: Set up Bun
+description: >-
+  Install the repository's pinned Bun and Node and, unless the job installs
+  later, its dependencies. Node is not optional: every caller builds, packages
+  or tests something that needs it. Checkout is the caller's job: a composite action in
+  this repository cannot run before the repository is on disk.
+
+  `remote-desktop-runtime.yml` deliberately does not use this action. Its
+  toolchain is part of a versioned build recipe: the workflow makes a change to
+  itself require a `remoteDesktop.recipeVersion` bump, and an edit here would
+  change that recipe without tripping the check.
+
+inputs:
+  install:
+    description: >-
+      `frozen` for `bun install --frozen-lockfile`, `ignore-scripts` to add
+      `--ignore-scripts` (no native build, for jobs that only typecheck or test),
+      or `none` for a job that must install later - after a secret check, a cache
+      restore, or a platform toolchain.
+    default: frozen
+
+runs:
+  using: composite
+  steps:
+    # A typo in `install` would otherwise silently drop `--ignore-scripts` or
+    # skip the install, and the job would fail somewhere less obvious.
+    - name: Validate inputs
+      shell: bash
+      env:
+        INSTALL: ${{ inputs.install }}
+      run: |
+        case "$INSTALL" in
+          frozen | ignore-scripts | none) ;;
+          *) echo "install must be frozen, ignore-scripts or none (got '$INSTALL')." >&2; exit 1 ;;
+        esac
+
+    - name: Set up Bun
+      uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+      with:
+        bun-version: 1.4.0
+
+    - name: Set up Node
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+      with:
+        # `.nvmrc` and nothing else. An overridable version input is the drift
+        # this action exists to remove, one level down.
+        node-version-file: .nvmrc
+
+    - name: Install dependencies
+      if: inputs.install != 'none'
+      shell: bash
+      run: bun install --frozen-lockfile ${{ inputs.install == 'ignore-scripts' && '--ignore-scripts' || '' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,17 +29,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Set up Bun
-        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
-        with:
-          bun-version: 1.4.0
-
-      - name: Set up Node
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
-        with:
-          node-version-file: .nvmrc
-
-      - name: Install dependencies
-        run: bun install --frozen-lockfile
+        uses: ./.github/actions/setup-bun
 
       - name: Check desktop application
         run: bun run check:desktop
@@ -56,17 +46,9 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Set up Bun
-        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+        uses: ./.github/actions/setup-bun
         with:
-          bun-version: 1.4.0
-
-      - name: Set up Node
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
-        with:
-          node-version-file: .nvmrc
-
-      - name: Install dependencies
-        run: bun install --frozen-lockfile --ignore-scripts
+          install: ignore-scripts
 
       - name: Run desktop tests
         run: bun run test:desktop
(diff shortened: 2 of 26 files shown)


NorbertBodziony:
Findings 1, 3 and 5 are fixed, 4 is fixed differently, and 2 is a deliberate decision rather than a missing commit.

**1 + 3 — runtime workflow.** These are the same defect seen from both ends, and 3 is the decisive one: routing that build's toolchain through a shared composite moves the Bun and Node versions somewhere the recipe-version check cannot see, so a later edit to the action would change the build recipe silently and a stale immutable runtime would be reused. Bumping `remoteDesktop.recipeVersion` would have cleared the red without fixing that, and would have republished a runtime whose inputs did not change. `.github/workflows/remote-desktop-runtime.yml` is now byte-identical to base and keeps its own pins; the action documents the exemption next to them.

**5 — `node-version` input.** Removed. `node-version-file: .nvmrc` unconditionally.

**4 — D1 guidance.** Correct that `src/backend/AGENTS.md` is not an ancestor of `apps/auth-api/migrations/`. Rather than returning it to the root, it moved to a new `apps/auth-api/AGENTS.md`, which is an ancestor and loads for the change it governs. `src/backend/AGENTS.md` keeps one line pointing at it and naming the contrast — irreversible-with-no-backup there, reversible-but-raced here. Root's pointer table has the fifth row.

**2 — repo-wide guard.** Not a lost commit. `scripts/repo-wide-guard.ts` existed on an earlier head of this branch and was removed on the repository owner's explicit decision, along with the `package.json` indirection and the AGENTS.md paragraph describing it. AGENTS.md:29 as it stands claims only that the ban is a rule, not that anything mechanizes it, so there is no gap between the prose and the code. Restoring the guard is a product decision that was made the other way, not a defect in this PR.

/norbiai review

NorbertBodziony:
Finding 2 is fixed. Finding 1 is accurate about the rule but is not worth acting on here.

**2 — unused `node` toggle.** Correct, and it was dead as of the previous push rather than speculative from the start: `node` existed only so `remote-desktop-runtime.yml` could set Node up on Windows alone, and that workflow was reverted to base one commit ago. Nothing has passed the input since. The input, its `NODE` validation branch and the step `if` are gone; Node is now always installed from `.nvmrc`, and the action is down to one input. Net −14 lines against the green head, touching nothing else.

**1 — one topic per PR.** The rule is real and this PR does need the word "and". Splitting it is not free, though: the two changes are already separate, self-contained commits that revert independently, `CI` and this review are green on the pair as they stand, and splitting would open a second PR and discard that result to re-run the same checks in two halves. The cost the rule exists to prevent — one change blocking or forcing reversion of the other — is already paid for by the commit boundary here. The repository owner has looked at this and decided to keep it as one PR; noting it so the next reader knows it was a decision rather than an oversight.

/norbiai review

NorbertBodziony:
Finding 1 is correct and I had it wrong. Fixed with a test.

Removing the call **and** its import compiles: I verified it by deleting both for `registerMemoryIpcHandlers`, and `typecheck:node` exits 0. So does a registrar nobody ever wired up in the first place. `tsc` only ever covered the half where the import is left behind, and both this PR's commit message and `src/main/AGENTS.md` claimed more than that — the paragraph is corrected, not just the code.

`ipc-channel-coverage.test.ts` gains "calls every registrar under src/main/ipc exactly once from the dispatcher": it reads every `export function register*IpcHandlers` under `src/main/ipc/` and counts call sites in `src/main/index.ts`, failing on 0 (orphaned or never wired) and on more than 1. Watched fail on the break above — `registerMemoryIpcHandlers (src/main/ipc/register-memory-handlers.ts) is called 0 times by src/main/index.ts`, with `tsc` green alongside it. It also asserts it found more than ten registrars, so a regex that goes blind is red rather than vacuously green, which is the failure mode `AGENTS.md` calls out for the anti-slop rules.

This is squarely Tests rule 7 — splitting index.ts is what creates the unwired state, so the guard belongs in the same change.

Finding 2 stands as the repository owner's decision to keep this as one PR; no further argument from me.

/norbiai review

NorbertBodziony:
Finding 2 is fixed. Also noting why the last run went red — it was not a finding.

**2 — repeated registrar boilerplate.** Correct, and the last round is itself the evidence: `src/main/AGENTS.md` had to be corrected about what `tsc` catches, and the fifteen copies only escaped drifting because they happened not to restate that part. The three-line policy block is gone from all fifteen files; each keeps its domain summary and nothing about policy. `src/main/AGENTS.md` loads when you open any file under `src/main`, so the pointer was telling the reader something they already had. −60 lines.

**The `NorbiAI PR Review` failure on `e9c8e6c2` was infrastructure, not findings.** The publish step hit `dial tcp 140.82.121.5:443: i/o timeout` posting to the issues API, and the gate blocks when it cannot confirm publication — which is the right call for that gate. The POST did land server-side: the comment is on the PR timestamped 12:39:09, and the client timeout is at 12:39:42. The review itself was `REVIEW_STATUS: success` with no P0 or P1 findings.

Head is now `df7b8525`. `typecheck:node` clean, `test:desktop -- src/main` 43 files / 340 tests.

/norbiai review

PR #192, opened 2026-09-03
Add EML attachment imports

### LINKED ISSUE #114: Fix attachment import failures for EML and ZIP files
## Summary

OpenBot reports that `.eml` and `.zip` attachments are supported, but the attachment import flow rejects them with:

`Error invoking remote method 'agent:import-attachments': Error: .eml is not supported.`

The user-facing guidance and the actual `agent:import-attachments` capability are inconsistent.

## Steps to reproduce

1. Ask OpenBot to open or analyze an `.eml` file, or inspect a `.zip` archive.
2. Attach the file through the chat attachment flow.
3. Observe the import error.

## Expected behavior

- Supported file types are accepted by `agent:import-attachments` and passed to the agent for analysis.
- For `.zip` files, extraction is bounded and safe, with clear handling for nested archives, unsupported files, oversized entries, and password-protected archives.
- If a format is not supported, the UI does not claim that it can be analyzed and shows a precise, actionable error.

## Acceptance criteria

- [ ] `.eml` attachments can be imported and their message content, headers, and attachments are available for analysis.
- [ ] `.zip` attachments can be imported with safe extraction limits and clear failure states.
- [ ] The advertised attachment support list matches the formats accepted by the backend.
- [ ] Import errors identify the unsupported format and the next supported action.
- [ ] Tests cover successful imports, malformed files, oversized archives, and unsupported formats.

## Reference

Attachment import error

Sniezka1927:
## What changed

- Expand EML messages into readable headers, text, HTML, and supported embedded attachments.
- Bound EML parsing by header size, MIME-part count, nesting depth, attachment count, and attachment byte limits.
- Route local and remote EML imports through the same normalization boundary, including rollback of partial remote draft uploads.
- Advertise EML in the shared attachment picker contract. ZIP remains unsupported and is outside this PR scope.

Fixes #114

## Verification

- [x] Narrow tests: 89 passed across EML imports, attachment contracts/previews, remote upload rollback, and IPC coverage/parsers
- [x] All Node type checks passed
- [x] Targeted Biome checks and git diff check passed
- [x] GitHub Check, Tests, API, Surfaces, Storybook, and Cloudflare preview jobs passed
- [ ] Integrated bun run dev smoke test: blocked because the fresh worktree has no .env.keys; no packaged fallback was used

## Risk and security impact

This changes local EML ingestion. Messages are expanded only after bounded MIME preflight. Unsafe embedded paths, nested EML attachments, unsupported embedded formats, oversized content, excessive totals, and excess attachment counts fail before storage or remote upload.


Sniezka1927:
ZIP support was intentionally removed at the requester’s direction. The PR title, body, changelog, picker contract, implementation, and tests now define the scope as EML-only; ZIP remains unsupported. Please re-evaluate the P1 against the updated PR scope. /norbiai review

Sniezka1927:
The P1 premise does not match postal-mime 2.7.6. MimeNode.processHeaders iterates headerLines from bottom to top and assigns contentType only while contentType.default is true, so the physically last Content-Type wins. readEmailHeaders also scans top-to-bottom and overwrites contentType, so it selects the same physically last value. I verified the reported multipart-first/text-plain-second case directly: PostalMime returned 0 attachments and treated the --openbot lines as text instead of building MIME nodes. There is no preflight/parser disagreement for duplicate Content-Type headers. /norbiai review

Sniezka1927:
The remaining P2 points to pre-existing code that this PR no longer changes: after c708b08, src/main/ipc/register-attachment-handlers.ts and src/main/remote-server-manager.ts are byte-for-byte identical to origin/main, and the PR diff is limited to the EML whitelist/MIME mapping, changelog, and focused tests. Raw EML support uses the existing single-file attachment pipeline and does not introduce or alter concurrent remote upload semantics. An atomic batch endpoint or client-generated draft IDs would require a separate Team API capability/protocol change, outside this PR. Please withdraw this baseline finding from PR #192. /norbiai review

PR #203, opened 2026-09-03
refactor(contracts): move IPC response decoders beside the contract

NorbertBodziony:
Stacked on #202. Pure relocation plus new boundary tests.

- New packages/contracts/src/ipc-decoders.ts (~60 pure response decoders, no Electron), exported via the ipc barrel and a dedicated ./ipc-decoders subpath.
- src/preload/index.ts 1199 -> 506 lines: wiring and DOM glue only.
- New ipc-decoders.test.ts: 9 accept/reject tests per decoder family. Fail-watched (broke decodeVoid, went red for the right reason, restored).
- No decoder logic changed; preload behavior identical.

Verification: contracts suite 118 pass, typecheck:contracts/node/renderer clean, biome clean. No biome-ignore, ts-expect-error, or any-widening added.

PR #204, opened 2026-09-03
refactor(main): register IPC handlers per domain outside the entry

NorbertBodziony:
Stacked on #203. Registration move, zero behavior change.

- src/main/index.ts 2697 -> 1587 lines. registerIpcHandlers is now a 6-call orchestrator.
- New src/main/ipc-handlers/: app, voice-auth, marketplace, agent (with the route*/upload* helpers that only registrations used), browser. Each takes an explicit deps interface, following the registerTeamIpcHandlers precedent; window access via a getMainWindow accessor.
- sendToRenderer channel typed as RendererIpcChannel (was string). renderer-ipc.test.ts now uses a real IPC_CHANNELS value instead of a dummy string; premise unchanged.
- Purged ~60 imports the move stranded in the entry.

Verification: typecheck:node clean, biome clean. Trust boundary preserved: ipc-channel-coverage (11), trusted-ipc (6), renderer-ipc (10), and src/main/ipc tests all pass (47 total). A generic subscribe() helper was tried and reverted because the coverage test forbids channel indirection by design. No biome-ignore, ts-expect-error, or any-widening added.


NorbertBodziony:
Superseded by #206. Entry split dropped: main shipped #200 which goes further.

PR #207, opened 2026-09-03
Isolate the local server per OpenBot account

### LINKED ISSUE #201: Isolate local servers between accounts and restore public server creation
## Problem

After logging out of one account and signing in with a different email address, OpenBot reuses the same local server that was created for the previous account.

This causes session and connection conflicts because the new account inherits or connects to server state belonging to another account. It also prevents creating a public server.

## Steps to reproduce

1. Sign in to OpenBot with account A.
2. Start or configure a local server.
3. Log out.
4. Sign in with account B using a different email address.
5. Observe that the local server from account A is still reused.
6. Try to create or configure a public server.

## Actual behavior

- The same local server is used across different accounts.
- Session and connection state can conflict or become invalid.
- Creating a public server is not possible.

## Expected behavior

- Local servers and their session, connection, and authentication state are isolated per account.
- Logging out clears the active account's server bindings and connection state from the client context.
- Signing in with another account creates or selects only that account's own server configuration.
- A user can create and use a public server independently of local server state.
- Existing valid configurations remain compatible after the fix, with a safe migration path where needed.

## Acceptance criteria

- [ ] Reproduced account-switching flow no longer reuses the previous account's local server.
- [ ] Server ownership and bindings are scoped to the authenticated account.
- [ ] Session and connection state cannot leak between accounts.
- [ ] Public server creation works after switching accounts.
- [ ] Logout, login, reconnect, and app restart flows are covered by automated tests.
- [ ] Clear error handling is provided when a server belongs to another account or is no longer available.

NorbertBodziony:
Fixes #201.

## The bug

`openbot-team-server-v1.json` held exactly one local host at its root, owned by whichever account ran `host:configure` first. Nothing re-bound that record when the signed-in account changed: `forwardCentralAuth` called `syncSignedInAccount(user)`, and `TeamStore.syncAccount` only refreshes profile fields on a member whose email already matches — it returns `false` for a stranger and leaves account A's host active.

So after signing out of A and into B, `HostStatus.configured` stayed `true` with A's `serverId`/`serverName`. `withLocalHostSummary` painted the rail's `local` row with A's name and `role: "owner"`; saving identity branched to `host:update-identity` → `assertOwnerAccount` → *"Sign in with the OpenBot email that created this host."*; `host:start` failed the same way, so B could not publish; and `host:configure` was unreachable behind *"The team server is already configured."*

## The fix

**`src/main/team-store.ts`** — the file now holds a `version: 2` envelope `{ version, activeAccountId, hosts: StoredTeam[] }`. `StoredTeam` itself is unchanged at `version: 1`, and `#state` is still a single `StoredTeam | null` — the *active* host — so `TeamApiServer`, `TeamWebRtcHostGateway` and every other `TeamStore` method are untouched.

- `initialize()` loads a v2 file as-is, migrates a bare v1 record into `hosts: [record]` and persists it, and still writes nothing for an unrecognised file. It then activates `activeAccountId`'s host — or, when that is null, the first host with no owner account, so an offline upgrade shows the host it showed yesterday. A host left behind by a sign-out always has an owner account, so sign-out still unbinds.
- `activateAccount(user)` binds by `owner.accountId`, then owner email (the existing v1 backfill path), then adopts an owner-less host — a password-`configure()`d one — and writes the binding so the next sign-in has something to match. No match leaves the store unconfigured rather than borrowing another account's host.
- `deactivate()` unbinds without removing. **`hosts` is append-only**: this file is user data with no backup, so a sign-out never destroys a host, its members, its invites or its logo.
- `configure` / `configureWithAccount` append to `hosts` and roll the push back if the write fails; the "already configured" guard is now per account.

**`src/main/host-service.ts`** — `syncSignedInAccount` becomes `applySignedInAccount(user | null)`: it stops the runtime when the account actually changes, activates or deactivates the store, and rebuilds `#status` from the newly active identity through a shared `initialHostStatus()` rather than patching it, so a second account cannot inherit the first one's `serverId`, name or `enabledOnLaunch`. Two subtleties: it short-circuits when the same account is reported again (a renamed profile or a new avatar no longer stops a host that is happily online), and `#startRuntimeOperation` now calls `syncAccount` instead of a full re-activation, which would stop the runtime it is starting.

**`src/main/index.ts`** — `forwardCentralAuth` calls `applySignedInAccount(null)` after `stop(false)` on `signed_out`, and `applySignedInAccount(state.user)` on `signed_in` before the `shouldAutoStartHost(status)` check, so B's own `enabledOnLaunch` governs. The startup path follows, and the development bootstrap activates the store right after `ensureDevelopmentAccount`.

Joined *remote* servers were already account-scoped by `#syncWebRtcHosts`, so they are untouched. Per-account `hiddenHostIds` and clearing the joined list on logout were explicitly left out of scope.

## Migration and downgrade

Forward: a shipped `version: 1` file is migrated in place on first `initialize()` and its host stays active. No host is ever removed.

Downgrade: an older build reading a `version: 2` file falls into the existing "unrecognised → no state" branch, so downgrading shows an unconfigured host. It does **not** keep that file safe, though — the released build persists an empty record over anything it did not recognise as soon as it signs an account in or configures a host, so a downgrade loses the v2 file. That is the pre-existing behaviour, not something this PR introduces, and it is the exposure of any format bump here; calling it out rather than pretending otherwise. This PR fixes that half going forward: a file this build cannot read is never written over, and configuring a server refuses outright with a message telling the user to move it aside.

## Surfaces touched

Desktop main process and the persisted team file. No IPC channel is added or changed, so `packages/contracts`, `src/preload/index.ts` and `src/renderer/src/preview/mock-openbot.ts` stay as they are — the renderer picks the status reset up through its existing `host.onEvent` subscription. No renderer, mobile, public web, palette, SQLite or `PRIVACY.md` change; nothing new leaves the machine. No UI change, so no before/after screenshots.

## Tests

Persisted state is a mandatory boundary and `TeamStore` is the lowest stable one here, so the coverage went into the existing `src/main/team-store.test.ts` (21 → 25 tests) rather than a new file:

- account B configuring a host leaves A's host, members, invites and logo intact, and re-activating A returns A's `serverId`, both members and the pending invite;
- signing out leaves nothing bound, and the next sign-in restores it;
- a host configured before accounts existed is adopted by the first account that activates, its owner is recorded, and it survives a restart;
- activating an account with no host neither throws nor mutates `hosts`.

Three existing assertions were repointed at `hosts[0]` through new `readStoredHost`/`readStoredHosts` helpers — their premise, the file's root shape, genuinely changed. The legacy-backfill test now writes a bare v1 record back, so it covers the v1→v2 migration too.

**Watched each one fail** (Tests rule 2), one break at a time, restoring between runs:

| Break | Red test | Failure |
| --- | --- | --- |
| `activateAccount` matches `hosts[0]` first | keeps each account's host separate | `expected true to be false` — B borrowing A's host, the bug in #201 |
| `deactivate()` leaves `activeAccountId` set | leaves no host bound after signing out | `expected true to be false` |
| adoption-binding block removed | adopts a host configured before accounts existed | `expected null to deeply equal { id: 'account-a', … }` |
| `activateAccount` persists via `#persist()` | activates nothing for an account that has no host | `The team server is not configured.` |

I dropped the optional `host-service.test.ts` the plan allowed for. `HostServiceOptions` requires the concrete `AgentService`, `SkillMarketplaceService`, `SidebarLayoutStore`, `MailboxStore` and `BrowserHost` classes — unlike `TeamApiServer`, which takes narrow structural interfaces — so constructing it needs `as unknown as` assertions, and the escape hatch is worse than the coverage.

## Verification

- `bun run test:desktop -- src/main/team-store.test.ts` — 25/25
- `bun run test:desktop -- src/main/team-api-server.test.ts src/main/remote-server-manager.test.ts src/main/team-webrtc-host-gateway.test.ts` — 55/55 (all three consume `TeamStore` and would catch a broken `#state` contract)
- `bun run test:desktop -- src/main/ipc-channel-coverage.test.ts` — 13/13
- `bun run typecheck:node` and `biome check` on the four changed files — clean
- No repo-wide checks; CI owns those.

**Not done:** the manual account switch against the running dev stack. It needs the shared dev instance, which other sessions are using.

No `biome-ignore`, `@ts-expect-error`, `overrides` entry, `any`/`unknown` widening or checker assertion was added.

### DIFF
--- a/scripts/seed-dev-state.test.ts
+++ b/scripts/seed-dev-state.test.ts
@@ -152,7 +152,10 @@ describe("development state seed", () => {
       );
     }
 
-    const team = new TeamStore(join(profilePath, "openbot-team-server-v1.json"));
+    const team = new TeamStore(
+      join(profilePath, "openbot-team-server-v2.json"),
+      join(profilePath, "openbot-team-server-v1.json"),
+    );
     await team.initialize();
     const members = team.listMembers();
     const owner = members.find((member) => member.role === "owner");
--- a/scripts/seed-dev-state.ts
+++ b/scripts/seed-dev-state.ts
@@ -19,7 +19,9 @@ import { resolveDevelopmentAppDataRoot } from "./development-state-paths";
 
 export const DEVELOPMENT_SEED_MANIFEST_FILE = "openbot-dev-seed-v1.json";
 
-const TEAM_FILE = "openbot-team-server-v1.json";
+const TEAM_FILE = "openbot-team-server-v2.json";
+/** Read only, exactly as the app reads it: the file a build without accounts owns. */
+const LEGACY_TEAM_FILE = "openbot-team-server-v1.json";
 const SETUP_FILE = "openbot-setup-v2.json";
 const SEED_VERSION = 1;
 const SEEDED_AT = "2026-08-21T10:00:00.000Z";
@@ -602,7 +604,7 @@ async function seedAgentExchanges(mailbox: MailboxStore): Promise<void> {
 }
 
 async function seedTeam(profilePath: string, botStore: BotStore): Promise<void> {
-  const team = new TeamStore(join(profilePath, TEAM_FILE));
+  const team = new TeamStore(join(profilePath, TEAM_FILE), join(profilePath, LEGACY_TEAM_FILE));
   await team.initialize();
   const owner = {
     id: "openbot-dev-owner",
--- a/scripts/seed-dev-state.ts
+++ b/scripts/seed-dev-state.ts
@@ -610,7 +610,7 @@ async function seedTeam(profilePath: string, botStore: BotStore): Promise<void> 
     name: "Dev Owner",
     avatarUrl: null,
   };
-  await team.configureWithAccount("OpenBot Dev Team", owner);
+  const teamIdentity = await team.configureWithAccount("OpenBot Dev Team", owner);
   const joined = [];
   for (const member of [
     { id: "openbot-dev-alice", email: "alice@example.com", name: "Alice Chen", role: "admin" as const },
@@ -629,7 +629,7 @@ async function seedTeam(profilePath: string, botStore: BotStore): Promise<void> 
   }
   await team.createInvite("member", "new-person@example.com");
   const ownerSession = await team.loginWithAccount(owner);
-  await team.setEnabledOnLaunch(true);
+  await team.setEnabledOnLaunch(teamIdentity.serverId, true);
 
   const [alice, jon, maya] = joined;
   if (!alice || !jon || !maya) throw new Error("The development team members could not be created.");
--- a/src/main/central-auth-manager.test.ts
+++ b/src/main/central-auth-manager.test.ts
@@ -761,6 +761,58 @@ describe("CentralAuthManager", () => {
     expect(Buffer.from(stored, "base64").toString()).not.toContain("machine-token-for-the-first-account");
   });
 
+  it("does not keep one account's host credential beside the next account's session", async () => {
+    const root = await createRoot();
+    const storagePath = join(root, "session.bin");
+    let verifications = 0;
+    const manager = new CentralAuthManager({
+      apiUrl: "https://api.openbot.run",
+      storagePath,
+      encrypt: (value) => Buffer.from(value),
+      decrypt: (value) => value.toString(),
+      fetch: vi.fn(async (input: string | URL | Request) => {
+        const url = new URL(input.toString());
+        if (url.pathname.endsWith("/start")) {
+          return Response.json({ challengeId: "challenge-1", expiresAt: 10_000 });
+        }
+        if (url.pathname.endsWith("/verify")) {
+          verifications += 1;
+          return Response.json({
+            sessionToken: verifications === 1 ? "session-one" : "session-two",
+            user: {
+              id: `user-${verifications}`,
+              email: `person${verifications}@example.com`,
+              name: null,
+              avatarUrl: null,
+            },
+          });
+        }
+        return Response.json({
+          hostId: "8f1c1f2e-1d9a-4a1a-9d1e-2f7c6b5a4d3c",
+          name: "Studio Mac",
+          membershipId: "member-1",
+          authEpoch: 1,
+          machineToken: "machine-token-for-the-first-account-0001",
+        });
+      }),
+    });
+    await manager.requestEmailCode("person1@example.com");
+    await manager.verifyEmailCode("challenge-1", "ABCD-EFGH");
+    await manager.registerRemoteHost({
+      hostId: "8f1c1f2e-1d9a-4a1a-9d1e-2f7c6b5a4d3c",
+      name: "Studio Mac",
+      ownerMembershipId: "member-1",
+    });
+
+    // Signing in as somebody else without signing out first.
+    await manager.requestEmailCode("person2@example.com");
+    await manager.verifyEmailCode("challenge-1", "ABCD-EFGH");
+
+    const stored = Buffer.from(await readFile(storagePath, "utf8"), "base64").toString();
+    expect(stored).toContain("session-two");
+    expect(stored).not.toContain("machine-token-for-the-first-account");
+  });
+
   it("updates the signed-in account name", async () => {
     const root = await createRoot();
     const requests: Request[] = [];
--- a/src/main/central-auth-manager.test.ts
+++ b/src/main/central-auth-manager.test.ts
@@ -695,6 +695,72 @@ describe("CentralAuthManager", () => {
     expect(requests.at(-1)?.method).toBe("DELETE");
   });
 
+  it("does not file a host credential under the account that signed in while it was issued", async () => {
+    const root = await createRoot();
+    const storagePath = join(root, "session.bin");
+    let releaseRegistration: () => void = () => undefined;
+    const registrationHeld = new Promise<void>((resolve) => {
+      releaseRegistration = resolve;
+    });
+    let registrationStarted: () => void = () => undefined;
+    const started = new Promise<void>((resolve) => {
+      registrationStarted = resolve;
+    });
+    let verifications = 0;
+    const manager = new CentralAuthManager({
+      apiUrl: "https://api.openbot.run",
+      storagePath,
+      encrypt: (value) => Buffer.from(value),
+      decrypt: (value) => value.toString(),
+      fetch: vi.fn(async (input: string | URL | Request) => {
+        const url = new URL(input.toString());
+        if (url.pathname.endsWith("/start")) {
+          return Response.json({ challengeId: "challenge-1", expiresAt: 10_000 });
+        }
+        if (url.pathname.endsWith("/verify")) {
+          verifications += 1;
+          return Response.json({
+            sessionToken: verifications === 1 ? "session-one" : "session-two",
+            user: {
+              id: `user-${verifications}`,
+              email: `person${verifications}@example.com`,
+              name: null,
+              avatarUrl: null,
+            },
+          });
+        }
+        registrationStarted();
+        await registrationHeld;
+        return Response.json({
+          hostId: "8f1c1f2e-1d9a-4a1a-9d1e-2f7c6b5a4d3c",
+          name: "Studio Mac",
+          membershipId: "member-1",
+          authEpoch: 1,
+          machineToken: "machine-token-for-the-first-account-0001",
+        });
+      }),
+    });
+    await manager.requestEmailCode("person1@example.com");
+    await manager.verifyEmailCode("challenge-1", "ABCD-EFGH");
+
+    const pending = manager.registerRemoteHost({
+      hostId: "8f1c1f2e-1d9a-4a1a-9d1e-2f7c6b5a4d3c",
+      name: "Studio Mac",
+      ownerMembershipId: "member-1",
+    });
+    // Attach the rejection handler before the switch, so settling it raises no unhandled error.
+    const settled = pending.catch(() => undefined);
+    await started;
+    await manager.requestEmailCode("person2@example.com");
+    await manager.verifyEmailCode("challenge-1", "ABCD-EFGH");
+    releaseRegistration();
+    await settled;
+
+    await expect(pending).rejects.toThrow("signed-in account changed");
+    const stored = await readFile(storagePath, "utf8");
+    expect(Buffer.from(stored, "base64").toString()).not.toContain("machine-token-for-the-first-account");
+  });
+
   it("updates the signed-in account name", async () => {
     const root = await createRoot();
     const requests: Request[] = [];
--- a/src/main/central-auth-manager.ts
+++ b/src/main/central-auth-manager.ts
@@ -139,6 +139,7 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
   #state: CentralAuthState = { status: "loading" };
   #sessionToken: string | null = null;
   readonly #teamHostTokens = new Map<string, string>();
+  #sessionWriteChain: Promise<void> = Promise.resolve();
   #remoteTicketJwks: Promise<z.infer<typeof remoteTicketJwksSchema>> | null = null;
   #initializationPromise: Promise<CentralAuthState> | null = null;
   #emailCodeRequest: EmailCodeRequest | null = null;
@@ -786,7 +787,17 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
     };
   }
 
-  async #writeStoredSession(): Promise<void> {
+  #writeStoredSession(): Promise<void> {
+    // Serialized: two writes racing inside their filesystem awaits would let the earlier
+    // one rename its snapshot over the later one, restoring a session the user has left.
+    this.#sessionWriteChain = this.#sessionWriteChain.then(
+      () => this.#writeStoredSessionNow(),
+      () => this.#writeStoredSessionNow(),
+    );
+    return this.#sessionWriteChain;
+  }
+
+  async #writeStoredSessionNow(): Promise<void> {
     if (!this.#sessionToken) return;
     if (!this.#options.canPersist()) {
       await rm(this.#options.storagePath, { force: true });
@@ -814,7 +825,13 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
   async #clearStoredSession(): Promise<void> {
     this.#sessionToken = null;
     this.#teamHostTokens.clear();
-    await rm(this.#options.storagePath, { force: true });
+    // Through the same chain as the writes, so a write already in flight cannot put the
+    // file back after it is removed.
+    this.#sessionWriteChain = this.#sessionWriteChain.then(
+      () => rm(this.#options.storagePath, { force: true }),
+      () => rm(this.#options.storagePath, { force: true }),
+    );
+    await this.#sessionWriteChain;
   }
 
   #restoreStoredSession(value: string): void {
--- a/src/main/central-auth-manager.ts
+++ b/src/main/central-auth-manager.ts
@@ -241,6 +241,7 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
     ownerMembershipId: string;
     devicePublicKey?: string | null;
   }): Promise<RegisteredRemoteHost> {
+    const sessionToken = this.#sessionToken;
     const storedMachineToken = this.#teamHostTokens.get(input.hostId.toLowerCase());
     const result = await this.#authorizedRequest(
       "/v2/remote/hosts/register",
@@ -255,6 +256,12 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
       },
       decodeRegisteredRemoteHost,
     );
+    if (this.#sessionToken !== sessionToken) {
+      // The credential belongs to the account that asked for it. Writing it now would file
+      // it under whichever session is stored next, so the caller is told the registration
+      // no longer applies instead.
+      throw new Error("The signed-in account changed while this server was being registered.");
+    }
     if (result.machineToken) this.#teamHostTokens.set(input.hostId.toLowerCase(), result.machineToken);
     await this.#writeStoredSession();
     return result;
--- a/src/main/central-auth-manager.ts
+++ b/src/main/central-auth-manager.ts
@@ -140,6 +140,8 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
   #sessionToken: string | null = null;
   readonly #teamHostTokens = new Map<string, string>();
   #sessionWriteChain: Promise<void> = Promise.resolve();
+  /** The account the stored host credentials were issued to, or none while signed out. */
+  #sessionAccountId: string | null = null;
   #remoteTicketJwks: Promise<z.infer<typeof remoteTicketJwksSchema>> | null = null;
   #initializationPromise: Promise<CentralAuthState> | null = null;
   #emailCodeRequest: EmailCodeRequest | null = null;
@@ -638,6 +640,11 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
         },
         decodeSessionResponse,
       );
+      // Signing in as somebody else without signing out first. The host credentials belong
+      // to the account that was issued them, and must not be filed under this session.
+      if (this.#sessionAccountId !== null && this.#sessionAccountId !== session.user.id) {
+        this.#teamHostTokens.clear();
+      }
       this.#sessionToken = session.sessionToken;
       await this.#writeStoredSession();
       return this.#setState({
@@ -824,6 +831,7 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
 
   async #clearStoredSession(): Promise<void> {
     this.#sessionToken = null;
+    this.#sessionAccountId = null;
     this.#teamHostTokens.clear();
     // Through the same chain as the writes, so a write already in flight cannot put the
     // file back after it is removed.
@@ -856,6 +864,9 @@ export class CentralAuthManager extends EventEmitter<CentralAuthEvents> {
   }
 
   #setState(state: CentralAuthState): CentralAuthState {
+    // Held apart from the state, which passes through `code_sent` on the way to another
+    // account: this is whose credentials the store is holding, until they are cleared.
+    if (state.status === "signed_in") this.#sessionAccountId = state.user.id;
     this.#state = state;
     const copy = this.getState();
     this.emit("changed", copy);
--- a/src/main/host-service.ts
+++ b/src/main/host-service.ts
@@ -356,10 +356,14 @@ export class HostService extends EventEmitter<HostEvents> {
         message: "Registered this OpenBot for WebRTC access.",
       });
     } catch (error) {
-      this.#setStatus({
-        phase: "error",
-        message: error instanceof Error ? error.message : "Could not reserve the public address.",
-      });
+      // A failure that arrives after another account signed in belongs to the host that is
+      // gone, so it must not overwrite the status the new account is looking at.
+      if (this.#isActiveHost(identity.serverId)) {
+        this.#setStatus({
+          phase: "error",
+          message: error instanceof Error ? error.message : "Could not reserve the public address.",
+        });
+      }
     }
     return this.getStatus();
   }
@@ -392,6 +396,12 @@ export class HostService extends EventEmitter<HostEvents> {
    * for a host the account no longer has would register or upload on the wrong account's
    * behalf. The store's own guards stop the local half; this stops the network half.
    */
+  #assertStillActiveHost(serverId: string): void {
+    if (!this.#isActiveHost(serverId)) {
+      throw new Error("The signed-in account changed while this server was being updated.");
+    }
+  }
+
   #isActiveHost(serverId: string): boolean {
     return this.#options.store.getIdentity()?.serverId === serverId;
   }
@@ -665,6 +675,10 @@ export class HostService extends EventEmitter<HostEvents> {
         (member) => member.membershipId === input.memberId,
       );
       if (!current || current.role === "owner") throw new Error("The remote member does not exist.");
+      // The directory read is a round trip, and the account can change during it. Mutating
+      // the previous account's host with the new account's authorization is what this guard
+      // stops; the same check runs again before the result is written back.
+      this.#assertStillActiveHost(hostId);
       if (input.disabled) await this.#options.removeRemoteMember(hostId, input.memberId);
       else
         await this.#options.updateRemoteMember(
@@ -673,6 +687,7 @@ export class HostService extends EventEmitter<HostEvents> {
           input.role ?? current.role,
           input.disabled === false,
         );
+      this.#assertStillActiveHost(hostId);
       const members = await this.#options.listRemoteMembers(hostId);
       const updated = members.find((member) => member.membershipId === input.memberId);
       if (!updated) throw new Error("The remote member does not exist.");
--- a/src/main/host-service.ts
+++ b/src/main/host-service.ts
@@ -281,20 +281,14 @@ export class HostService extends EventEmitter<HostEvents> {
     const previousServerId = this.#status.serverId;
     if (this.#boundAccountId !== undefined) {
       this.#runtimeGeneration += 1;
-      // Isolation cannot depend on the teardown succeeding: a WebRTC disconnect rejects
-      // on a command error or a timeout, and leaving the previous account's host bound is
-      // by far the worse failure. Report it and rebind anyway.
-      try {
-        // The same three steps as `stop`, and for the same reason: a start still in flight
-        // belongs to the previous account. The bumped generation makes it abort at its
-        // next checkpoint, and draining it here is what stops `start()` from handing the
-        // new account that superseded operation instead of starting its own host.
-        await this.#stopRuntime();
-        await this.#startOperation;
-        await this.#stopRuntime();
-      } catch (error) {
-        console.error("Unable to stop the host runtime while switching accounts:", error);
-      }
+      // The same three steps as `stop`, and for the same reason: a start still in flight
+      // belongs to the previous account. The bumped generation makes it abort at its next
+      // checkpoint, and draining it here is what stops `start()` from handing the new
+      // account that superseded operation instead of starting its own host. Each step is
+      // attempted on its own, so a gateway that will not come down cannot skip the drain.
+      await this.#attemptTeardown(() => this.#stopRuntime());
+      await this.#attemptTeardown(() => this.#startOperation);
+      await this.#attemptTeardown(() => this.#stopRuntime());
     }
     let activated = false;
     try {
@@ -876,6 +870,19 @@ export class HostService extends EventEmitter<HostEvents> {
 
   async shutdown(): Promise<void> {
     await this.stop(false);
+  }
+
+  /**
+   * Isolation cannot depend on a teardown step succeeding: a WebRTC disconnect rejects on a
+   * command error or a timeout, and leaving the previous account's host bound is by far the
+   * worse failure. Reported, and the steps after it still run.
+   */
+  async #attemptTeardown(step: () => PromiseLike<unknown> | null): Promise<void> {
+    try {
+      await step();
+    } catch (error) {
+      console.error("Unable to stop the host runtime while switching accounts:", error);
+    }
   }
 
   async #stopRuntime(): Promise<void> {
--- a/src/main/host-service.ts
+++ b/src/main/host-service.ts
@@ -380,6 +380,9 @@ export class HostService extends EventEmitter<HostEvents> {
   async updateIdentity(input: UpdateHostIdentityInput): Promise<HostStatus> {
     this.#options.store.assertOwnerAccount(this.#options.getSignedInUser());
     const identity = await this.#options.store.updateIdentity(input);
+    // Before anything is published: the store checked the account before it resolved, and a
+    // switch landing in this gap would show the previous account's name and logo.
+    if (!this.#isActiveHost(identity.serverId)) return this.getStatus();
     this.#setStatus({
       serverName: identity.serverName,
       logoUrl: identity.logoVersion ? serverLogoUrl(identity.logoVersion) : null,
@@ -387,7 +390,6 @@ export class HostService extends EventEmitter<HostEvents> {
     });
     this.#api.refreshIdentity();
     const ownerMembershipId = this.#requiredOwnerMemberId();
-    if (!this.#isActiveHost(identity.serverId)) return this.getStatus();
     await this.#options.registerRemoteHost?.({
       hostId: identity.serverId,
       name: identity.serverName,
@@ -704,6 +706,9 @@ export class HostService extends EventEmitter<HostEvents> {
       const updated = members.find((member) => member.membershipId === input.memberId);
       if (!updated) throw new Error("The remote member does not exist.");
       await this.#options.store.syncRemoteDirectory(hostId, members);
+      // Recording the directory is a write too, so the switch can land inside it and the
+      // member below would be the previous account's.
+      this.#assertStillActiveHost(hostId);
       return {
         id: updated.membershipId,
         username: updated.email,
--- a/src/main/host-service.ts
+++ b/src/main/host-service.ts
@@ -254,7 +254,14 @@ export class HostService extends EventEmitter<HostEvents> {
     const previousServerId = this.#status.serverId;
     if (this.#boundAccountId !== undefined) {
       this.#runtimeGeneration += 1;
-      await this.#stopRuntime();
+      // Isolation cannot depend on the teardown succeeding: a WebRTC disconnect rejects
+      // on a command error or a timeout, and leaving the previous account's host bound is
+      // by far the worse failure. Report it and rebind anyway.
+      try {
+        await this.#stopRuntime();
+      } catch (error) {
+        console.error("Unable to stop the host runtime while switching accounts:", error);
+      }
     }
     if (user) await this.#options.store.activateAccount(user);
     else await this.#options.store.deactivate();
--- a/src/main/host-service.ts
+++ b/src/main/host-service.ts
@@ -46,7 +46,7 @@ import { appendRemoteDiagnosticLog } from "./remote-diagnostics";
 import { RemoteScreenGateway } from "./remote-screen-gateway";
 import type { SkillMarketplaceService } from "./skill-marketplace-service";
 import { TeamApiServer } from "./team-api-server";
-import type { AuthenticatedMember, RemoteDirectoryMember, TeamStore } from "./team-store";
+import type { AuthenticatedMember, RemoteDirectoryMember, TeamIdentity, TeamStore } from "./team-store";
 import type { TeamWebRtcBridge } from "./team-webrtc-bridge";
 import { TeamWebRtcHostGateway } from "./team-webrtc-host-gateway";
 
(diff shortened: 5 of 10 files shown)


NorbertBodziony:
Fixed the P1 in cbff6b21 — the finding was correct. Startup only handled `signed_in`, so a sign-out that settled before `hostService` was constructed left `forwardCentralAuth` nothing to deactivate, and the persisted `activeAccountId` kept the previous account's host configured (and unconfigurable) with nobody signed in. Startup now also calls `applySignedInAccount(null)` on `signed_out`; `loading` and `error` keep their host and let the event listener settle it. The persisted consequence is already covered at the store boundary by "leaves no host bound after signing out, and restores it on the next sign-in", which restarts a `TeamStore` from the same path after `deactivate()` and asserts nothing activates.

/norbiai review

NorbertBodziony:
Fixed the P1 in 86b938eb — the finding was correct, and it is a race my own change introduced: while the active host was fixed for the process lifetime a captured `hostId` could not go stale, but a per-account host can be swapped mid-flight.

Closed it at the store boundary rather than in each caller. `syncRemoteDirectory(serverId, remoteMembers)` now takes the host it was loaded for as a **required** argument, so `tsc` forces all four `HostService` call sites (`start`, `listMembers`, `updateMember`, `removeMember`) to name it, and the store rejects a directory whose `serverId` is no longer the active host with "This server is no longer the active one for the signed-in account."

New test: `refuses a remote directory loaded for a host that is no longer active`. Watched it fail both ways with the guard removed — first `promise resolved "undefined" instead of rejecting`, then, with the rejection assertion relaxed so execution continues, `expected 'account-a-owner' to be '69484685-…'` — i.e. account A's owner membership really is written over B's without the guard, exactly the lockout described. 26/26 in `team-store.test.ts`, 55/55 across `team-api-server` + `remote-server-manager` + `team-webrtc-host-gateway`, `typecheck:node` and `biome check` clean.

/norbiai review

NorbertBodziony:
All three fixed in cd7ddc92. The review was against `cbff6b21`, but every finding still applied to `86b938eb` — that commit only touched `syncRemoteDirectory`.

**1. Runtime cleanup failure prevented isolation** — correct. `await this.#stopRuntime()` sat between the account change and the rebind, and a WebRTC disconnect rejects on a command error or a timeout, so a failed teardown threw out of `applySignedInAccount` before the store was rebound. The teardown is now best-effort and logged, and the rebind always runs. The sign-out path in `forwardCentralAuth` had the same shape around `stop(false)` and is wrapped the same way — stopping is best-effort, unbinding is not.

**2. Remote-host sync failure skipped the local switch** — correct. `applySignedInAccount(state.user)` now runs *before* `syncRemoteHosts()`, and the remote sync carries its own `catch`, so a control-plane failure can no longer leave B signed in with A's host selected and possibly online. `shouldAutoStartHost` still runs last, keeping its original position relative to the sync.

**3. Concurrent configuration could persist duplicate hosts** — correct. `#writeLogo()` yields after the guard, so two logo-bearing requests could both append. The guard is now `#assertNoHostFor(accountId, email)`, rechecked after the awaited preparation, and it removes the logo it just wrote when rejecting. `configure()` rechecks `#state` after `hashPassword` for the same reason.

New test: `stores one host when an account configures twice at once` — two concurrent `configureWithAccount` calls with a logo, then a restart to confirm the surviving host is the one the identity reported. Watched it fail twice with the recheck removed: `expected [ 'fulfilled', 'fulfilled' ] to deeply equal [ 'fulfilled', 'rejected' ]`, and with that assertion dropped so execution continued, `expected [ … ] to have a length of 1 but got 2` — both hosts really are persisted without it.

Findings 1 and 2 are `HostService`/`index.ts` wiring and are not covered by a test: `HostServiceOptions` requires the concrete `AgentService`, `SkillMarketplaceService`, `SidebarLayoutStore`, `MailboxStore` and `BrowserHost`, so constructing it needs `as unknown as` assertions, and `forwardCentralAuth` needs the Electron app. Flagging that rather than adding an escape hatch.

27/27 in `team-store.test.ts`, 68/68 across `team-api-server` + `remote-server-manager` + `team-webrtc-host-gateway` + `ipc-channel-coverage`, `typecheck:node` and `biome check` clean.

/norbiai review

NorbertBodziony:
The review of cd7ddc92 timed out after 15 minutes rather than finding anything — re-running it once.

/norbiai review

NorbertBodziony:
Thanks — the blocking one was right on both halves. Fixed in bc014c5e. (The bot reviewer timed out and then exited 1 on cd7ddc92, so this is the review of record.)

**#1 [blocking] — fixed, both halves.**

*In-flight start survives the switch.* Confirmed: `applySignedInAccount` bumped `#runtimeGeneration` and stopped the runtime but never touched `#startOperation`, so `start()` for B returned A's superseded operation and B stayed idle despite `enabledOnLaunch`. It now drains it with the same three steps `stop` already uses — `#stopRuntime()`, `await #startOperation`, `#stopRuntime()` — inside the existing try/catch, so a rejecting start still cannot block the rebind. I deliberately did *not* just null `#startOperation` out: the old operation aborts at its next checkpoint by calling `#stopRuntime()`, which would then tear down B's freshly started runtime. Draining is what makes the abort land before B starts.

*Preference written onto the wrong host.* Also confirmed — `setEnabledOnLaunch(true)` ran before the generation check. Rather than only reordering, `setEnabledOnLaunch(serverId, enabled)` now names its host the same way `syncRemoteDirectory` does and the store rejects a mismatch, so the invariant is enforced where the write happens instead of at one call site. `start` passes the identity it registered; `stop` skips the write when the host changed under it, since its own awaits span the same window. The generation check moved ahead of the write too.

**#4 — answered with tests, not a manual switch.** You were right that the headline behaviour was unproven. `HostServiceOptions` now types the five collaborators it only forwards (`agents`, `skills`, `sidebarLayout`, `mailbox`, `browser`) by the narrow shapes `TeamApiServer` already declares — `ConstructorParameters<typeof TeamApiServer>[0]["agents"]` and friends. No behaviour change and `index.ts` still passes the real services, but `HostService` is now constructible from the doubles `team-api-server.test.ts` already has, with no `as unknown as` anywhere. Two cases added there under `describe("HostService account binding")`: a switch stops reporting the previous account's server (`configured`, `phase`, `serverId`, `serverName`, `enabledOnLaunch` all reset), and sign-out then sign-in reports the account's own server again. Both go red with the status rebuild reverted to a patch — `expected true to be false` on `configured`, which is precisely the #201 symptom.

Still not covered, and I would rather say so than dress it up: the start-dedupe half of #1. Reaching `registerRemoteHost` needs a non-null `#webrtcGateway`, which needs `teamWebRtcBridge` — still typed as the concrete class, and narrowing that too is a bigger seam change than this PR should carry. Verified by inspection against `stop`'s existing pattern. `shouldAutoStartHost` living in `forwardCentralAuth` is likewise untested.

**#3 — fixed.** `#assertNoHostFor` now scans every stored host for an owner-less one, not just `#state`. New test `refuses a second host beside an owner-less one that is not active`; red with the check reverted to the active host: `promise resolved "{ …(6) }" instead of rejecting`.

**#2 — fixed on the read path.** `listMembers` returns the now-active host's own members when the identity changed while the directory loaded, so a mid-switch read no longer surfaces the store's guard string. `updateMember`/`removeMember` still reject: they are writes the user initiated against a host that is gone, and failing them is correct — I left the message as is rather than adding a mapping layer, happy to take a follow-up if you want it generic.

**Nits.** #6 documented — `configure()` deliberately leaves `activeAccountId` alone because no account owns that host yet, and #3's all-hosts scan is now what stops a second owner-less one appearing beside it. #5 and #7 left alone on purpose: the extra writes are idempotent and the branch to skip them is more state to keep correct than the churn is worth. Say the word and they are a follow-up.

Also fixed a caller I would have missed: `scripts/seed-dev-state.ts` passes the new `serverId` to `setEnabledOnLaunch`.

99 tests green across `team-store` (29), `team-api-server` (22), `remote-server-manager`, `team-webrtc-host-gateway` and `ipc-channel-coverage`. `typecheck:node` and `biome check` clean. Every new assertion watched failing for its own reason. Manual account switch against the dev stack still not done.

NorbertBodziony:
Both P1s from the review of `bc014c5e` are fixed in `ced5958a`.

**Stale configuration reactivating the signed-out account** (`team-store.ts`). `configureWithAccount` now captures `activeAccountId` before the logo write and rejects with *"The signed-in account changed while this server was being created."* if it moved, rolling back the logo it wrote. The comparison is reliable because both `activateAccount` and `deactivate` set `#file.activeAccountId` synchronously before their first await, so a switch that started at any point during the configure is already visible by the time the recheck runs. There is a comment at the capture saying so.

New test, `refuses a configuration that finishes after another account has signed in`: A starts a configure with a logo, B activates during the await, the configure must reject and leave zero hosts on disk. Watched it fail with the recheck deleted — `expected true to be false` on `store.configured`, i.e. exactly the reported symptom: A's stale operation rebinds the store under B's session.

**Remote-session disconnect failure skipping the local rebind** (`index.ts`). `disconnectRemoteSessions()` is now best effort in its own try/catch, and `forwardCentralAuth` always continues to `applySignedInAccount` regardless of how it went.

`bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/remote-server-manager.test.ts src/main/team-webrtc-host-gateway.test.ts src/main/ipc-channel-coverage.test.ts` — 100 passing. Biome and `typecheck:node` clean.

Still outstanding and unchanged: the manual account switch against the dev stack (plan step 4) has not been run.

NorbertBodziony:
Both findings from the review of `ced5958a` are fixed in `541a93fc`.

**[P1] Stale identity update applying account A's logo to B** — correct, and the mechanism is worse than a wrong return value: `#persist` only requires *some* active host, so A's name and image were written onto whichever host was active. `updateIdentity` now rejects if the active host moved during `#writeLogo`, removes the logo it just wrote, and returns `identityOf(state)` — the host it actually changed — rather than rereading `#state`. `getIdentity` goes through the same `identityOf` helper so the two cannot drift.

New test, `refuses an identity update that lands after another account has signed in`. Watched it fail twice with the guard deleted, both halves of the consequence:
- `promise resolved "{ …(6) }" instead of rejecting` — the stale update succeeds;
- with that assertion dropped so the run reaches the disk check, `expected [ 'Renamed', 'Studio Air' ] to deeply equal [ 'Studio Mac', 'Studio Air' ]` — the signed-out account's host is renamed on disk, and gains the logo.

**[P2] Stale stop disabling the newly active account's host** — correct, and it was my own earlier fix aimed the wrong way: reading the identity *after* the awaits is precisely how B's id gets there. `stop` now captures the server id before `#runtimeGeneration += 1` and returns early, touching neither the preference nor the status, unless that same host is still active.

**No test for the stop race, and I'm not pretending otherwise.** I wrote one — configure B, set its preference, switch to A, `stop()` unawaited, switch back to B — and it was green with the guard deleted. With nothing listening, `stop`'s teardown chain completes in microtasks while `activateAccount` is still writing its file, so the harness cannot interleave the two; making it interleave needs a real bound runtime. Per test rule 2 I deleted it rather than keep a test that proves nothing. The captured id alone also means the store's own `setEnabledOnLaunch` guard would now catch this case, so the early return is the difference between a clean no-op and a thrown error.

`bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/remote-server-manager.test.ts src/main/team-webrtc-host-gateway.test.ts src/main/ipc-channel-coverage.test.ts` — 101 passing. Biome and `typecheck:node` clean.

Manual dev-stack account switch (plan step 4) still not run.

NorbertBodziony:
All four findings from the review of `541a93fc` are fixed in `c3e8e616`. All four were real; nothing withdrawn.

**[P1] Automatic activation overwriting an unrecognized team file** — the serious one, and it fails the "nothing backs this file up" rule outright. The store now records that the file existed but could not be read, `#persistFile` refuses in that state, and signing in no longer writes when there is no host to record — which is what keeps that refusal from failing every sign-in rather than only the configure it should. Configuring a server fails with a message telling the user to move the file aside.

Watched the new test fail twice with the guard removed: `promise resolved "{ …(6) }" instead of rejecting` for the configure, and then — with that assertion dropped so the run reaches disk — the whole thing, verbatim:
```
- {"version":3,"hosts":[{"serverId":"from-the-future"}]}
+ {"version":2,"activeAccountId":"account-a","hosts":[{"version":1,"serverId":"d974968e…
```
This also makes the PR body's downgrade note wrong in the other direction, so I've corrected it: an older build does **not** leave a v2 file alone — it clobbers it on the next auth event, exactly as described above. Pre-existing behaviour rather than something this PR introduces, but the body claimed otherwise and now says so plainly.

**[P1] Configuration finishing during a switch targeting the new account** — the guard covered the logo write but not `#persistFile`. It now compares the host it created after the write and returns `identityOf(created)`. The created host stays on disk: it is A's own and signing back in as A must return it. New test with B already holding a host, so the pre-fix path returns B's identity rather than the null it would return if B had none; red with the guard removed as `promise resolved "{ …(6) }" instead of rejecting`.

**[P1] WebRTC teardown failure skipping API teardown** — `#api.stop()` is now in a `finally`. **No test**: `#api` is constructed inside `HostService` and the gateway comes through the bridge option, so forcing the gateway to throw needs a seam that does not exist yet. Flagging it rather than claiming coverage.

**[P2] Failed account persistence leaving memory bound to the wrong account** — `activateAccount` and `deactivate` now restore `#state`, `activeAccountId` and the adopted owner's fields on a failed write. New test using the existing rename-the-directory trick; red with the rollback removed as `expected null to deeply equal { …(6) }`, i.e. the store reporting no host while A is still signed in.

One test-helper change worth naming: `readStoredHosts` now treats a missing file as no stored hosts, because "nothing worth recording" writes no file at all.

`bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/remote-server-manager.test.ts src/main/team-webrtc-host-gateway.test.ts src/main/ipc-channel-coverage.test.ts` — 104 passing. Biome and `typecheck:node` clean.

Manual dev-stack account switch (plan step 4) still not run.

NorbertBodziony:
Both findings from the review of `c3e8e616` are fixed in `32ac1003` (force-pushed once to fix a malformed commit message; no content difference).

**[P1] Identity update escaping into the next account session** — correct, and I traced the path you named: with A's identity returned, `host-service.ts:349` evaluates `#requiredOwnerMemberId()` after the switch, so `registerRemoteHost` goes out with A's `hostId` and B's `ownerMembershipId`. The store recheck now spans the write, which makes that path unreachable — there is no yield between `store.updateIdentity` resolving and the register call, so the arguments cannot go stale after it. I did not also capture the owner membership up front: with the guard in place it would be dead code, and I would rather not add a second mechanism nothing can exercise. Say the word if you want it anyway as defence in depth.

The rename itself stays on disk. It is what that account asked for, the write already succeeded, and signing back in must show it.

New test, red with the guard removed as `promise resolved "{ …(6) }" instead of rejecting`.

**[P2] Failed account persistence leaving memory bound to the wrong account** — you were right that the first fix only covered its own write. `syncAccount`'s profile refresh and `#prune`'s expiry sweep are now split into `#applyAccount` and `#dropExpired`, mutation without a write, so an activation mutates in memory and commits with **one** `#recordActiveAccount()`. The rollback restores the whole file (`structuredClone` up front, `#state` re-derived by index) rather than the three fields I happened to think of — which also covers the adopted owner and anything added later. Public `syncAccount` and `#prune` keep their own writes for their own callers.

The existing rollback test still names the consequence and is still red without it: `expected null to deeply equal { …(6) }`, the store reporting no host while A is signed in.

`bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/remote-server-manager.test.ts src/main/team-webrtc-host-gateway.test.ts src/main/ipc-channel-coverage.test.ts` — 105 passing. Biome and `typecheck:node` clean.

Still uncovered and stated each round: the `#stopRuntime` `finally` (no seam to make the gateway throw), the `stop` race (the harness cannot interleave it), and the manual dev-stack account switch.

NorbertBodziony:
All four findings from the review of `32ac1003` are fixed in `786b92cd`. All four were real; I also fixed the same window in `configure`, which the review only flagged in `updateIdentity`.

**[P1, REMAINS] Identity update escaping into the next session** — you were right that my store-side guard ends where the store returns. `HostService` then awaits `registerRemoteHost` and `updateRemoteHostLogo`, and a switch in either window sends A's server to the directory under B's authentication. Both steps now run only while their own host is still active (`#isActiveHost`), the owner membership is read before the first await, and `configure` gets the same treatment — its trailing `#setStatus` was also free to overwrite B's status.

I did not implement the "persist credentials only if the captured auth session is current" half: `registerRemoteHost` is supplied by `index.ts` from the central auth manager, and threading a captured session through it is a change to that seam rather than to this one. Worth its own PR if you want it — say so and I will open one.

New test, red with the guard removed as `expected [ Array(1) ] to deeply equal []`. It needs a barrier on the registration call, because the store guard already covers everything before it.

**[P1, NEW] Pending invite reads disclosing the previous account's data** — correct, and the more serious of the two new ones: `listInvites` had no active-host check where `listMembers` did. Added. Red without it, showing exactly the disclosure:
```
+     "email": "invited-by-a@example.com",
```

**[P1, NEW] Account B announced while A remains accessible** — correct. `HostService.unbindChangedAccount(user)` is a synchronous in-memory unbind that `forwardCentralAuth` calls before it sends `authEvent`; the file is still written by `applySignedInAccount`. It deliberately does nothing when no account has been applied yet, so startup keeps the status the constructor published for a host that may already be online. Red as `expected true to be false` on `configured` when made a no-op.

Worth noting one consequence: after the sync unbind, `stop()` sees an unconfigured status and returns early, so the sign-out path's teardown now rests entirely on `applySignedInAccount(null)` — which tears the runtime down on `#boundAccountId`, not on the status, so it still happens.

**[P2, REMAINS] Failed persistence** — I take the point: restoring A put the store back in step with itself but not with central authentication, which had already moved on. Changed to fail closed, as you suggest — the file keeps every host, and nothing is bound until an auth event or a restart records the switch. `applySignedInAccount` publishes the resulting status in a `finally`, so a throw cannot leave the rail naming a host the process no longer answers for.

That inverts the premise of the test I added last round, so I rewrote it rather than loosened it: it now asserts nothing is bound, the file still holds the host, and re-activating the account returns it. Red without the change as `expected null to deeply equal { …(6) }` before; now red the other way if the fail-closed line goes back to restoring.

`bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/remote-server-manager.test.ts src/main/team-webrtc-host-gateway.test.ts src/main/ipc-channel-coverage.test.ts` — 108 passing. Biome and `typecheck:node` clean.

Still uncovered, stated every round: the `#stopRuntime` `finally`, the `stop` race, and the manual dev-stack account switch.

NorbertBodziony:
`820f8565` closes all four.

**[P1] Identity update can escape into the next account session** — done, and I was wrong to defer it: the fix does not need the `index.ts` seam at all. `registerRemoteHost` now captures `#sessionToken` before the request and refuses to store the machine token or rewrite the session if it changed, matching what `updateAvatar` and `updateName` already do a few hundred lines below. It throws rather than returning quietly, so the caller cannot act on a registration that belongs to a session that is gone.

**[P1] Account B is announced while account A remains accessible** — each `forwardCentralAuth` call takes a generation; a queued callback that is no longer the latest returns immediately. The newest transition does the disconnect, the activation and the autostart for the state the renderer was actually told about, so dropping the intermediates loses no work.

**[P1] Remote member mutation crosses an account switch** — `updateMember` asserts the captured host is still active after the directory read (before `removeRemoteMember`/`updateRemoteMember`) and again before the write-back.

**[P2] Failed registration overwrites the next account's status** — `configure`'s catch publishes the error only while that host is still active.

Watched each new test fail with its own guard removed: A's machine token written into `session-two`; the member mutation reaching A's host id; B's status flipped to `phase: "error"`.

Still uncovered: the `forwardCentralAuth` generation guard (no seam to drive `index.ts` from a test), the `#stopRuntime` `finally`, and the `stop()` race, as reported before.

NorbertBodziony:
`3e8591b2` takes all four.

**[P1] Identity update can escape into the next account session** — correct, the token comparison did not cover the write itself. `#writeStoredSession` now runs through a single chain, and `#clearStoredSession`'s removal joins the same chain, so a write already in flight cannot rename a session the user has left back into place.

**[P1] Account B is announced while account A remains accessible** — the generation is rechecked after `disconnectRemoteSessions()`, so a callback that was awaiting when a newer account was announced returns instead of activating.

**[P1] Invite completion crosses into the next account** — `createInvite` asserts the captured host is still active after `createRemoteInvite`, before the link is built, the mail is sent or the invitation is returned. The invitation itself stays on the host that asked for it and appears in that account's invite list; revoking it would mean spending the new account's authorization on the old account's host.

**[P2] Successful configuration can restore stale account status** — `configure` rechecks immediately after `configureWithAccount` resolves, before any status or presence publish.

One new test, watched fail with its guard removed: the invitation mail went to the guest naming A's server after B had signed in.

Not covered by a test: the session write chain and the `configure` publish gap have no seam to interleave from — same as the `index.ts` generation guard, the `#stopRuntime` `finally` and the `stop()` race, all reported earlier.

NorbertBodziony:
`9c4800f7` takes all four.

**[P1] Mutable email can transfer an already-owned host** — the most valuable finding of the round, and it was two bugs facing opposite ways. Account ID is now the identity: activation matches an owner by address only when that owner has no account ID (the pre-accounts and v1-migration case), `#applyAccount` finds the owner member by account ID before address, and `assertOwnerAccount` prefers the ID. So a recycled address no longer inherits a host, and an account that changes its own address keeps the one it owns.

**[P2] Failed activation permanently suppresses retry** — a failure now leaves `#boundAccountId` unknown rather than recorded, so the next report of the same account runs activation again instead of short-circuiting into an unconfigured store.

**[P1] Invite completion crosses into the next account** — rechecked after the email await, and `revokeRemoteInvite` runs only while that host is still active, so a stale completion neither returns the link to another account's renderer nor spends the new account's authorization on the old account's host.

**[P1] Account B is announced while account A remains accessible** — the generation is rechecked after `applySignedInAccount` too; a stale callback unbinds and returns, leaving no host bound until the newest account's own callback binds it.

Three new tests, each watched fail with its guard removed: the stranger holding the recycled address received A's server; the renamed account was refused ownership of its own host; the account whose activation failed stayed unconfigured when it was reported again.

Uncovered, and stated plainly: the post-email invite recheck (that boundary already has a test in the same method), the `index.ts` generation guard, the session write chain, the `configure` publish gap, the `#stopRuntime` `finally` and the `stop()` race. The last five have no seam to interleave from without adding one to production code for the test's sake.

NorbertBodziony:
`dedfc7bf` takes both.

**[P2] Recycled email prevents the new account from configuring a host** — a fair catch on my own last commit: I applied the account-ID rule where a host is bound and not where one is created, so B was refused both. `#hostFor` now follows the same rule as `activateAccount`, and the recycled-address test configures a host as the stranger to hold it there. Watched it fail with the rule removed — "The team server is already configured."

**[P1] Member lookup can return the previous account's directory** — rechecked after `syncRemoteDirectory`, which is a write, not only before it. A switch landing during that write now answers with the active host's own members.

Not covered by a test: the recheck after `syncRemoteDirectory`, which needs a seam to interleave a switch into a store write — the same reason as the session write chain, the `configure` publish gap, the `index.ts` generation guard, the `#stopRuntime` `finally` and the `stop()` race.

NorbertBodziony:
`42037661` takes both. They are mirrors of guards already in this file, so they are two lines and no new test.

**[P1] Member update can return the previous account's directory data** — `updateMember` now rechecks after `syncRemoteDirectory`, the same as `listMembers` does.

**[P2] Identity update publishes stale host details before validating** — the active-host check moved to immediately after `store.updateIdentity` resolves, before the status, the identity refresh and `#requiredOwnerMemberId`.

No new test for either: both are the stale-completion window the existing team-store and host-service race tests already name, and neither has a seam to interleave a switch into a store write from. Adding one would mean putting a hook in production code for the test's sake.

For what it is worth on the shape of this PR: it started as "one host per account" and has taken eight review rounds, now at +1350/-111. The remaining class — a recheck after every await that touches a store write — is real but converging, and I would rather land this and take anything further as its own PR than keep widening the diff. Happy either way; it is @norbertbodziony's call.

NorbertBodziony:
`b7331af9` takes the one that was left, and the analysis was right: while the *first* host is being written nothing is bound, so neither the store's `activeAccountId` nor `unbindChangedAccount` can see an account announced during that write — both are looking at a host that does not exist yet.

I took a different marker than suggested. Rather than record an announced account inside `HostService`, `configure` captures the account it was asked for and compares it with `getSignedInUser()` after the write: `CentralAuthManager` sets that state in the same call that announces the account to the renderer, so it is the earliest thing the main process can compare against, and there is no second copy to keep in step with it. If it moved, the configuration is refused; the host stays on disk for the account that asked for it, as with every other create in this PR.

Test: the account is announced while the host and its logo are being written, with the switch still queued — the state `forwardCentralAuth` actually leaves. Watched it fail with the check removed: A's server was published to B's renderer and registered under B's authentication.

NorbertBodziony:
`5d3c4568`. Correct — refusing the call was not enough on its own: `configureWithAccount` binds the host as it creates it, so the members and identity behind it kept answering the account that had just been announced, the owner's address and name included, while the status said unconfigured.

The host is now unbound and the status republished before the refusal is thrown. It stays on disk for the account that asked for it, like every other host this change creates.

The existing test reads the members after the refusal; watched it fail with the unbind removed — account A's owner came back.

NorbertBodziony:
**[NEW][P1] Migrating the only team file makes downgrade capable of destroying every host** — fixed in `0bd5959c`.

The read path and the write path are now separate files. `TeamStore` takes `(path, legacyPath)`; `src/main/index.ts` passes the new `openbot-team-server-v2.json` as the file it writes and the shipped `openbot-team-server-v1.json` as a read-only fallback. Initialization prefers the v2 file, and only if it is absent does it read the legacy record and copy it into the v2 envelope — the legacy file is never written, renamed or deleted. A downgrade therefore finds the single-host file exactly as the last account-less build left it, with its members, invites, keys and logo reference, rather than an unparseable file it would overwrite on the next configure. `#logoRoot` stays anchored to the legacy path, so no logo asset moves.

Test: `adopts the host of a build without accounts without writing over its file` in `src/main/team-store.test.ts` — loads a v1 record from the legacy path, resolves its logo, renames the server, then asserts the legacy file is byte-for-byte what it was and the rename landed in the v2 file. Watched it fail: pointing the write path back at the legacy file turns the last assertion red with the v2 envelope having replaced the v1 record, which is the consequence itself.

This supersedes the downgrade note in the PR body — a downgrade now keeps its host rather than showing an unconfigured one.

Checks: `bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts` (68 passed), `biome check` on the touched files, `bun run typecheck:node`.

Still uncovered by tests, unchanged from the previous round: the `index.ts` generation guard, the session write chain in `CentralAuthManager`, the `configure` publish gap, the recheck after `syncRemoteDirectory`, `#stopRuntime`'s `finally`, and the `stop()` race — none has a seam to interleave an account switch into a store write without adding a hook to production code for the test's sake.

NorbertBodziony:
All three fixed in `d518cd35`.

**[P1] Shared logo storage lets either version delete the other version's logo** — each file now owns its asset directory: `#logoRoot` is derived from the v2 path, `#legacyLogoRoot` from the legacy one. Importing a legacy record copies its logo into the v2 root under the same version name (`#adoptLegacyLogo`, copy and never move), so `updateIdentity`'s `#removeLogo` can only ever touch this build's copy. Neither format can now delete an asset the other's record points at.

**[P2] Changes made after downgrading are permanently shadowed by the stale v2 copy** — the envelope records `legacyImport: { fingerprint, serverId }`, a SHA-256 of the legacy file as it was read. Every start re-reads that file; a different fingerprint means an older build has written it since, so its record replaces the host with the same `serverId` and becomes the active one. The copy it supersedes is moved to `replacedHosts` rather than dropped — both are the user's work and neither file is a backup of the other. An unreadable v2 file skips the reconcile entirely, so the "something else owns this path" guard still holds.

**[P2] Development seeding writes the v2 envelope under the legacy filename** — `scripts/seed-dev-state.ts` now constructs `TeamStore` with the same path pair the app uses (`openbot-team-server-v2.json` plus the legacy file read-only), and `scripts/seed-dev-state.test.ts` opens the same pair.

Tests, both in `src/main/team-store.test.ts`:
- the existing legacy-adoption test now replaces the logo after the import and asserts the older build's asset file is still readable;
- new `takes in what a build without accounts wrote after the upgrade, keeping the copy it supersedes` — renames the host in v2, then rewrites the legacy file as an older build would, and asserts the restarted store shows the older build's name with the superseded copy under `replacedHosts`.

Watched both fail together: pointing `#logoRoot` back at the legacy directory turns the first red with `ENOENT` on the older build's logo, and skipping the legacy re-read once our file exists turns the second red with `expected 'Renamed here' to be 'Renamed there'`. Both are the consequence itself.

Checks: `bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts scripts/seed-dev-state.test.ts` (75 passed), `biome check` on the touched files, `bun run typecheck:node`.

Uncovered by tests, unchanged: the `index.ts` generation guard, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, `#stopRuntime`'s `finally`, and the `stop()` race.

NorbertBodziony:
Both fixed in `d8c4f37c`.

**[P1] Downgrade reconciliation replaces newer host data wholesale** — `legacyImport` now keeps the imported record itself (a `structuredClone`, so later edits to the live host cannot drift the baseline), and a changed fingerprint runs a three-way merge instead of a swap. `reconcileHost` takes what either side changed against the baseline for the identity fields, and `reconcileEntries` merges members, invites and sessions by id: an entry either side added is kept, and an entry present in the baseline but gone from one side stays removed — so a session signed out in one build is not handed back, and a member who joined in the other is not lost. The pre-merge host is still filed under `replacedHosts`, now as an archive beside a merge that keeps the data rather than the only home for it. Where no baseline was recorded there is nothing to reconcile against, so this build's host stays active and the other copy is filed rather than dropped.

**[P2] Switching away and back can leave the original account unbound** — `applySignedInAccount` only short-circuits when the store is still bound (`store.configured`), so after `unbindChangedAccount` cleared it, reporting the same account activates it again instead of assuming a host that is already running. A repeated sign-out is left alone, which keeps the password-configured development host from being deactivated by an announce it did not cause.

Tests:
- `src/main/team-store.test.ts` — the reconcile test now renames the host in the older build *and* accepts an invitation in this one, asserting both survive the restart (`"Renamed there"` with `["Owner", "Alice"]`).
- `src/main/team-api-server.test.ts` — new `puts the account's server back when the same account is reported after the unbind`.

Watched both fail: restoring the wholesale replacement turns the first red with `expected [ 'Owner' ] to deeply equal [ 'Owner', 'Alice' ]`, and forcing `stillBound` true turns the second red with the status still unconfigured. While writing the first I found the baseline was aliasing the live host — the merge saw Alice on both sides and dropped her — which is why the baseline is cloned; the test caught it before the fix was finished.

Checks: `bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts scripts/seed-dev-state.test.ts` (76 passed), `biome check` on the touched files, `bun run typecheck:node`.

Uncovered by tests, unchanged: the `index.ts` generation guard, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, `#stopRuntime`'s `finally`, and the `stop()` race.

NorbertBodziony:
All three fixed in `79ad6f0d`.

**[P1] Reconciliation can silently re-enable a disabled member** — an entry both builds changed is no longer taken whole from either side. `reconcileFields` settles each field against the baseline: the side that moved it wins, this build's value stands when the other side did not touch it, and a field dropped on one side goes only if this build never set it. So a member disabled here keeps `disabled: true` while the rename made in the other build is still adopted. The removal policy for whole entries is unchanged. (The first shape of this used `as T` twice; `noUnsafeTypeAssertion` was right to reject it, so the baseline is now carried as the entry itself rather than a JSON string and the field walk goes through `Object.entries` — no assertion anywhere.)

**[P2] Host credentials remain attached when accounts switch directly** — `CentralAuthManager` tracks `#sessionAccountId`, set whenever the state becomes `signed_in` and cleared with the stored session. `verifyEmailCode` drops `#teamHostTokens` when the verified account differs from it, so a direct switch cannot persist the previous account's machine credential beside the new session. Held apart from `#state` deliberately: that passes through `code_sent` on the way to the other account, which is why reading the previous account off it did not work — the first version of this fix was green in the wrong way and the test caught it.

**[P2] A superseded start can republish the previous host as online** — `unbindChangedAccount` now increments `#runtimeGeneration`, so a start in flight aborts at its next checkpoint instead of reporting the previous host online (or its failure as an error) on the new account's status, and `start` re-checks supersession after the `setEnabledOnLaunch` write before emitting `phase: "online"`.

Tests:
- `src/main/team-store.test.ts` — new `keeps a member disabled here when the other build only changed their name`. Watched it fail: taking the other side's entry whole turns it red with `"disabled": false`, which is the restored login access itself.
- `src/main/central-auth-manager.test.ts` — new `does not keep one account's host credential beside the next account's session`. Watched it fail: the stored file held `machine-token-for-the-first-account-0001` beside `session-two`.

The third fix is **not covered**, and I would rather say so than fake it: reaching that path needs `start()`, which requires a `TeamWebRtcBridge` — a concrete class, so a double for it costs an `as unknown as` assertion in the harness. That is the escape hatch this repo asks me to avoid, and the guard is not worth buying one.

Checks: `bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/central-auth-manager.test.ts scripts/seed-dev-state.test.ts` (100 passed), `biome check` on the touched files, `bun run typecheck:node`.

Uncovered by tests, unchanged otherwise: the `index.ts` generation guard, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, `#stopRuntime`'s `finally`, and the `stop()` race.

NorbertBodziony:
Both fixed in `3a0990b4`.

**[P1] Superseded activation can temporarily expose the previous account's host** — `applySignedInAccount` now re-reads `#signedInAccountId()` on the far side of the runtime teardown and returns without activating when another account has been announced meanwhile. `#boundAccountId` is left `undefined` by the existing `finally`, so the account that is actually signed in still gets its host bound by its own queued switch — the guard closes the window without creating a stuck-unconfigured one.

**[P2] A transient logo-copy failure becomes permanent after migration** — the copy is no longer tied to the start that took the record in. Every initialization re-runs `#adoptLegacyLogo` for the imported host, with `COPYFILE_EXCL` so it can never write over this build's own newer image, which means a copy that failed once is retried while the legacy asset is still sitting there.

Tests:
- `src/main/team-api-server.test.ts` — new `does not bind the previous account's server when its switch is applied too late`: A configures, B signs in, then A's queued switch arrives; the status stays unconfigured and `listMembers()` throws. Watched it fail with `expected true to be false` — A's host bound and readable under B.
- `src/main/team-store.test.ts` — new `copies the logo of an imported host on a later start when the first copy could not`: the image is renamed away across the import and put back before the next start. Watched it fail with `ENOENT` on the host's logo path, which is the missing image itself.

Checks: `bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/central-auth-manager.test.ts` (96 passed), `biome check` on the touched files, `bun run typecheck:node`.

Uncovered by tests, unchanged: the `index.ts` generation guard, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, `#stopRuntime`'s `finally`, the `stop()` race, and the supersession check inside `start()` (that one needs a `TeamWebRtcBridge` double, which costs an `as unknown as`).

NorbertBodziony:
Both findings on `3a0990b4` are fixed in `9b3d50e0`.

**P1 — recycled address takes ownership through the reconcile.** Confirmed: `reconcileFields` treated `accountId` like any other field, so a build without accounts — which binds a member by address alone — could write a new account's id onto the owner and this build would import it. `reconcileMembers` now holds that one field back: an account id is taken from the other side only for a member this build knows none for. Everything else about a member still merges field by field, so the disabled-member case still holds.

**P2 — a password change committing after it reported failure.** Confirmed, and worse than the report in one way: `#persist` requires *some* active host, so if another account had already been activated the write went through immediately, onto the previous account's host. `#requireUnchangedState(state)` after the credential awaits in `changePassword`, `login` and `acceptInvite` refuses the mutation outright.

Both new tests watched failing for the right reason with the fix reverted:
- `expected { …(6) } to be null` — account B activating and getting account A's host.
- `promise rejected "The username or password is incorrect." instead of resolving` — Alice's password gone after `changePassword` told her it failed.

`bun run test:desktop -- src/main/team-store.test.ts src/main/team-api-server.test.ts src/main/central-auth-manager.test.ts` → 98 passed. `biome check` and `bun run typecheck:node` clean.

Still uncovered by a test, unchanged from the last round: the generation guard in `index.ts`, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, `#stopRuntime`'s `finally`, the `stop()` race, and the supersession check inside `start()` — that last one needs a `TeamWebRtcBridge` double, which costs an `as unknown as` I'm not buying.

NorbertBodziony:
Both findings on `9b3d50e0` are fixed in `49b2ee5c`.

**P1 — half-stopped API after a failed remote-screen teardown.** Confirmed. The HTTP server now comes down in a `finally` and the remote-screen failure propagates, so `stop()` still reports what went wrong but never leaves a listener behind for `start()` to hand back.

New test `closes its listener when the remote screen cannot be stopped`, watched failing for the right reason with the `finally` removed: `promise resolved "Response { status: 200 … url: 'http://127.0.0.1:61673/v1/compatibility' }" instead of rejecting` — the stale listener answering after the failed stop. The double is a plain object literal, no assertions.

**P2 — `replacedHosts` growing without bound.** Confirmed, and the reconcile branch was the worse half: it archived the pre-merge host on every downgrade round-trip even though the merge is precisely what keeps both sides, so that copy was pure write cost. Dropped. The no-baseline branch — the only one that would otherwise lose work — now keeps a single record rather than appending. No test: nothing reads the field, and the merge behaviour it sat beside is already covered by the three reconcile tests.

I did not add a restoration path. That is a real gap and I would rather name it than build a half of one here: the field exists so a downgrade round-trip on a file with no import baseline does not silently discard the other copy, and recovering it today means editing the JSON by hand.

`bun run test:desktop -- src/main/team-api-server.test.ts src/main/team-store.test.ts src/main/central-auth-manager.test.ts` → 99 passed. `biome check` and `bun run typecheck:node` clean.

Still uncovered by a test, unchanged: the generation guard in `index.ts`, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, the `stop()` race, and the supersession check inside `start()` — that last one still needs a `TeamWebRtcBridge` double costing an `as unknown as`.

NorbertBodziony:
The review gate went green on `49b2ee5c` — no unresolved P0 or P1 findings, and every CI job passes. The one P2 it left is fixed in `0aedc47c`.

**P2 — a teardown error skipping the drain.** Confirmed. The three steps shared one `try`, so a `#stopRuntime()` rejection jumped straight to activation with the superseded start still pending, and the next account's `start()` would be handed that operation instead of starting its own host — private despite its own saved `enabledOnLaunch`. Each step now runs through `#attemptTeardown`, which reports a failure and lets the remaining steps run.

No test, for the reason I have named on the last few rounds rather than working around: reaching either `#stopRuntime()` or `#startOperation` from the harness needs `start()`, and `start()` needs a `TeamWebRtcBridge` — a concrete class, so a double costs an `as unknown as`. I would rather leave this one uncovered and visible than buy the escape hatch. It is the same gap as the supersession check inside `start()`; a seam that takes a bridge interface would close both, and that is a change to `HostService`'s constructor contract rather than to this PR.

Everything else uncovered is unchanged: the generation guard in `index.ts`, the session write chain, the `configure` publish gap, the recheck after `syncRemoteDirectory`, and the `stop()` race.

`bun run test:desktop -- src/main/team-api-server.test.ts src/main/team-store.test.ts src/main/central-auth-manager.test.ts` → 99 passed. `biome check` and `bun run typecheck:node` clean.

PR #218, opened 2026-09-04
refactor(main): decompose remote-server-manager.ts

NorbertBodziony:
## What

`src/main/remote-server-manager.ts` was 2828 lines, the largest file in the repo, and
`src/main/AGENTS.md` named it: *"remote-server-manager.ts is the outlier here; splitting it is its own
change, but do not add a concern to it because it is already the file that has too many."* This is
that change.

**970 lines now**, plus 19 siblings with one reason to change each. Thirty commits, in rising order
of risk — read them one at a time; the whole diff at once is not reviewable.

## Three defects the structure was hiding

**1. Silent aliasing of persisted state.** `#requireServer` returned a *live reference* into
`#state.servers`, and six places mutated the on-disk state through it with no `this.#state =` in
sight. It worked only because `#persist()` happened to `structuredClone` at call time; a mutation
with no `#persist` after it was invisible until an unrelated later write. `tsc` could not see any of
it. `RemoteServerDirectory` now returns `Readonly<StoredRemoteServer>`, so each of the six became a
compile error and then a named intention (`store.update(...)`, `client.probeRemoteDesktop(...)`, an
`onServerIdentity` callback). **After e036cdee there is no way to change persisted state except
through the store, and `tsc` enforces it.**

**2. User data loss (e0e1765c).** `readStoredRemoteServers` returned `null` for the *whole file* if
one entry failed validation. `initialize` then kept its empty default and the first `#persist()`
replaced `servers.json` — every joined server, gone. No test reached that path. One bad entry now
costs the user that entry; if `activeServerId` was one of the dropped ones it falls back to the local
server. Against root AGENTS.md "Preserve all user data".

**3. One bridge doing three jobs.** `#applyConnectionError` classified an error, wrote state, *and*
synchronously tore down a WebSocket — which is why a failed HTTP request could close the event
socket. It is now a pure table (`classifyRemoteConnectionError`), a write
(`RemoteServerConnections.reportError`), and one callback in the manager's constructor. The request
path never names the event stream and the event stream never names the error path.

Callbacks rather than a second `EventEmitter`, and that is a correctness requirement: today
`#applyConnectionError` runs *synchronously* before `socket.close(1003)`, and the reconnect guard
reads a `protocolFailed` set earlier. An emitter hop would reorder that. Two tests pin the sequence.

## Signature change

`request<T>` becomes `(serverId, path, decoder, init?)` — e7b7669e, nothing else in the diff.
`typecheck:node` is the whole guard: a `string` in position 2 cannot be an options object, so a
missed call site cannot compile.

## Tests

Four new files, each for a boundary that did not exist before: `remote-server-store.test.ts`
(persisted state, rule 7, at the lowest boundary — `servers.json` was previously only reachable
through `manager.initialize()`), `remote-server-event-stream.test.ts`,
`remote-server-host-directory.test.ts`, `remote-server-connection-status.test.ts`. No
`remote-*-decoding.test.ts`: those are pure functions over shapes `tsc` already checks, and the two
consequences worth naming already had tests. The header of `remote-host-decoding.ts` says why, so
nobody adds forty assertions later.

`remote-server-manager.test.ts` goes 1843 → 762 lines, split by moving whole `describe` blocks to the
file whose behaviour they actually pin. Four invite-link tests are deleted outright —
`packages/contracts/src/invite-links.test.ts` already covered them, and the one assertion the plan
thought was unique turned out to be covered too (`validation.test.ts:67`), so nothing was added.

`remote-server-test-harness.ts` (shape copied from `src/backend/agent-service-test-harness.ts`, a
plain `.ts` contributing no `it`) fixes two real defects, not just duplication:

- The suite had **eight** WebSocket fakes and **five declared `readonly readyState`**, so closing one
  left it reading OPEN — both `readyState !== WebSocket.OPEN` guards in the event stream were
  unreachable from the tests that looked like they covered them. `FakeEventSocket.close` always moves
  `readyState`, and dispatches `close` in a microtask because the stream's `error` listener calls
  `close()` and *then* rejects; a synchronous close event would resolve that promise first.
- `stop()` is async and every test called it bare in a `finally` right before `rm(directory)`, so a
  pending write could land after the directory was gone. `stopRemoteFixtures` awaits it once.

`deferredRoute().arrived` replaces the suite's only `setTimeout` wait — a zero delay that slipped
past `no-sleep-in-tests.grit` because the pattern starts at 1.

Header assertions that used to sit *inside* `fetchMock` bodies moved onto
`stubTeamFetch(...).requests(path)`: an `expect` in a mock reports the mock's source location and
cannot fail when the route is never reached.

### Watch it fail

**Every** moved or new test has a named one-line mutation that reddens it, listed in the commit that
moved it. Two findings came out of doing it rather than claiming it:

- In `#connectEvents`, removing `!this.#servers.has(serverId)` *or* `controller.signal.aborted`
  alone stays green — `remove()` aborts the controller, so the abort check already blocks the socket.
  Both had to go to make the test red. Same for the auth-pause check in `scheduleReconnect` and the
  `!protocolFailed` clause in the reconnect guard. They are redundant on today's paths and are
  defence in depth against a future status table that calls a protocol failure retryable. Moved
  verbatim; left as they are.
- `suspendReconnect` no longer aborting the controller is the check that the registry→stream hook
  survived the split. It goes red. The hook is intact.

## Deliberately not touched

Released `team-protocol` v1/v2/v3 adapters and their intended duplication — HTTPS frames V1 or V3 by
negotiation, WebRTC frames V2 JSON, and "unifying the transports" means deleting duplicated *call
sites*, never merging two wire encodings. `remote-server-client.ts`'s header says so, because that is
the one edit that breaks a shipped protocol. The control-plane methods stay explicit behind
`RemoteControlPlaneTransport`: they hit the **account service, not the host** — `revokeInvite(inviteId)`
does not even take a `hostId` — and hiding two authorities behind one method would cost the next
reader an hour. Also untouched: the legacy no-`appVersion` socket branch, `LOCAL_SERVER_ID` special
casing, the duplication idempotency protocol, `webRtcRequestBody`'s compatibility shim, and the
public `EventEmitter` map the renderer subscribes to.

## Surfaces

Desktop main process only. No renderer, mobile, public web, palette, IPC contract, migration or
privacy change. `src/main/AGENTS.md` is rewritten where it described the old shape.

## Approvability

No `biome-ignore`, `@ts-expect-error`, `@ts-ignore`, `biome.json` override or widening to `any`.

## Checks

Per root AGENTS.md, no repo-wide runs. After every commit: the affected test files via
`bun run test:desktop -- <path>`, `biome check <touched paths>`, and `bun run typecheck:node`. The
pre-commit hook also runs all eight `typecheck:*` projects. CI owns the full suite.


NorbertBodziony:
Fixed in `55f65558`.

The finding is right, and it is the other half of the data loss `e0e1765c` claimed. That commit fixed the reader dropping one unreadable *entry*; it left the case where the file itself makes no sense — an unknown `version`, or no `servers` array — returning null into a `load()` that kept its empty default and resolved. `initialize()` then syncs WebRTC hosts and persists over it.

Both file headers already asserted this could not happen (`remote-server-stored-shape.ts`: an unknown version "is refused outright … guessing at it would replace a file that build can still read"; `load()`: a failed read reaches the caller "because continuing would leave an empty list that the next write replaces the user's servers with"). The prose was correct and the line under it was not. `load()` now throws — the same thing a corrupt file has always done, since `JSON.parse` already threw from that line.

The review also asked for the store-level load-and-persist path rather than the decoder alone. `src/main/remote-server-store.test.ts` is new for exactly that boundary, against a real path. Watched both tests fail:

- restore `if (stored) this.#state = stored` → `promise resolved "undefined" instead of rejecting`
- make the reader return null again for one bad entry → `The remote server list is not in a format this version can read.`

/norbiai review

NorbertBodziony:
All three fixed in `15a3e83f`.

**[P1] Unreadable entries are no longer deleted.** The finding is right that skipping was the same data loss one write later: an entry with an intact id, apiUrl, fingerprint and encrypted token but one field this build rejects was left out of the list and then written away by the next `persist()`.

I did not take the throw/write-blocked option, because refusing the whole file for one bad entry is the bug `e0e1765c` fixed and it costs the user every *other* server. Instead the entries are preserved: `readStoredRemoteServers` carries them in `unreadableServers` and `serializeStoredRemoteServers` puts them back into `servers` untouched, so the file keeps every entry it arrived with. In memory only — a key of its own would be invisible to exactly the older build that is the reason for keeping them, so the on-disk format does not change. An entry whose id a real server later takes is dropped, since two entries sharing an id would leave the file disagreeing with itself. A non-record entry is not preserved: it carries no token and no identity.

This build still refuses to *use* such an entry. It no longer gets to delete it. The shape test now asserts the entry survives, so it no longer expects silent deletion.

**[P2] Login no longer fails on the probe.** `probeRemoteDesktop` rethrows anything that is not 404, 426 or 503, so a timeout or a 500 on an optional screen-sharing probe rejected `login` and set the server to `error` — over credentials already on disk and working. I took the second option: the probe is best-effort, the flag keeps its previous value, and `#refreshRemoteDesktop` corrects it later. Persisting once at the end would not have helped, since the probe has to authenticate with the token the sign-in just wrote.

This one ships without a test, deliberately: `login` has no coverage at all today, and the first test needs an identity-proof fixture that generates a keypair and signs a live challenge — a larger change than the two-line fix, in a flow this PR does not otherwise touch. Flagging it rather than hiding it.

**[P3] Coalescing now asserted once.** Removed the 65-line integration case. What it uniquely covered moved down to `RemoteEventRefresh`: retrying a failed conversation refetch for a revision announced while it was away, and the queue path's own coalesce-and-retry — which the unit suite did not reach at all, so deleting the integration test alone would have lost it.

Watched all four new/changed tests fail:

- stop collecting unreadable entries → both the shape and store tests redden (`expected [] to deeply equal [ { id: 'corrupt' … } ]`)
- drop the id-collision filter → `expected [ { id: 'beta' … }, …(1) ] to deeply equal [ ObjectContaining{…} ]`
- `if (false) continue` on the conversation retry guard → the retry test blocks on its second request
- delete `if (request.dirty) continue` in the queue path → the queue test blocks the same way

/norbiai review

NorbertBodziony:
Round 3 addressed in `736933a9`.

**[P1] collision filter deleted preserved entries during reconciliation** — confirmed, and the mechanism is exactly as described: an unreadable WebRTC entry is absent from `store.servers`, so `#syncWebRtcHosts` hands `reconcileWebRtcHosts` a list without it, the directory advertisement looks new, and `replaceServers` minted an entry whose id then tripped the filter. Fixed by removing the filter rather than by guarding it: serialization now writes every preserved entry, and `RemoteServerStore` retires one only where intent is known — `remove` (the user asked) and `adopt` (joining or signing in verified the host identity). `replaceServers` never does. A duplicate id on disk is inert; the reader splits readable and unreadable entries into separate lists.

**[P2] active selection discarded for a preserved entry** — fixed with `unreadableActiveServerId`: the file keeps naming the server the user was on while this build runs locally, and only `setActiveServerId` clears it.

**[P3] preservation tested twice** — the parser-level case is gone. The store test asserts it on disk; the shape file keeps only what the reader alone decides.

Watched five mutations fail, each with its own diff — listed in the commit message.

/norbiai review

NorbertBodziony:
Round 4 addressed in `e0cac16c`. All three confirmed.

**[P2] select() rollback** — right, and the rollback was the only half-applied path left: `setActiveServerId` clears the retained id, so restoring just the active id left it gone and the next unrelated persist wrote the local fallback as the user's choice. Selection is now one value — `store.selection` before, `store.restoreSelection(previous)` on failure — rather than an id a caller can restore half of.

**[P2] WebRTC rejoin** — right, that path gets its entry from `#syncWebRtcHosts` → `replaceServers` and never reaches `adopt`. `store.retireUnreadable(hostId)` is its half, called immediately after the fingerprint check, which is the point that distinguishes it from reconciliation.

**[P2] ordering** — right, and it undercut the preservation: `servers` order is the sidebar order the user arranged, so any unrelated write reshuffled a list this build cannot display. Each preserved entry now carries the slot it was read from and is spliced back into it, ascending, clamped to the end of a list that has since shrunk.

Watched three mutations fail — restoring only the active id, `retireUnreadable` returning early, appending instead of splicing — diffs in the commit message.

/norbiai review

NorbertBodziony:
Round 5 addressed in `697771b2`. All three confirmed; the first is the sharpest finding in this thread.

**[P1] preserved pins bypassed during reconciliation** — right, and it contradicted the file's own header. An undecodable entry is absent from `store.servers`, so `reconcileWebRtcHosts` saw no existing server and took the advertised key for a host whose fingerprint was in the file all along. `store.preservedIdentities` now carries id, key and fingerprint into the merge and `resolveHostKey` treats them as pins. I did not take the suppress-recreation option: it would drop the host from the sidebar until a manual rejoin, where honouring the pin blocks the substitution and still lets an untouched host reconnect.

**[P1] malformed WebRTC responses not recorded as protocol errors** — right, `rethrowAsRemoteRequestError` converts only `TeamWebRtcRequestError`, so a decoder failure escaped as a plain error and the classifier ignored it. Decoding now sits outside the transport catch and wraps in `RemoteProtocolError("protocol_error", …)`, matching `requestJson`. I left the transport's own adapter-decoding half alone — `team-webrtc-client-transport.ts` already raises `protocol_error` on a frame it cannot decode (its test asserts it), and it is not a file this PR touches.

**[P2] ordering** — right, an absolute index does not survive a removal. Each entry now names the entry it preceded and is spliced in front of it, falling to the end only when that successor is gone too.

Watched three mutations fail — diffs in the commit message.

/norbiai review

NorbertBodziony:
Fourth round, all three taken.

**[P1] Malformed WebRTC responses are not recorded as protocol errors** — right, and the transport half is the one I had missed. `decodeTeamProtocolV3WebRtcHttpResponse` in `requestResponse` now raises `TeamWebRtcRequestError(502, "protocol_error", …)` for a body the released adapter refuses, next to the envelope check that already did. `decodeTeamProtocolSupportV1` is wrapped in both compatibility paths as `RemoteProtocolError("protocol_error", "The host returned invalid compatibility information.")`, matching the HTTPS arm word for word. `refreshWebRtcCompatibility` also calls `connections.reportError` itself: it runs off the transport `connected` event and the manager throws its rejection away, so nothing else would have recorded it.

**[P1] Reconciliation persists duplicate records for preserved hosts** — right, and the argument that kept both records in round three no longer applies. That harm was "dropping the copy destroys the pinned key and fingerprint", and it was true when nothing else held them. It is not true now: the round-four pin fix has `reconcileWebRtcHosts` copy both onto the entry that takes the id (`fingerprint: pinnedFingerprint(existing) || advertisedFingerprint(...)`, `...(publicKey ? { publicKey } : {})`), which is the only place the pin is read from. So `serializeStoredRemoteServers` writes the readable entry and not the preserved copy, which restores the uniqueness the file needs — a duplicate id makes `reorder` reject every ordering for the build that can finally decode the entry, permanently. The preserved record stays in memory, so it keeps feeding `preservedIdentities` until a verified join retires it.

**[P2] Opaque entries can be mistaken for ordering anchors** — right. Only readable servers anchor a slot now: the walk is over `state.servers`, emitting preserved entries whose `beforeId` matches, and leftovers go last in read order. An entry this build could not read no longer gets to decide where another one lands.

Watched each fail (single-line mutation → the test it protects, with the diff):

| Mutation | Test | Failure |
| --- | --- | --- |
| `body = decodeTeamProtocolV3WebRtcHttpResponse(...)` unwrapped | reports an undecodable response body as a protocol failure | `expected Error: Invalid Team protocol v1 HTTP resp… to match object { code: 'protocol_error' }` |
| `decodeTeamProtocolSupportV1(value)` unwrapped in `#negotiateWebRtcCompatibility` | fails a host closed when its compatibility answer does not decode | `expected to throw 'invalid compatibility information' but got 'Invalid Team API compatibility respon…'` |
| drop `reportError` from `refreshWebRtcCompatibility` | same | `expected { state: 'offline' } to match object { state: 'error' }` |
| keep every preserved entry regardless of id | never writes two entries under one id when reconciliation recreates a preserved one | `expected [ { id: 'beta' }, …(1) ] to deeply equal [ ObjectContaining{…} ]` |
| restore the `findIndex` over already-spliced entries | keeps unreadable entries in the order they arrived, whatever they call themselves | the two opaque entries come back swapped |

74 tests across the nine remote-server and WebRTC transport suites pass; `typecheck:node` and `biome check` clean.

/norbiai review

NorbertBodziony:
Fifth round, both taken.

**[P1] Malformed WebRTC responses are not recorded as protocol errors** — right on both counts, and the shape of the bug is that `TeamWebRtcRequestError` means nothing to the classifier, so any call reaching the transport without translating it is invisible. `refreshWebRtcCompatibility` awaited outside its reporting `try`; `probeRemoteDesktop` called the transport directly and its callers turn every rejection into `false` or `null`. Both now go through one `#hostRequest`, which applies `rethrowAsRemoteRequestError` once, and the probe records a `protocol_error` instead of answering "no screen sharing". The same translation repairs something older: the probe’s 404/426/503 rule (`error instanceof RemoteRequestError`) never matched on the WebRTC arm at all, so a host that answers the capabilities route with 404 was throwing rather than returning `false`.

**[P1] Reconciliation permanently discards preserved records** — right, and it is the half of the previous finding I got backwards. Both records cannot go to disk (duplicate ids break `reorder` permanently for the build that can read them), so one must lose, and I picked the wrong one: only the preserved record is irreplaceable. The synthesized WebRTC entry carries `encryptedToken: ""` and nothing the user typed — the next host directory sync mints it again, with the pinned key and fingerprint `reconcileWebRtcHosts` reads out of the preserved record — while the preserved entry's unrecognised fields exist nowhere else. So serialization now writes the preserved entry, in the position reconciliation gave the host, and omits the synthesized one. The app still runs on the readable entry; it is the disk that keeps the record only a newer build can read. One entry per id either way, which keeps the previous finding closed.

Watched each fail:

| Mutation | Test | Failure |
| --- | --- | --- |
| `refreshWebRtcCompatibility` calls `this.#requireTransport().request` directly | records a protocol failure the transport raised on a route nothing else reports | `expected { state: 'offline' } to match object { state: 'error' }` |
| same, in `probeRemoteDesktop` | same | same |
| drop the probe's `reportError` for `protocol_error` | same | same |
| `servers.push(claim ? claim.entry : server)` → `servers.push(server)` | writes the preserved entry, not the one reconciliation minted over its id | `expected [ { id: 'beta', …(7) } ] to deeply equal [ { id: 'beta', …(7) } ]`, differing in the name and role only the preserved record carries |

`origin/main` merged (`65a43525`, no conflicts). 75 tests across the nine remote-server and WebRTC transport suites pass; `typecheck:node` and `biome check` clean.

/norbiai review

NorbertBodziony:
The last review ran against `65a43525`, the merge commit — the first two findings were already fixed in `bdc9dbdc`, pushed before it started. Leaving them here for the next pass to confirm:

- **[REMAINS][P1] Malformed WebRTC responses** — `refreshWebRtcCompatibility` and `probeRemoteDesktop` both go through `#hostRequest` now, which applies `rethrowAsRemoteRequestError` once; the probe reports a `protocol_error` rather than answering "no screen sharing".
- **[REMAINS][P1] Reconciliation permanently discards preserved records** — serialization writes the preserved record and omits the synthesized one, in the position reconciliation gave the host.

**[NEW][P1] WebRTC protocol failures leave the event data plane active** — right, and it is exactly the asymmetry the HTTPS test beside it was written for. `suspendReconnect` ends an HTTPS server's socket because it aborts the controller that owns it; a WebRTC host has no controller and no socket there, so its data channel stayed open and kept pushing events from a host the app had just decided it could not understand. `#suspendServer` now suspends first and then disconnects the transport, so the `disconnected` event finds the pause in place and does not schedule a reconnect around it — as you said it would.

The test is the twin of "closes the event data plane after a malformed HTTP payload", over a new `fakeWebRtcTransport` in the harness: the real class over a bridge that never connects, since the manager takes the concrete transport.

Watched it fail: dropping the `disconnect` from `#suspendServer` gives `expected "disconnect" to be called with arguments: [ Array(1) ]` on "closes the event data plane of a WebRTC host that answers with nonsense".

76 tests across the nine remote-server and WebRTC transport suites pass; `typecheck:node` and `biome check` clean.

/norbiai review

NorbertBodziony:
Both halves of the remaining P1 are now in. The review at `bdc9dbdcc0fc` predates `8609c681`, which added `#suspendServer` — it disconnects the WebRTC transport when a failure suspends reconnects, so the data channel closes with the retries.

`0bdd70bd` covers the last clause, which was genuinely open: the `disconnected` event that disconnect raises ran `setState(serverId, "offline")` unconditionally, overwriting the `error`/`incompatible` state the failure had just recorded. The HTTPS arm already guards the same call with `!protocolFailed`; the WebRTC arm now asks `RemoteEventStream.isReconnectSuspended` for the same fact. Watched it fail — restoring the unconditional `setState` reddens "closes the event data plane of a WebRTC host that answers with nonsense" with `expected { …(12) } to match object { state: 'error', …(1) }`.

NorbertBodziony:
Both findings were valid and are fixed in `2811c4b4`.

1. The probe was worse than reported: `decodeRemoteDesktopCapabilities` ran unwrapped on the WebRTC arm, *and* the catch tested `error instanceof RemoteRequestError && error.code === "protocol_error"`, which never matches the `RemoteProtocolError` the HTTPS arm's `requestJson` already raises for an unreadable body — the two classes are siblings, neither extends the other. So both arms left the host healthy. The decode now goes through `decodeOrProtocolError` and the catch names both classes.
2. `retryConnection`'s WebRTC branch now calls a new `RemoteEventStream.resumeReconnect`, which `restart(serverId, true)` also uses, so the two arms cannot drift.

Watched all three guards fail: unwrapping the probe decoder reddens the new client test with `but got 'Invalid platform.'`; narrowing the catch back to `RemoteRequestError` reddens it with `expected { state: 'offline', …(3) } to match object { state: 'error', …(1) }`; dropping `resumeReconnect` reddens the manager test with `expected { …(12) } to match object { state: 'offline' }`.

/norbiai review

NorbertBodziony:
Valid, and fixed in `224f55b5` by the first of the two suggestions: `login` now calls `restart(server.id, true)` *before* the probe, so the probe is the last writer — the order the WebRTC `connected` listener already had.

Watched it fail: moving the restart back after the probe reddens the new test "keeps a protocol failure the sign-in's last request recorded" with `expected [ StubbedEventSocket {} ] to have a length of 2 but got 1` — the socket the restart should already have opened is not there when the probe's answer arrives.

/norbiai review

NorbertBodziony:
Both fixed in `7224e703`.

1. Taken at the source rather than in the probe's catch: `requestJson` now raises `RemoteProtocolError("protocol_error", "The host returned invalid data.")` when a **successful** response body does not parse, which is the same error its shape-decode failure two lines below already raises. The message is the one `classifyRemoteConnectionError`'s `SyntaxError` branch produced, so nothing a user reads changes — but every caller that checks for a protocol failure by class now sees it, not just the classifier. Watched it fail: rethrowing the raw error reddens the new client test with `expected […] to throw error including 'invalid data' but got 'Expected property name or '}' in JSON at position 1'`.
2. The two extra copies of the comment are gone — a bad restore during the mutation check for `224f55b5`, thanks for catching it.

/norbiai review

NorbertBodziony:
Both fixed in `f43fafc6`.

1. Taken as "clear only the issue that existed when negotiation began": `ensureCompatibility` captures `issueFor(server.id)` before it starts and its `.then` calls a new `RemoteServerConnections.clearStaleIssue`, which clears only while the recorded issue is still that same one. The client's sink no longer needs plain `clearIssue` at all. Watched it fail: clearing unconditionally reddens the new test "does not let a negotiation started earlier withdraw a failure recorded while it was out" with `expected { state: 'error', …(3) } to match object { issue: { code: 'protocol_error' } }`.
2. The probe and the write of its answer are now one best-effort chain with a single `.catch`, so neither can reject a sign-in whose token is already on disk. This one is stated rather than tested: reproducing it needs a `store.update` that fails mid-test, and every way to force that (removing the state directory under the fixture, chmod) is a race or platform-specific rather than a test of the consequence.

/norbiai review

NorbertBodziony:
I think this one should be withdrawn, and I have corrected the comment it is measured against in `05481217`.

The concrete claim is right: `update()` publishes in memory before `persist()` resolves, so a failed write leaves the new `remoteDesktopAvailable` in `list()` and off disk. My comment said "keeps its previous value", which is true when the *probe* rejects and not when its *write* does; it now says so.

What I do not want to do is the fix — publishing store mutations only after the queued write succeeds. Applied to `update`, it makes the important case worse than the cosmetic one it repairs. The write that matters on this path is the one immediately above: the session token this sign-in just minted. If that write failed and the store rolled the value back, the app would keep using the *old* token — which the host has already invalidated — while the token that actually works is discarded. Publishing first is what keeps a failed disk write from also breaking the live connection, and `update` returns the updated view precisely so the caller can use those credentials for its next request.

The drift it leaves is bounded and one-directional: memory is ahead of disk until the next successful write, which then commits it. That is how every mutation in this file behaves (`adopt`, `remove`, `reorder`, `unhideHost`), and how the manager behaved before the split — making one of them transactional and not the others would be the inconsistency. For `remoteDesktopAvailable` specifically the committed value is also self-correcting: `#refreshRemoteDesktop` re-probes and rewrites it.

Turning the store transactional is a coherent change, but it is a design change to persisted-state handling with its own failure modes, not a fix to this PR's decomposition, and it should be its own commit with its own tests rather than an addendum to a sign-in comment.

/norbiai review

PR #232, opened 2026-09-04
refactor(backend): extract AgentService core into eight controllers

NorbertBodziony:
## What

Continues #217: `src/backend/agent-service.ts` goes from 2472 to 1241 lines. The facade keeps the public API (`new AgentService(...)` call sites untouched) and event emission; each extracted controller owns one concern behind a `*`Hooks interface and never imports the facade:

| Controller | Owns |
|---|---|
| `ImageGenRuntime` | streaming image-gen items, interrupt flags, operation map |
| `DeltaBuffer` | delta coalescing (100 ms debounce / 8 KiB cap), per-turn flush |
| `MailboxSync` | mailbox↔snapshot merge, `emitQueue`, delivery retry |
| `BootRecovery` | unresolved deliveries, persisted-turn recovery, provider-history backfill |
| `AttachmentGateway` | `attach_files_to_response` + TOCTOU-safe file open + in-flight dedup |
| `ThreadLifecycle` | ensure/resume/retire/recovery, handoff budget, runtime refresh |
| `DrainScheduler` | mayDrain guard composition, schedule/drain/startDelivery (takes `ThreadLifecycle` directly) |
| `TurnLifecycle` | notifications, turn completion, association, agent-to-agent relay |

`#emit`/`#emitError`/`#emitRuntimeSnapshot` stay on the facade (one-liners over `EventEmitter`). No migrations, no IPC/contract changes, no renderer/mobile surfaces.

## Verification

- `bunx tsc --noEmit -p tsconfig.node.json` clean; commit-hook `typecheck:*` (all projects) green
- `bunx biome check src/backend/agent-service.ts src/backend/agent/` clean (35 files)
- Narrow suites, all green: `agent-service.queue` (23), `restart` (8), `routines` (15), `providers` (11), `questions` (2), `agent/provider-runtime` (8), `agent/routine-scheduler` (6)
- Watch-it-fail per slice, each red for the intended reason before restore (mayDrain→16 red, handoff summary, attachment isWithin boundary, ack-failed-turn, delta/image paths). One lesson: breaking `activeTurnId` clearing hangs the suite instead of failing fast — the barriers hold, just slowly.

Review follow-up (a2c94bad): restored the `sortConversationMessages` call `AttachmentGateway` dropped between the message push and the persist, and removed three hook members with no readers (`ThreadLifecycle.setPendingHandoff`, `DrainHooks.emit`, `ImageGenHooks.emit`) with their facade wiring.

Known gaps the move exposed (no new tests added per the default-no-test rule): the `queue-changed` emit line, `syncDeliveryMessage` turnId sync, the duplicate-attachment-path guard, and attachment ordering within a turn are not pinned by the suite — candidates for follow-up assertions.

Model/harness: Muse Spark via opencode, backend node vitest only.

### DIFF
--- a/src/backend/agent-service.queue.test.ts
+++ b/src/backend/agent-service.queue.test.ts
@@ -472,6 +472,8 @@ describe.sequential("AgentService: queue", () => {
       clients.set(provider, client);
       return client;
     });
+    const events: AgentEvent[] = [];
+    service.on("event", (event) => events.push(event));
     await service.initialize();
 
     await service.sendMessage({ botId: "chief", text: "Question 1" });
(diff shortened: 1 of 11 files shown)


NorbertBodziony:
Both remaining findings are addressed in a2c94bad.

**P2 — attachment sorting.** Confirmed against pre-split `agent-service.ts:1383`: the extraction dropped the `sortConversationMessages(snapshot.messages)` that sat between the `push` and the persist. Restored in the same position. The catch path already used `findIndex`, so it is sort-safe.

**P3 — dead API surface.** `ThreadLifecycle.setPendingHandoff`, `DrainHooks.emit` and `ImageGenHooks.emit` had no readers; removed along with their facade wiring and the `AgentEvent` imports that became unused in `drain-scheduler.ts` and `image-gen-runtime.ts`.

No test added for the sort: nothing in the suite pinned attachment ordering before the split either, and reproducing it needs a mailbox append interleaved into the awaited file opens. It stays on the known-gaps list in the description.

Verification: `bunx tsc --noEmit -p tsconfig.node.json` clean, commit-hook `typecheck:*` (all projects) green, `bunx biome check src/backend/agent-service.ts src/backend/agent/` clean (35 files), and the seven narrow suites green (73 tests: queue, restart, routines, providers, questions, agent/provider-runtime, agent/routine-scheduler).

Model/harness: Claude Opus 5 via Claude Code.

/norbiai review

PR #239, opened 2026-09-04
refactor: finish the bot to agent rename across code, data and copy

NorbertBodziony:
Finishes the `bot` → `agent` rename the repository has carried half-done for a long time. The product
concept is called an agent everywhere now — identifiers, IPC channels, CSS, user copy, the mobile app,
and the instructions and tool parameters the models read — and the `AGENTS.md` "expect both" paragraph
is gone.

## Decisions this implements

All five were the developer's, taken after the costs below were laid out:

1. **Scope**: desktop + contracts, `apps/mobile`, user-facing copy, and the model-facing surface.
2. **Persisted data**: migrate everything, including rewriting existing `bot-<uuid>` id *values*.
3. **Marketplace collision**: local agent takes `agentId`; the marketplace offer becomes `listingId`.
4. **Delivery**: one PR.

### The accepted cost: paired phones must be re-paired

Rewriting id values invalidates identifiers other devices hold. After the host updates, a paired phone
asks for `GET /v1/agents/bot-7b62…` and gets a 404; the wire protocol is frozen, so no "your id changed"
event may be sent. Mobile degrades gently — it shows the existing *"This agent is no longer available on
the selected server."* — but pins and unread cursors in `SecureStore` go dead. **`CHANGELOG.md` says
plainly that phones must be unpaired and paired again.**

## What is deliberately *not* renamed

Each survivor is nameable, and the new `AGENTS.md` "bot" bullet names them:

- **The Team API v1–v3 wire.** `botId`, `bots`, `bots-changed` stay on the wire forever. The new
  `packages/contracts/src/team-protocol/current-agent-keys.ts` is the only translator.
- **`bots.json`, `mailbox.json`, `legacy-import:bots:v1`.** Read paths and a `command_id` a shipped
  release already wrote to the user's disk. Renaming the first two loses data; renaming the third re-runs
  the legacy import on every install.
- **`~/OpenBot/Bots/…` path prefixes** in `workspacePathFromInput`, and `"first-bot"` avatar seeds — the
  seed is an input to the face generator, not a word, so changing it would restyle every existing agent.
- **`BloubBot`, lucide's `Bot`**, the crawler-detection regex, and the fixture human names "Nöra Bot" /
  "Nörbert Bot".
- **`avatarSeed` values** keep whatever they hold, even when a seed equals an old id. Renaming them would
  change every existing agent's face and diverge from `marketplace_agent_versions.avatar_seed` already
  published in D1.

## The silent-wire risk, and what closes it

`TEAM_PROTOCOL_V1_EVENT_KEYS` is a key allowlist and `projectV1Object` **drops** anything not on it, while
`v1-adapter.ts` bridges types with a bare cast. Renaming `AgentEvent.botId` alone would leave `tsc` green,
the codec silently dropping the field, `encodeTeamProtocolV1CurrentEvent` returning `null`, and remote
clients simply not receiving conversation deltas or turn events — no exception, no red test.

`current-agent-keys.ts` was added and wired into all six `*Current*` entry points of `v1-adapter.ts`
(including the two paths that bypass the codec: the `turn-progress` branch and `isInstalledSkillsRoute`)
plus `v3-adapter.ts`'s own branches, **with an identity map first**, so that commit is a provable no-op
against untouched fixtures. Two entries invert the naive global map:

- `marketplaceSource` maps the other way — on the wire `agentId` means a marketplace *listing*.
- The sidebar-layout subtree already said `agentId` on both sides and is untouched.

## Migrations

- **v12** `migrateReactionsForAgentActors` — `projection_reactions.actor_bot_id` → `actor_agent_id` (part
  of the primary key, so a table rebuild), and `actor_kind` `'bot'` → `'agent'` with the `CHECK` updated.
  `LATEST_SCHEMA_SQL` stops being an alias of the v8 baseline, so
  `openbot-database-schema-parity.test.ts` is a real guard here rather than a no-op.
- **v13** `rewriteGeneratedAgentIds` — rewrites every `bot-<uuid>` to `agent-<uuid>` across id columns,
  `thread_id` in the parent and all seven child tables, `orchestration_events.aggregate_id` **and**
  `command_id` (no FK protects `aggregate_id`, and `hosted-site-event-log.ts` reconstructs
  `markerCommandId` as a string), and the JSON blobs. Key *renames* use `json_set`/`json_remove`, never
  textual `replace()` — a message where the user or the model quotes an identifier is realistic in a
  coding app. Id *values* use `replace()` on the full `bot-<uuid>` token, which is what makes historical
  messages quoting a workspace path point at the new location. Runs with `disableForeignKeys: true` and
  `vacuumAfterCommit: true`.
- **v14** `refreshProviderSessionsForDynamicTools` — runtime only, because the tool parameter names change.

**Honest about the cost:** v13 scans the whole chat history in one transaction at startup with no progress
UI, on a database with no backup. `runMigration` rolls the whole thing back on a throw, so an interrupted
upgrade leaves the pre-migration database.

**Tolerant readers stay** regardless of the migration, because a database the user restored from their own
file copy never ran v13: `thread-replay.ts`, `hosted-site-event-log.ts`, `mailbox-store.ts` (a new
`toCurrentMailboxState` pre-pass, because its validators threw *before* normalization and hard-blocked
startup) and a `COALESCE` in `agent-roster.ts`'s SQL all read `agentId ?? botId` and write only the new
name.

## The workspace move

`~/OpenBot/Bots/bot-<uuid>` → `~/OpenBot/Agents/agent-<uuid>` happens in `AgentStore.initialize()`, after
the database migrations — not inside one, because `runMigration` rolls its transaction back on a throw and
a directory that already moved cannot be rolled back. It is resumable rather than atomic: on one volume
`rename` is a single syscall, and what an interrupted run already did is read from whether the target
exists, never from a marker file. `EXDEV` does **not** fall back to copying — an interrupted copy of a
40 GB workspace is lost data, so the workspace stays put with its stored absolute path still correct.
Failing here never blocks startup. `deleteAgent` now also removes `agent.workspacePath`, since the EXDEV
fallback legitimately leaves a workspace outside the derived root.

Absolute paths the provider's own transcript still names under the old root are rebased once via
`rebaseLegacyWorkspacePath` and then go through the **unchanged** `isWithin` containment check — the
fallback finds the file again, it does not open a second way out of the workspace.

## Watch it fail (rule 2)

A mechanical rename has almost nothing to break honestly, so the three (plus two) places that carry real
new behaviour were each broken and watched go red for the intended reason:

1. **`current-agent-keys.ts`** — deleted the `agentId → botId` map entry; `v1.test.ts` went red on a
   wire-shape assertion.
2. **Migration v13** — dropped the `thread_id` update in one child table; the new migration test went red
   on the missing message, not on a column name.
3. **The routine tool allowlist** — reverted the five `"agentId"` literals in `routine-scheduler.ts`;
   `routine-scheduler.test.ts` went red with `Unexpected routine argument: agentId.` on two tests, proving
   the five allowlist edits were required.
4. **The mailbox tolerant pre-pass** — removed `toCurrentMailboxState`; `mailbox-store.test.ts` went red
   with *"Stored mailbox projection is invalid."* on the old-shaped state.
5. **`#reconcileWorkspaceDirectories`** — early-returned at the top; the new `agent-store.test.ts` case
   went red with `ENOENT … /home/OpenBot/Agents/agent-a26c98c0-…/notes.md` — the workspace never followed
   its stored path.

Exactly two new `it` blocks were added (both in `openbot-database.test.ts`, twins of the existing
migration/rollback pair) plus one in `agent-store.test.ts` for the resumable move. Everything else extends
an existing test.

## Deviations from the plan

- **v13 is value-only.** The column rename it originally carried is all in v12; v13 touches values.
- **The codemod is a single commit.** A pre-commit hook runs Biome and all nine typechecks on every
  commit, so splitting the codemod per directory would have left intermediate commits that cannot pass
  their own hook.
- **The copy pass runs before the codemod**, so the decapitalization ("Automatic Bot color" → "Automatic
  agent color") is reviewable on its own instead of hiding inside 292 files.
- **The model surface was absorbed into the codemod** rather than kept as its own commit, for the same
  hook reason; the atomic units A (`agentId`) and B (`recipientAgentIds`) each still moved with their
  schemas, allowlists and the `drain-scheduler.ts` prompt that dictates the parameter name to the model.
- **Marketplace scope widened** to `marketplaceSource.listingId`, which the plan had only as a wire note.
- **Mobile needed no wire-boundary carve-outs.** The plan expected five; the five turned out to be
  current-shaped already, since everything mobile reads passes through the v2/v3 adapters.
- **New: `TeamProtocolV1CurrentClientEvent`**, because the shim needed a name for the current-shaped side
  of the boundary that the frozen types could not supply.
- **New: `toCurrentMailboxState`**, the tolerant pre-pass described above.
- **The legacy `bots.json` import path deliberately keeps its `bot-` ids.** That import runs *after* the
  migrations in `AgentStore.initialize()`, so those agents never see v13. Rewriting their ids there would
  dangle the `mailbox.json` import that follows and references the same ids — losing user data. Their
  workspace paths already exist, so `#reconcileWorkspaceDirectories` skips them and the legacy root is
  correctly not removed. `isGeneratedAgentId` accepts both prefixes for exactly this reason.

## Surfaces touched

Desktop renderer, mobile (`apps/mobile`), IPC contracts (`packages/contracts` **and**
`src/renderer/src/preview/mock-openbot.ts`), `packages/team-client`, `packages/brand`, `apps/auth-api`,
migrations plus the separate latest schema, and documentation (`AGENTS.md`, `src/backend/AGENTS.md`,
`src/renderer/AGENTS.md`, `README.md`, `apps/mobile/README.md`, `PRIVACY.md`, `CHANGELOG.md`).

Five IPC channels moved, because the namespace *is* the entity: `agent:list-bots` → `agent:list`,
`agent:create-bot` → `agent:create`, `agent:duplicate-bot` → `agent:duplicate`, `agent:update-bot` →
`agent:update`, `agent:delete-bot` → `agent:delete`. `ipc-channel-coverage.test.ts` derives its list from
`Object.keys(IPC_CHANNELS)` and would have caught three of four moving.

`Sidebar.test.tsx` swapped its nine `.bot-list` `querySelector` calls for
`getByRole("navigation", { name: "Chat list" })` rather than renaming the class in the selector — one edit
removes the markup coupling permanently.

## Verification

Per `AGENTS.md`, no repo-wide check was run. All nine typechecks green (contracts, node, renderer,
storybook, brand, team-client, api, logging, sites); `bun run check:ui` green; `biome check` clean on every
touched path. Green test files: `v1/v2/v3.test.ts`, `ipc.test.ts`, `ipc-channel-coverage.test.ts`,
`openbot-database-schema-parity.test.ts`, `openbot-database.test.ts`, `agent-store.test.ts`,
`mailbox-store.test.ts`, `duplication-gate.test.ts`, `agent-service.{queue,providers,restart}.test.ts`,
`routine-scheduler.test.ts`, `input-parsers.test.ts`, `redaction.test.ts`, `team-api-server.test.ts`,
`remote-server-event-stream.test.ts`, `App.test.tsx`, `Sidebar.test.tsx`, `SettingsModal.test.tsx`,
`SkillsMarketplaceModal.test.tsx`.

**No UI change to show** — this PR renames the words behind the interface; the two copy edits that change
what a user reads are `OnboardingFlow`'s "A team of Bots that works with you." → "A team that works with
you." and "Hand off any task to your team of Bots" → "Hand off any task to your team", because "a team of
agents" reads like infrastructure and the `Meet OpenBot` heading already carries the frame.

### Three pre-existing Biome warnings left in place

None are introduced here; each is the correct code and `biome-ignore` is unavailable to the harness:

- `packages/logging/src/redaction.test.ts:148` — `typeof` narrowing an `unknown` at a trust boundary,
  which is the case the warning exists to permit.
- `src/main/team-api-server.test.ts:494` and `:495` — module mocks the test needs to stand in for the host.
- `src/main/team-webrtc-host-gateway.test.ts:122` — same.

## Not verified

The end-to-end pass on a migrated profile (open a pre-migration agent's history, follow a file link in an
old message, complete a hosted-site publication) has **not** been run — it needs a live dev instance and a
real profile. Worth doing before merge, along with timing v13 against a copy of a real profile.

## Review round 2

Merged `origin/main` (the branch was `CONFLICTING`, which is why no `pull_request` workflow ran for the
first nine hours — GitHub cannot build a merge ref it cannot compute). The sidebar decomposition that
landed on main was taken as-is and the codemod re-run over it.

Six findings from the automated review, all verified against the code and all valid:

1. **Released Team API query parameters had been renamed.** The versioned adapters translate JSON bodies
   only, never a URL, so `/v1/workspace-files?agentId=` and the conversation-search parameter would have
   404'd every shipped client and been 404'd by every shipped host. Both readers in `team-api-server.ts`
   and both writers in `remote-server-manager.ts` are back to `botId`, commented with why. The existing
   `team-api-server.test.ts` case is what caught it once the handler was restored.
2. **v2 and v3 decoders returned wire-shaped objects.** `JSON.parse(encode(...))` was a no-op before the
   key shim and wrong after it: `POST /v1/browser/open` reached a handler reading `ownerAgentId` while
   carrying `ownerBotId`, losing the tab's owner with nothing thrown. Both adapters now decode what they
   encode, asserted in `v2.test.ts` and `v3.test.ts`.
3. **`avatarSeed` was rewritten by migration v13.** The seed is the input to the face-drawing function, not
   an identifier — the plan called this out and the migration did it anyway. Restored in the roster
   projection and in the event payloads a replay rebuilds it from.
4. **Uploaded avatar directories were stranded.** v13 rewrites the id and `avatarUrl` derives the directory
   from it, so every uploaded avatar silently fell back to a drawn face. The directory now follows the id,
   and `deleteAgent` cleans up the pre-rename one.
5. **`deleteAgent` recursively removed a path read from the database.** It now derives the pre-rename path
   instead; a recursive delete never follows a string this code did not build.
6. **A failed directory move aborted startup.** The database has already migrated by that point, so a
   directory name must never lock the user out. All of the reconciliation is non-fatal now.

**Watched it fail**, twice more: dropping the seed restore makes the v13 test report the id where the seed
belongs, and skipping the avatar move makes the store test report `ENOENT` on the new path. Reverting the
two adapter fixes fails `v2.test.ts`/`v3.test.ts` on exactly `ownerAgentId` vs `ownerBotId`.

## Review round 3

Seven more findings, all verified and fixed. Two of them were introduced by the round-2 fixes.

1. **Migration v13 renamed ids the application never minted** (P1, worst outcome in the set). The
   candidate query was `LIKE 'bot-%'`, but `bot-` is also an ordinary word: `isValidAgentId` is
   `/^[a-z0-9][a-z0-9-]{0,63}$/`, and an imported `bots.json` carries whatever ids that install used. A
   `bot-research` alongside an existing `agent-research` collides on the primary key, the migration
   throws, `runMigration` rolls back — and the next launch repeats it, so the app never opens again.
   Now gated on the UUID suffix.
2. **`legacyAgentId` had the mirror defect on the delete path.** It translated any `agent-*` id, so
   deleting `agent-research` recursively removed `~/OpenBot/Bots/bot-research`, which can be a different
   agent's workspace. Same gate, one shared predicate.
3. **The v2 request encoder and decoder shared one implementation**, so the round-2 fix to the inbound
   direction moved the defect to the outbound one: `/v1/browser/open` went on the wire spelled
   `ownerAgentId`. Split into two functions.
4. **The v3 WebRTC encoders called the v3 decoders.** Harmless while the two vocabularies were the same
   words; after the rename an outbound duplicate response left as `{ agent: … }`, which the receiving
   peer's frozen codec rejects outright. Now they call the encoders.
5. **The avatar-seed restore was unconditional.** It rewrote a seed that already read `agent-<uuid>`
   before the migration. The seed is now masked before the substitution and unmasked after, so only a
   seed this migration actually changed is changed back.
6. **Startup directory probing was fatal.** `directoryExists` rethrows `EACCES`/`EPERM` and both probes
   sat outside the `try`, so a permission error aborted `initialize()` *after* the database had migrated.
   A probe that cannot answer now means "leave this alone".
7. **Windows workspace roots were half-migrated** (P2 — `dist:win` is a shipped target). Root detection
   matched only `/`, so the leaf id was rewritten under a `Bots` parent that `rebaseLegacyWorkspacePath`
   does not recognize. Both separators are handled, in the raw path and its JSON-escaped form.

**Watched it fail** on all three new assertions. Removing the UUID gate makes the migration throw and the
database refuse to open — the exact failure the finding predicted. Reverting the v2 encoder fails on
`ownerBotId` vs `ownerAgentId`; reverting the WebRTC one fails inside the frozen v3 codec on
`!Object.hasOwn(value, "bot")`.

The new migration test also caught a real defect in fix 5 on its first run: the sentinel was a raw NUL
byte, which fails the `json_valid` CHECK on the payload columns mid-migration. It is the six-character
JSON escape now.

## Two consequences the reviewer flagged and I did not change

Both come from the same missing thing — nothing on disk records *why* an id is spelled `bot-<uuid>` — and
both are trade-offs rather than oversights, so they are here for the human reviewer to overrule if they
want them handled differently.

1. **v13 rewrites every `bot-<uuid>`, including one a caller chose.** `getOrCreate` accepts a caller-
   supplied id, so a user-defined `bot-<uuid>` is indistinguishable from a generated one and is renamed
   too, invalidating any external reference to it. Migrating everything is decision 2 above; with no
   provenance the choice is all-or-none. Preserving ambiguous ids instead would leave the rename half-done
   in exactly the way this PR exists to end.
2. **A browser tab whose owner was deleted can be reowned by the surviving twin.** When `bot-U` and
   `agent-U` coexisted and `bot-U` is deleted, its persisted tab matches `agent-U` through `legacyAgentId`.
   Requiring the thread to corroborate does not help: `stableThreadId` makes the deleted agent's thread id
   and `legacyThreadId(agent-U)` the same string by construction, so the same transfer passes while tabs
   with no owner thread break. Closing it needs persisted v13 rename provenance, which is a schema change
   beyond this PR.

## CI note

`team-api-server.test.ts > shares sidebar layout mutations…` timed out once on this branch. It times out
on `main` too — 3 and 9 September, on unrelated commits — and passes locally 6/6 and on re-run here. It is
a pre-existing flake in a coalescer test, not a regression from this PR, and fixing it belongs in its own
change.

---

Model: Opus 5. Harness: Claude Code.


NorbertBodziony:
All three P1 findings from the last round were real and are fixed in 9ac7c31f + a443b9d2.

1. **Legacy ID derivation can delete another agent's files** — `AgentStore` now derives a legacy directory name through `#unclaimedLegacyId`, which returns `null` when another stored agent is living under that name. Used by `deleteAgent` (both recursive removals) and by both startup reconciliation moves.
2. **Migration v13 rewrites user-defined bot-prefixed IDs** — `readAgentIdRenames` now reads the full set of existing agent ids and drops any rename whose target is already taken, so a `bot-<uuid>` sitting beside an `agent-<uuid>` that shares its UUID is left alone.
3. **Blanket text replacement can violate memory uniqueness** — the substitution is now one statement per table rather than one per column (so a row is rewritten whole, never half) and uses `UPDATE OR REPLACE`. `OR IGNORE` was tried first and is worse: it leaves the skipped row's `agent_id` spelling an agent that no longer exists, so the memory survives attached to nobody. Collapsing the duplicate is what `UNIQUE(agent_id, normalized_text)` means. It cannot reach an identifier, because finding 2's preflight leaves every rewritten identifier one-to-one.

**Watch it fail**, per `AGENTS.md` rule 2 — each break produced the predicted failure:

- Removed `if (taken.has(newId)) continue;` → *"leaves an agent id the application did not mint alone"* red with 3 agents instead of 4, the `bot-<uuid>` row gone.
- Replaced `UPDATE OR REPLACE` with plain `UPDATE` → *"rewrites agent ids without losing a thread…"* red with `OpenBot database migration to version 13 failed` — the lockout the finding describes.
- Made `#unclaimedLegacyId` return the derived name unconditionally → *"deletes agents persistently…"* red on `ENOENT … /home/OpenBot/Bots/bot-<uuid>/notes.md`, the sibling's workspace destroyed.

Green: `openbot-database.test.ts` 29, `agent-store.test.ts` 26, `openbot-database-schema-parity.test.ts` 3, `mailbox-store.test.ts` 25.

/norbiai review

NorbertBodziony:
All three findings addressed in 0d42c2d3.

**1. Blanket text replacement can violate memory uniqueness** — taken as recommended. `projection_agent_memories.text` and `normalized_text` are now excluded from the substitution, so no memory is deleted, skipped, or left with its two text columns disagreeing. You are right that `OR REPLACE` was discarding a record with its own origin, source turn and timestamps. The row's identifiers are still rewritten around the sentence, so the memory stays attached to its agent; the only cost is that a memory quoting the pre-rename id keeps that spelling — which is what the user typed.

`UPDATE OR REPLACE` stays for the rest of the scan. With memory text excluded, the remaining unique TEXT columns are derived identifiers and `file_deletion_outbox.path`, where the rename is one-to-one (finding 2's preflight from the previous round guarantees it) and a genuine duplicate key is a duplicate row.

**2. New migrations lack failure and retry coverage** — added `rolls back a failed agent id rewrite and succeeds on retry` for v13: a `CHECK` blocker makes the substitution throw after it has already written rows, then the test asserts the marker is absent, the agent id and its thread messages are untouched, and that dropping the blocker lets the retry complete to v14 with an empty `foreign_key_check`.

Scoped to v13 deliberately. v12 is the same `runMigration` table rebuild the existing `rolls back a failed baseline migration and succeeds on retry` already covers for v8 — it adds no new mechanism, and a second copy would assert the same code twice. v13 does: it is the only migration that writes across every table with foreign keys disabled. I also tried asserting foreign-key enforcement was restored after the failure and removed it: the pragma lives on the connection, and a failed `initialize()` discards it, so nothing observable could distinguish a restored switch from a missing one. That assertion could only ever have passed.

**3. Deleting an unmigrated legacy agent leaves its workspace** — `deleteAgent` now always clears `join(#legacyAgentsRoot, id)`, the agent's own id under the old root, which no sibling can own. The ownership-guarded pair (the *derived* legacy id) is still conditional.

**Watch it fail**, per `AGENTS.md` rule 2 — each break produced the predicted failure:

- Removed the memory-text exclusion → `rewrites agent ids…` red, one memory of the two gone.
- Turned the `ROLLBACK` in `runMigration` into a `COMMIT` → `rolls back a failed agent id rewrite…` red with `agent-6d3e8b17-…` on disk where the un-migrated `bot-6d3e8b17-…` should be.
- Dropped the agent's own id from the delete's legacy paths → `deletes agents persistently…` red: the workspace still readable after a successful delete.

Green: `openbot-database.test.ts` 30, `agent-store.test.ts` 26, `openbot-database-schema-parity.test.ts` 3.

/norbiai review

NorbertBodziony:
Round 6 follow-up, plus one gap the new tests found.

### Fixed

**2 — `bots.json` array key.** `LEGACY_AGENTS_STATE_KEY` had followed the rename to `"agents"`, so the legacy import read a key the file never has and discarded every agent in it. Back to `"bots"` with the other two frozen literals. *Watched it fail:* reverting to `"agents"` fails 3 tests with `Agent state is corrupt or from a newer OpenBot version; refusing to overwrite it.`

**3 — browser tabs.** A tab a released build wrote spells the owner `ownerBotId` and can still carry a `bot-<uuid>` value. Parsing now lives in `browser-state.ts` as `storedBrowserTab(value): StoredBrowserTab | null`, accepting the old key and moving the id value with it. *Watched it fail:* dropping the fallback turns `"ownerAgentId": "agent-6d3e8b17-…"` into `null`, i.e. every tab open at the moment of the upgrade comes back unowned.

**4 — duplication marker and receipt.** Both are rows a released build wrote: the receipt spells them `sourceBotId` and `bot`, and the marker file is named after the duplicate's *pre-rename* id. Recovery now resolves the agent through either spelling and — this is the part the test caught — addresses it by the id it has **now**. It had been using the marker's file name, which filters nothing out of the roster and hard-deletes an id no row carries. Cleanup covers both roots and both spellings. Two `it`s, because the failure modes are disjoint: an unreadable marker on a *committed* duplication deletes an agent the user kept, while a stale lookup on an *uncommitted* one leaves a phantom in the sidebar. *Watched both fail*, each for its own reason.

**5 — v12 rollback.** Added the retry leg to `openbot-database.test.ts`. *Watched it fail:* turning `runMigration`'s `ROLLBACK` into `COMMIT` makes the retry die with `table projection_reactions_v12 already exists`.

**New — user-named workspaces never moved.** `#unclaimedLegacyId` returned `null` when the id's spelling was unchanged, which is true for `chief` and any id the user picked. But the *root* moved as well, so those workspaces stayed under `OpenBot/Bots` forever while `PRIVACY.md` says they are under `OpenBot/Agents`. Nobody else can hold the name, because ids are unique. *Watched it fail:* `OpenBot/Agents/chief/notes.md` goes missing.

### Not taking — 1, gating v13 on creation provenance

The request is that v13 stop rewriting a `bot-<uuid>` the user defined themselves, and decide from persisted creation provenance instead. Two reasons this stays as written.

There is no such provenance. Nothing in the schema records who minted an id; `isGeneratedAgentId` reads syntax, and adding a provenance column would have to guess retroactively for every existing row — the same judgement the reviewer is asking me not to make, one migration earlier and with less information.

And rewriting it is the change that was asked for. The decision on scope was explicitly *rewrite everything*, with the cost accepted in the plan: paired phones hold the old id, the wire protocol is frozen so no "id changed" event can be sent, and the agreed mitigation is a `CHANGELOG.md` entry telling users to unpair and pair again. Restricting the rewrite by provenance reintroduces exactly the half-rename this PR exists to remove — some ids say `agent`, some say `bot`, and every future reader has to know why.

The genuinely destructive case is a collision, where a `bot-<uuid>` and an `agent-<uuid>` sharing a UUID both exist. The `taken` preflight already declines there and leaves the `bot-` id alone, and `#unclaimedLegacyId` refuses to touch a directory another agent is living in for the same reason.

NorbertBodziony:
Findings 2, 3 and 4 were all correct, and each one was a real hole my tests had walked around. Fixed in `9a14d2aa` and `3ed81c72`.

**2 — pre-rename duplication state.** Right, and worse than "unreadable": `#recoverPendingDuplications` runs inside `initialize()`, so the throw takes the whole startup with it. My tests used `"chief"` as the source precisely because that id survives v13 unchanged, which is the one case that cannot reproduce it. The marker's source is now resolved through the roster before it is compared (`#agentByEitherSpelling`), and the test source is a generated id with the marker holding the pre-rename spelling of both the file name and the source. *Watched it fail:* comparing raw gives `Error: The agent duplication receipt is invalid.` out of `initialize()`.

**3 — browser tab ownership.** Both halves correct. Rewriting `bot-` to `agent-` by shape is a guess the file cannot make — v13 declines on a collision, and a `bots.json` import runs after the migrations with its own ids — and I had left `ownerThreadId` untranslated entirely, so `#canUseToolTab` compared a thread id that no longer exists and refused every tool call against the tab the agent itself opened. `storedBrowserTab` now translates the released `ownerBotId` *key* only and leaves every id value exactly as found; `reownStoredBrowserTab(tab, agents)` does the mapping against the roster that `restore()` is handed from `src/main/index.ts`. An owner that resolves to nobody keeps the id it was found with, because inventing one hands the tab to a stranger. *Watched it fail* twice: without the mapping the owner stays `bot-<uuid>`; without the current-spelling guard, a colliding `bot-`/`agent-` pair gives the tab to the wrong agent.

**4 — custom-id workspaces.** Correct, and my previous test was staged from post-rename data, so it proved less than it looked. The destination is now derived as `join(#agentsRoot, agent.id)` rather than read out of `workspacePath`, and the moved path is persisted. The test now rewrites `agent_json.workspacePath` to `OpenBot/Bots/chief` first — the released state — and asserts both the file and the stored path afterwards. *Watched it fail:* reading the destination from the stored path leaves `OpenBot/Agents/chief/notes.md` missing.

**1 — v13 rewriting bot-prefixed ids.** I'm not changing this and I'm not going to keep arguing it either, because it isn't a defect I can resolve from the code: the scope was set explicitly by the developer as *rewrite every id*, with the external-identity cost accepted in the plan and written into `CHANGELOG.md`. Your point that re-pairing only covers phones is fair, and there is genuinely no persisted provenance to narrow the rewrite with — so the choice is "rewrite all of them" or "rewrite none", and which of those to ship is the developer's call, not mine. Flagging it to them rather than deciding it here.

/norbiai review

NorbertBodziony:
Round 9 — three of the four findings fixed, one escalated.

**2 — sidebar layout (`c48da86b`).** Real, and the worst of the set: `reconcileAgents` filtered `openbot-sidebar-layout-v1.json` straight against the post-v13 roster, so every renamed agent read as deleted and lost its section and its place in the order, with nothing to undo it. Legacy ids are now remapped through the roster first, and an id still in the roster answers for itself — v13 declines to rename onto a taken id, so `bot-<uuid>` and `agent-<uuid>` can coexist and the entry belongs to the one the user filed.

**3 — browser tab reowning (`664b500f`).** Real, and I had it half-fixed. A generated agent's thread id is `openbot-thread-<uuid>` with no agent id inside it, so v13 never rewrote it; matching on *either* field and returning early declared the tab correct while its owner id was still `bot-<uuid>`, and `#canUseToolTab` checks both. The owner is now resolved once — exact id, exact thread, then the historical spellings — and both fields follow it.

**4 — `rebaseLegacyWorkspacePath` (`4f212bfe`).** Real. The workspace move gives up on `EXDEV` or a permission error and leaves the directory where it was, but v13 has already rewritten the paths in that agent's stored messages to where it *would* have gone. Only the new→old direction existed, so for exactly the users whose move failed, every file link in their history stayed broken. Both directions go through the unchanged `isWithin` check; nothing about containment moved.

**1 — user-defined `bot-` ids: still declining.** Unchanged from the last round. v13 rewriting a `bot-`-prefixed id the user chose is the scope the developer picked ("przepisać wszystko"), not a defect I can quietly narrow. It needs their call, not mine.

**Watch it fail**, per the repo rule, each confirmed to go red for the intended reason:
- sidebar: dropping the remap → `expected undefined to be '7b44d5f9-…'`; making the legacy lookup win over the roster → `expected { …(1) } to have property "bot-6d3e8b17-…"`. The first draft of the collision assertion passed vacuously (the layout already held the new id by then) — rebuilt on a fresh store so it is load-bearing.
- browser tab: stopping at the first match → the generated-thread case goes red; asking the historical spelling before the current one → the collision case goes red.
- rebase: dropping the reverse branch → red; dropping `isWithin` → `expected '/Users/dev/OpenBot/Agents/secret.env' to be null`.

`src/backend/workspace-paths.test.ts` is a new file — that helper had no coverage at all and it is the lowest stable boundary for a path resolver that sits in front of a containment check. Everything else extended an existing `it`.

Suites: `sidebar-layout-store` 10, `browser-state` 4, `workspace-paths` 1, `agent-store` 28, `agent-service.providers` 11. `biome check` and `typecheck:node` clean on the touched paths.

/norbiai review

NorbertBodziony:
Round 10 — five of the six fixed in `9d8ef6d5`, one still declined.

**1 — substring merge in v13.** Real, and the worst thing found so far. The rename travels as a text substitution, so it reaches a generated id wherever it appears — including inside a longer id a caller chose. `getOrCreate` takes any string, so `bot-<uuid>-copy` is a legal id, and rewriting the token inside it renamed a second agent nobody asked about; where `agent-<uuid>-copy` already existed, the row-level `OR REPLACE` settled that collision by deleting an agent. An id that is a piece of another id is now skipped, alongside the existing target-taken guard. I did **not** take the suggested remedy of removing the whole-database substitution — rewriting quoted ids and workspace paths inside message text is the scope the developer chose, so I closed the hole instead of narrowing the feature.

**3 — dynamic map keys.** Real. `responses` and `references` are keyed by data, not vocabulary; a question id of `agentId` crossed the wire as `botId` while `questions[].id` stayed `agentId`, and no client could pair the answer with its question. Those two keys are now passed through while their values are still translated. The same class hit a route the finding did not name: `GET /v1/agents/conversation-reads` returns a body that is *entirely* a map keyed by agent id, so it joins sidebar layout as an untranslated path.

**6 — sidebar pins.** Real, and the same shape as the layout bug from round 9. Pins are reconciled against the roster the server sends, current spelling first. Because the renderer needed it, `legacyAgentId` and `isGeneratedAgentId` moved to `@openbot/contracts/validation` — one definition rather than two that can drift.

**4 — export versions.** Agreed. Backup manifest 3 → 4, diagnostics 2 → 3.

**5 — analytics.** Agreed. Generation 4 → 5 on all three surfaces in lockstep, and `ANALYTICS.md` now states that generation 5 reports `origin: agent` where 4 reported `bot` and that the two cannot be combined.

**2 — user-defined `bot-` ids: still declining**, for the third time and for the same reason. `bot-<uuid>` from `getOrCreate` is indistinguishable from an app-minted one without provenance nobody persisted, and adding a provenance column to invent that distinction is a product decision. The developer chose "rewrite everything". What I *can* do is stop that scope from destroying data, which is findings 1 and 2's actual overlap and is now closed.

**Watch it fail**, each red for the intended reason:
- v13 guard removed → the collision test drops from 7 agents to 6, showing `bot-…-copy` merged into `agent-…-copy`.
- `DYNAMIC_MAP_KEYS` emptied → `expected [ 'botId' ] to deeply equal [ 'agentId' ]`.
- pin remap removed, and separately the roster-first check removed → both go red on the repin test.

Suites: `openbot-database` 31, `v1`/`v2`/`v3` 11/14/7, `sidebar-pins` 6, `analytics` (main 20, renderer 17), `App` 25, `App.servers` 26, `agent-store` 28, `browser-state` 4, `sidebar-layout-store` 10, `workspace-paths` 1, `validation` 10, `check:api` 158. `typecheck:node`, `:renderer`, `:contracts`, `:storybook` clean.

/norbiai review

NorbertBodziony:
Round 11 — one fixed, two declined with reasons. `735bd880`.

### 3. Wire translation corrupts dynamic response-map keys — **fixed**

Correct, and I missed the request half. `RespondToPromptInput.answers` is `Record<questionId, string[]>`
(`isV1PromptAnswers`, `v1.ts:2793`) and is reached through `encodeTeamProtocolV1CurrentHttpRequest` /
`decodeTeamProtocolV1CurrentHttpRequest`, which the round-10 test never touched — it only covered the
response. `"answers"` is now in `DYNAMIC_MAP_KEYS`, and the existing dynamic-map test carries the
`POST /v1/prompts/respond` round trip. The resolution-side `answers` is an array of strings, so it never
reaches `walkObject` and the set cannot affect it.

*Watch it fail:* removed `"answers"` from the set →
`AssertionError: expected [ 'botId' ] to deeply equal [ 'agentId' ]` on the request body.

### 1. Substring migration can merge unrelated agents — **declining the proposed remedy**

The remedy — "exclude filesystem path/name columns from ID substitution and apply only the separate
workspace-root rewrite to paths" — breaks two paths that are correct today, because the id substitution is
what supplies the *leaf* of a path whose root the separate rewrite supplies:

- `~/OpenBot/Bots/bot-<uuid>` → `~/OpenBot/Agents/agent-<uuid>`. The root rewrite gives `Bots` → `Agents`;
  only the id substitution gives `bot-<uuid>` → `agent-<uuid>`. Excluded, every stored workspace path
  becomes `Agents/bot-<uuid>` — a directory that does not exist on disk after `#reconcileWorkspaceDirectories`
  moved it.
- Avatars live in a directory named by the agent id (`userData/avatars/agents/<id>/<version>.png`,
  `agent-store.ts:475`) and that directory is moved by `#reconcileAvatarDirectory` (`:633`). The root
  rewrite does not reach it at all; only the id substitution does.

So excluding path columns trades a coincidence for a certainty. On the scenario itself: it needs a file
whose basename is exactly `bot-<uuid>.txt` where `<uuid>` is that same user's own generated agent id. The
app never writes such a file — attachments are named by attachment id, avatars by version — so it has to be
a user file the user happened to name after their own agent's UUID.

On the `OR REPLACE` half you are right that the comment was wrong: it named `projection_files.path`, which
does not exist in this schema. The path-unique column is `file_deletion_outbox.path`, and collapsing two
queued deletions of the same rewritten path is a dedup, not a loss. Comment corrected in this commit.
`projection_attachments.path` is not unique, so nothing there can collapse.

### 2. Migration v13 rewrites user-defined bot-prefixed IDs — **declining, fourth time**

Unchanged from rounds 9 and 10: rewriting every `bot-<uuid>` id is decision 5 of five the developer made
explicitly after being shown this exact cost. `readAgentIdRenames` already declines a rename onto a taken
id and (since round 10) an id that is a substring of another, so no agent is merged or lost — a user who
typed `bot-<uuid>` by hand gets it respelled. That is the intended outcome of the change, not a defect in
it. Narrowing the scope is the developer's call, not the reviewer's or mine.

/norbiai review

NorbertBodziony:
Round 12 — both P2s fixed, the P1 declined again. `e8d027a5`.

### 2. Transfer manifest v1 now has two incompatible schemas — **fixed**

Correct. `.openbot-transfer.json` renamed `recipientBotIds` → `recipientAgentIds` and `ownerBotId` →
`ownerAgentId` while still calling itself version 1, so one number described two shapes. Bumped to 2 at all
three write sites (`mailbox-store.ts:1029`, `:1134`, `:1291`) with the reason on the field.

One correction to the finding's framing: there is no consumer to lose metadata. `TRANSFER_MANIFEST_FILE`
has exactly two references in the repo — the constant and the one `writeFile` — so nothing here selects a
parser by version. The sidecar is written for the user and the model looking at the transfer directory.
That is why I bumped the number rather than reverting the field names: the number exists for whoever reads
it later, and it should not lie.

*Watch it fail:* set the message-transfer write site back to 1 →
`AssertionError: expected { version: 1, …(7) } to match object { version: 2, …(5) }` in
`mailbox-store.test.ts:218`.

### 3. Managed model instructions still call agents bots — **fixed**

Right, and it is squarely in scope — the model-facing prompt surface is one of the four surfaces this PR
covers, and `syncAll` writes the file into every agent workspace unconditionally, so the edit propagates on
the next sync. `resources/managed-skills/openbot-site-hosting/SKILL.md:19` now says "this agent's workspace".

### 1. Migration v13 rewrites user-defined bot-prefixed IDs — **declining, fifth time**

The finding is accurate about the behaviour and I am not disputing it: v13 renames a `bot-<uuid>` id whether
the app minted it or a caller supplied it, and there is no provenance record to tell them apart. That is the
change, not a bug in it — decision 5 of five the developer made explicitly after being shown this cost, and
the reason the migration exists at all. The guards that matter are in place: `readAgentIdRenames` declines a
rename onto a taken id and declines an id contained in another, so nothing is merged, deleted, or orphaned.
A hand-typed id gets respelled and keeps its agent, its workspace, its thread and its history.

"Preserve ambiguous IDs" is a scope decision, and narrowing the developer's stated scope is theirs to make,
not mine. Flagging it again does not move that.

/norbiai review

NorbertBodziony:
Round 13 — both P2 disk findings fixed, the P1 and the performance finding declined. `2f5e42b6`.

### 2. Existing target directory can replace the real workspace — **fixed**

You are right, and the reasoning generalizes: the move is a single atomic `rename`, so it never leaves both
directories behind. Two at once therefore is not the interrupted-move story that branch was written for —
it is a destination that was never this agent's. The branch now requires the old directory to be confirmed
absent (`legacy === false`, not merely "not present", so a failed probe still declines), and an ambiguous
pair leaves the record naming the workspace the agent has actually been reading.

*Watch it fail:* dropped `legacy === false` →
`Expected: ".../OpenBot/Bots/chief" Received: ".../OpenBot/Agents/chief"` at `agent-store.test.ts:139`.

### 4. Existing avatar target strands the uploaded avatar — **fixed**

Same pair, and your remedy is the right one. `#reconcileAvatarDirectory` now falls through to
`#adoptLegacyAvatarFile`, which moves the single file `avatarUrl` names — nothing else in that directory is
ever read — and only when no file already answers to that name in the target, since one that does is the
newer upload. Both cases are covered by extensions to the existing reconciliation test rather than new
files.

*Watch it fail:* replaced the fall-through with `return` →
`promise rejected "Error: ENOENT ... /avatars/agents/agent-<uuid>/<version>.png" instead of resolving`.

### 3. Migration performs hundreds of full database scans — **declining**

The arithmetic is right and the conclusion is overstated. Passes are `2N + R + N` where `N` is the number of
agents actually renamed and `R` the distinct workspace roots — not the number of agents. `readAgentIdRenames`
already drops every id that is not UUID-shaped, every one whose target is taken, and every one contained in
another, so `N` is the generated agents on a pre-rename profile: single digits for a real user, and zero on
every launch after the first, where the migration returns before preparing a statement.

Restructuring this into one traversal means folding all renames into a single nested `replace()` chain per
table. That trades passes for parameter count — columns × pairs × 3 bound variables in one statement — in the
one piece of code in this PR that cannot be rolled back by a later release and runs against a database with
no backup. I am not making that trade during review on an unmeasured hypothesis. What is bounded already:
`runMigration` wraps the whole thing in `BEGIN IMMEDIATE` and rolls back on any exception, so an interrupted
upgrade leaves the pre-migration database intact.

Timing v13 against a copy of a real profile is still on the pre-merge checklist and is the developer's call
to make with a real profile in hand — it is one of the two things in this PR I have not been able to verify.

### 1. Migration v13 rewrites user-defined bot-prefixed IDs — **declining, sixth time**

Unchanged. The behaviour is intended and stated, no agent is merged, deleted or orphaned, and narrowing the
developer's stated scope is not mine to do.

/norbiai review

NorbertBodziony:
Round 14 — four fixed, including the performance finding I declined last round. `9b2e9d7b`.

### 3. Migration performs hundreds of full database scans — **fixed**

You were right to push on this. My "single digits in practice" answer was an expectation, not a bound, and
`INPUT_LIMITS.agents` is 100, so the worst case is four hundred scans of an unbounded history at startup.

Sixteen replacements now share one pass, nested into a single expression per table, so the four phases cost
roughly `4 × ceil(N / 16)` passes rather than `4N`. The bound exists because SQLite has an expression
nesting limit — measured on this build at between 500 and 999 nested `replace` calls, so sixteen leaves the
margin an unfamiliar SQLite needs.

Equivalence is what makes this safe to do here, and it rests on properties the code already guarantees:
the phases stay ordered (seeds masked, roots, ids, seeds restored), and within a phase no pair can feed
another — `readAgentIdRenames` drops any id contained in another id, and a sentinel is unique to its rename.
Checked rather than asserted: the migration suite passes unchanged at `SUBSTITUTION_BATCH` of 1, 2 and 16.

*Watch it fail:* removed the mask phase → `expected { Object (seed) } to deeply equal { Object (seed) }` at
`openbot-database.test.ts:1186`, the assertion that an agent's face survives the rename.

### 2. Legacy imports bypass the provider-session refresh — **fixed**

Correct, and the reasoning is exact: `bots.json` is imported after the migrations run, so its session was
the one active pre-rename session on the machine that v14 never saw. It is now bound and immediately
deactivated — kept, not dropped, because the row is what a later handoff and the thread's history read.

*Watch it fail:* removed the deactivation → `expected { …(10) } to be null` on `activeProviderSession`.

### 4. Empty-state copy uses the wrong article — **fixed**, and it was not alone

`connected-screen.tsx:150` now reads "Add an agent to start working from your phone." A sweep for the same
mechanical artifact across the diff found eleven more — "a Agent" and "a agent" in comments in
`WorkspaceConversation.tsx`, `WorkspaceOverlays.tsx`, `conversation.tsx`, `drain-scheduler.ts`,
`duplication-gate.ts` and others, plus one test name. All fixed; `\ba (agent|Agent)\b` now matches nothing
in the repo.

### 5. README still directs users to the retired workspace root — **fixed**

Both lines now say `~/OpenBot/Agents/<agent-id>`, and the data-boundary entry names the old root as what a
pre-rename profile holds and where a workspace whose move could not run stays readable — the fallback
`#reconcileWorkspaceDirectory` actually implements.

### 1. Migration v13 rewrites user-defined bot-prefixed IDs — **declining, seventh time**

Unchanged, and unchanged for the same reason: intended behaviour, no agent merged or lost, and the scope is
the developer's.

/norbiai review

NorbertBodziony:
Round 15 — finding 2 fixed in `96c2a0d0`, findings 1 and 3 declined with reasons.

### 2. Legacy conversation messages are left in an unreadable shape — **fixed**

Correct, and the worst thing found in several rounds. I checked the reachability rather than take it on
faith: `syncMailboxMessages` writes the mailbox-derived exchange message into the snapshot and
`persistConversation` stores it, so `projection_thread_messages.message_json` on any pre-rename profile
holds `exchange.senderBotId`, `recipientBotIds`, `deliveries[].recipientBotId`, and reactions actored
`{kind:"bot", botId}`. `isAgentExchangeSummary` requires the agent-named keys and `isConversationReaction`
requires `kind === "agent"`, so those messages fail the guard outright — and the consequence is not a
missing field, it is `decodeConversationMessageJson` throwing and taking the whole page down. Every user
whose agents ever messaged each other, or who ever saw an agent react, would have opened that conversation
to an error.

`src/backend/database/legacy-conversation-message.ts` normalizes the released key names and the actor
discriminant on the way in. It is used at `decodeConversationMessageJson`, which is the single choke point
for persisted reads, and at the three places `thread-replay.ts` validates an event payload — replay reads
the same shapes out of `orchestration_events`. Only keys and the `kind` discriminant move; message text is
never touched, which is why the migration declined to do this with `replace()` in the first place. A
message already in the current spelling skips the walk on one property check.

The fixture in the v13 migration test now carries the released message shape, and the test asserts the page
read returns it with the current keys. *Watched it fail twice:* dropping the normalization, and separately
dropping only the `kind` mapping, each go red with `Error: Invalid conversation message.` from the page
read — the exact failure the finding describes.

Suites: `openbot-database` 31, `conversation-read-store` 3, `team-chat-store` 2, `agent-service.queue` 23,
`mailbox-store` 25, `agent-store` 28, `agent-service.providers` 11. `biome check` and all ten `typecheck:*`
clean.

### 3. Legacy-ID fallback can transfer a browser tab — **declining, because the proposed remedy does not close it**

The scenario is real as described, but the fix you name does not prevent it. Thread ids are
`openbot-thread-${agentId}` (`openbot-database.ts:311`), so the deleted `bot-U` owned
`openbot-thread-bot-U`, and `legacyThreadId(agent-U)` computes `agent-U.threadId.replace("agent-U",
"bot-U")` — which is `openbot-thread-bot-U`. In a colliding pair the two legacy spellings coincide *by
construction*. Requiring the stored thread to match the candidate's exact or legacy thread therefore
admits exactly the same transfer, while removing the fallback for a tab whose `ownerThreadId` is `null`,
which is the case that has no thread to corroborate with.

Nothing local distinguishes the two situations, and that is the point: a pre-rename tab and a
deleted-agent tab are byte-identical — an owner id no live agent holds, in the `bot-` spelling. The only
thing that separates them is which ids v13 actually renamed, and that is not recorded anywhere.
Persisting v13's rename map would close this properly and would also give the sidebar layout, the pins and
this fallback a fact to consult instead of a shape — but it is a new table and new plumbing at this point
in the change, so it is the developer's call, not one I should slip in.

Worth stating plainly what the exposure is: reaching it needs two agents whose ids share a UUID under
different prefixes, which cannot happen by generation (`agent-${randomUUID()}`) and only arises from a
caller-supplied id, plus the deletion of the `bot-` one.

### 1. Migration v13 rewrites user-defined bot-prefixed IDs — **declining, eighth time**

Unchanged. Intended scope, chosen explicitly, with the external-identity cost accepted and written into
`CHANGELOG.md`; there is no persisted provenance to narrow it with, so the choice is all or none and that
is the developer's.

/norbiai review

NorbertBodziony:
Round 16, on `b260a757`.

**4 (sidebar layout can persist duplicate agent ids) — fixed.** `reconcileAgents` now dedupes the
translated `agentOrder`, and builds `agentAssignments` so an entry already spelled the way the roster
spells it wins over one remapped from a legacy spelling — which also makes the result independent of the
order the two entries happen to sit in the file. Without that, whichever entry came last silently decided
the user's section.

**5 (pin reconciliation can persist duplicate pins) — fixed, one level lower than suggested.** Rather than
normalizing at the call site, `reownSidebarPinnedItems` now normalizes its own result, so no caller can
skip it. `reconcileActiveServerPins` also drops the server key when nothing survives, matching the other
write path.

**2 (duplicate recovery resolving a marker by legacy spelling) — declining, with the reachability
argument.** Two halves:

- *Accidentally:* not reachable. The marker file is named after the **duplicate being created**, whose id
  is a freshly minted `randomUUID()` — never a pre-existing agent's. So `bot-N.pending` resolving to
  `agent-N` finds either that same half-created duplicate or nothing. That is exactly what
  `agent-store.test.ts:451` covers, and it is the path the fallback exists for.
- *Deliberately:* the fallback grants nothing. The marker directory is already trusted input — anyone who
  can write `<exact-id>.pending` into it deletes that agent on the next launch through the exact-match
  branch, with no legacy spelling involved. And that same actor has filesystem access to the workspace
  anyway. Removing the fallback removes no capability.

The cost of adopting the suggested remedy verbatim is concrete: in `duplicateAgent` the marker is written
before the copy and removed only *after* the commit receipt, while `#persist("agent.duplicated")` writes
the row before that — so an uncommitted marker legitimately coexists with a persisted row. Declining to
resolve it by legacy spelling would leave the half-copied agent in the roster with a stranded workspace,
turning a covered recovery into a phantom.

**1 and 3** stand as answered in rounds 12 and 15: v13's scope is one of the five decisions the developer
made explicitly, and the browser-tab remedy is ineffective because `stableThreadId` makes the deleted
`bot-U`'s thread id and `legacyThreadId(agent-U)` the same string by construction.

**Watch it fail** (rule 2), both new assertions, each red for the intended reason:

- `sidebar-layout-store.ts` order dedupe removed → `expected [agent-U] to equal [agent-U, agent-U]`.
- assignment merge reverted to the plain `Object.fromEntries` map → the agent comes back filed under
  `Archive` instead of the `Research` section it was filed in under the current spelling.
- `reownSidebarPinnedItems` normalization removed → the pin list comes back with `agent-U` twice.

No new test files and no new `it` blocks — both existing v13 tests grew the assertion.

Surfaces touched: desktop renderer (`sidebar.tsx`, `sidebar-pins.ts`) and backend (`sidebar-layout-store.ts`).
Narrow suites green: `sidebar-layout-store` 10, `sidebar-pins` 6, `App` 25, `App.servers` 26. Biome and all
ten typechecks clean.

/norbiai review

NorbertBodziony:
Round 17, on `8afdc7e6`.

**2 (dynamic-map data mistaken for protocol vocabulary) — valid, fixed.** The keys were already left
alone, but each one was still handed down as the *context* for the value beneath it, so a question id
that happened to spell a discriminant decided what the answer under it meant: `answers: { kind:
["agent"] }` went onto the wire as `{ kind: ["bot"] }`, changing the user's answer, and a message id
spelling `layout` made an entire referenced message opaque. A data key now hands down the same empty
context the top-level call uses — which also keeps a nested object under a dynamic map on the normal
translation path, rather than treating its keys as data too.

**Watch it fail** (rule 2): with the data key carried on as context again, the extended dynamic-map test
in `v1.test.ts` goes red with `expected { kind: [ 'bot' ] } to deeply equal { kind: [ 'agent' ] }` — the
user's answer rewritten, which is the consequence the assertion names. No new test file and no new `it`
block; the existing dynamic-map test grew the case.

**1 (v13 rewrites `bot-<uuid>` ids) and 3 (browser-tab legacy fallback) — standing as answered.** Both now
rest on the same missing thing: persisted rename provenance. There is none, and adding it is a schema
change beyond this PR. Rewriting every `bot-<uuid>` is decision 2 of the five the developer made
explicitly for this branch — with `getOrCreate` accepting caller-provided ids, no rule separates a
generated id from a user-chosen one, so the choice is all-or-none and it is theirs, not mine to reverse.
Recorded in the PR body as a known consequence.

Surfaces touched: contracts (`current-agent-keys.ts`) — the desktop, mobile and Team API clients all read
through it.

Suites green: `v1` 11, `v2` 14, `v3` 7, `team-api-server` 9. Biome and all ten typechecks clean.

/norbiai review

PR #242, opened 2026-09-05
test(tooling): give ui-foundation-check the fixture its rules require

NorbertBodziony:
Follows #240, which widened this guard. This one makes the guard checkable.

## Why

Every GritQL rule here owns a fixture proving both halves — the lines it must reject, marked
`// flag`, and the correct neighbour it must leave alone. AGENTS.md says why: *"A pattern that
matches nothing is green and enforces nothing."*

`scripts/ui-foundation-check.ts` enforces **eleven checks and owned no fixture at all**. That is how
two of them went blind: the CSS scan named `styles.css` and `styles/` by hand and so missed
`preview/preview.css`, and the skip list caught `.test.tsx` but not `.test.ts`. Both were found in
#240 by reading the script, not by running it. A third would not have been found that way — and a
third turned up here, along with three regex branches that had never been able to match anything.

## What this does

- Extracts `checkUiFoundation(rendererRoot, labelRoot)`, returning `{ failures, manualCompositeCount }`
  instead of logging and exiting. The roots are arguments so a fixture tree can be walked; `labelRoot`
  is separate so the CLI keeps printing repository-relative paths.
- Puts the CLI behind `if (import.meta.main)`, as nine other scripts here already do. Verified with a
  throwaway probe: under vitest `import.meta.main` is `false`, so importing the module runs nothing.
- Adds **`tools/ui-foundation/fixtures/renderer/`** — a miniature renderer that breaks every check,
  beside the correct neighbour each must leave alone. The test asserts the report line for line:
  **20 findings and 5 budget lines, exactly those.**
- Adds **`tools/ui-foundation/fixtures/renderer-clean/`** — a second renderer that breaks nothing,
  asserted to report nothing. A check that fires at most once per file cannot have both halves in one
  tree: in the tree that breaks it, the violation accounts for the failure whether or not a widened
  pattern has also started rejecting the correct code beside it.
- Fixes a path-prefix bug and three dead regex branches the fixtures exposed (below).

Two smaller things the fixture forced:

- `Good.tsx` distinguishes `<Button>` from `<button>` and a token from a literal; `tokens.css` covers
  the two values the ratchet exempts by hand (`border-radius: 0`, `transition: none`) and the
  lookaround boundary case (`var(--openbot-red-500)` names a colour a hyphen joins to its
  neighbours); `Bad.test.ts`, `Bad.test.tsx` and `components/ui` hold violations that **must** stay
  invisible.
- The fixture lives in `tools/`, not `scripts/`, deliberately: `scripts/**/*.test.ts` is in both
  `vitest.config.ts` and `tsconfig.node.json`, so `Bad.test.ts` would have been collected as a real
  test and typechecked. `tools/` is in no tsconfig and no vitest project. The added
  `!tools/ui-foundation/fixtures` in `biome.json` mirrors the existing `!tools/biome/anti-slop` — a
  file exclusion, not a rule being switched off; verified that `biome check tools/ui-foundation`
  checks 0 files.

## Three defects the fixtures found

**1. Any sibling directory starting with `ui` was skipped entirely.** `components/ui-kit` and
`components/uiLegacy` both satisfy `path.startsWith(uiRoot)`, so they read as being inside the design
system and every check in the file skipped them in silence — the same failure as the two original
blind spots. No such directory exists in `src/renderer/src` today, so nothing was being missed in
practice; the guard was one directory name away from it. The comparison now appends a path separator.
`components/ui-kit/Sneaky.tsx` carries one violation for each of the **two** places that prefix is
compared — a native control for the per-file walk, a `role="dialog"` for the composite ratchet —
because fixing one call site and not the other leaves half the hole open.

**2. `background-color` and every `border-*-color` branch could never match.** A hyphenated property
has to be quoted — `border-color` is not a bare JS key — and the colour pattern required the colon
immediately after the property name. Dead from the day they were written, and green about it. Found
by writing the `border-color` fixture. `inlineFoundationValue` beside it already allowed the quote;
the colour pattern now does too. `src/renderer/src` has no such style, so `check:ui` stays at 0/0 and
no budget moves.

**3. Every `|`-separated alternative shared a file with a sibling, so only one of each pair was
observable.** Deleting `textarea` from the native-control alternation, `fill` from the colour
properties, `hsl(` from either value list, `transition` from the inline foundation values or the
`option` role from the composite pattern all left the whole suite green: these checks report once per
file, so the second finding was never visible. `components/branches/` now holds one file per
alternative. `CompositeRoles.tsx` carries six roles in a single file, because that count is per
occurrence rather than per file.

The line that directory's README draws: **a `|`-separated alternative is a word someone can delete,
so it gets a file; an optional suffix group is a refinement of an alternative already covered, so it
does not** — unless deleting it changes the report, which is why `(?:-duration)?` has a file and
`(?:-color)?` cannot have one (the bare `color` alternative absorbs it, which is also why the border
case is spelled `border-top`). Coverage stops where the consequence does: `rgba`/`hsla` are in,
the direction words `right`/`bottom`/`left` are out. Adding files that cannot fail dilutes the
tree the same way a snapshot does.

## Numbers that carry the negative half themselves

- **colours = 6**, one per value form the pattern names: a hex, a named colour, an `hsl()`, an
  `rgba()` and an `hsla()` in `legacy.css`, and an `rgb()` in `preview/preview.css`. The `rgb()` is in
  the file the old hand-written scope missed, so narrowing the scan back to a directory-name list
  reads 5. The named colour is matched by a *second* pattern summed into the same count, which none of
  the functional forms exercise, so losing it also reads 5. The token in `tokens.css` must stay
  uncounted, or this reads 7 and the word boundaries have gone.
- **composites = 9**: one per role the pattern names, plus the two extra places the walk has to reach
  (`features/inbox/InboxPane.tsx` outside `components/`, and `ui-kit/Sneaky.tsx`). `Bad.test.tsx` and
  `components/ui/Button.tsx` each hold a `role="dialog"` the ratchet must not see; either exclusion
  breaking raises this to 10, both to 11.

## Watch it fail

Every break below was applied to the final fixture tree, measured against one baseline
(**25 report lines, composite 9, colours 6, clean tree 0**), and reverted with `git checkout`. Each
lands in the test it was meant to, and none of them is *"expected 3 children, got 2"*.

| broken | result |
| --- | --- |
| CSS scan narrowed back to a directory-name list | colours **6 → 5** |
| named-colour pattern dropped from the sum | colours **6 → 5** |
| word boundaries (`(?<![-\w])`) removed | colours **6 → 7**, the token `--openbot-red-500` starts counting |
| skip list back to `.test.tsx` only | **25 → 27** lines, both from `Bad.test.ts` |
| design-system skip loses its path separator | **25 → 24** lines *and* composite **9 → 8** — the `ui-kit/Sneaky.tsx` line and its `role="dialog"`, one per call site |
| composite walk narrowed back to `components/` | composite **9 → 8**, the `features/inbox/InboxPane.tsx` menu |
| composite count stops excluding `components/ui` | composite **9 → 10** |
| composite count stops excluding test files | composite **9 → 10** |
| namespace pattern widened past the `Primitive` alias | **only the clean tree** goes red; the 25-line assertion still passes, which is the masking `renderer-clean` exists to cover |
| namespace finding's path hard-coded again | **only** the label test, on the `complex.tsx` line |
| `labelRoot` ignored | **only** the label test |

And once per alternation branch, each losing exactly the line its own fixture contributes:

| deleted branch | line lost |
| --- | --- |
| `button` | `Bad.tsx` **and** `ui-kit/Sneaky.tsx` (two lines) |
| `input` / `textarea` / `select` | `branches/Native{Input,Textarea,Select}.tsx` |
| `fill` / `stroke` | `ColourFillNamed.tsx` / `ColourStrokeHex.tsx` |
| `rgba?\(` | `ColourBackgroundRgb.tsx` **and** `ColourBackgroundRgba.tsx` |
| `hsla?\(` | `ColourBorderTopHsl.tsx` **and** `ColourTextHsla.tsx` |
| `#[\da-f]{3,8}` | `Bad.tsx` **and** `ColourStrokeHex.tsx` |
| `font-size` | `InlineFontSize.tsx` |
| `transition` | `InlineTransition.tsx` **and** `InlineTransitionDuration.tsx` |
| `(?:-duration)?` | `InlineTransitionDuration.tsx` alone — which is why it gets a file |
| role `menu` / `option` | composite **9 → 8** |

Honestly about the process: the first pass ran against an **uncommitted** refactor, so the
`git checkout` that reverted each break threw the refactor away after the first one, and the next two
measured the pre-refactor script. Caught immediately (`no tests`, plus a stack frame at an
indentation the refactor does not have), refactor restored, committed, and the whole set re-run
against the right file. Two later breaks applied through `python3 -c` were silent no-ops from nested
shell escaping, caught by `grep -c` reporting the pattern still present, and redone with heredocs that
assert the replacement happened. The tables above are a single final pass with that guard in place.

## Review findings

All findings from the automated review were acted on. Three were valid:

1. **The namespace finding ignored `labelRoot`.** It built its message from a hard-coded string while
   every other one uses `relative(labelRoot, file)`, so under the CLI it printed
   `components/ui/complex.tsx` where its neighbours print `src/renderer/src/...` — the one finding you
   could not paste into an editor, and a contradiction of the labeling contract the exported
   function's own doc comment states.
2. **The correct adapter was masked by the invalid one beside it.** Fixed with `renderer-clean`, which
   covers the negative half of every once-per-file check rather than just this one.
3. **The fixture proved less than the test claimed** — the alternation gap above.

One was not: the `button` branch was reported twice as unguarded. It was not — `ui-kit/Sneaky.tsx`
holds no other native control, so deleting `button|` turns the suite red through that file, measured
both times. What *was* true is that `Bad.tsx` still had a nested `<input />` from before
`branches/NativeInput.tsx` existed, so its native-control line no longer said which branch it was
proving. Removed in `be4de02a`; that is fixture attribution, not coverage.

The direction words `right`/`bottom`/`left` were declined on the consequence grounds stated above,
with the reason recorded in the branches README — happy to add them if you prefer mechanical
completeness over a tree where every file can fail.

## Merged `main` twice

**#241** landed the `features/<domain>/` refactor and changed this same script: it widened the
composite role count from `components/` to the whole renderer, because a component that moves next to
its domain would otherwise leave the budget at zero by going blind. That is the same class of fix as
this branch and it wins the conflict outright — the merged file keeps main's renderer-wide walk, and
this branch's changes go back on top. That widening had no fixture of its own; narrowing it back would
have stayed green. `features/inbox/InboxPane.tsx` now holds a `role="menu"` outside `components/`, and
reverting the walk turns the suite red.

**#244** translated this script's messages and `docs/ui-foundation.md` independently, with different
wording for the same meaning. The merge takes main's wording and moves the expectations to it —
`untokenized` and `the count may only go down` included, rather than the British spelling this branch
had. `docs/ui-foundation.md` was the one deferral left here and is no longer outstanding.

The `AGENTS.md` conflict was a false one both times: paragraphs added at the same place, all kept.
That paragraph and the fixture README now also say **which half goes where** — both still described
the one-tree model from before `renderer-clean` existed ("adding both halves to that tree"), which
would have sent the next agent to put a negative half exactly where it is invisible.

## Surfaces

Repository tooling and documentation. The renderer is **read**, not changed. No mobile, web, IPC
contract, migration or schema change.

## Checked

`bun run test:desktop -- scripts/ui-foundation-check.test.ts` → 4/4; `bunx biome check` clean on the
changed paths; `bun run typecheck` (all projects) clean via the pre-commit hook; `check:ui` green;
`bun scripts/ui-foundation-check.ts` still works as a CLI. All re-run after each merge.

The new test file is justified by AGENTS.md rule 1 — the boundary "this guard actually catches what it
claims" did not exist. No `biome-ignore`, no `@ts-expect-error`, no `any`/`unknown`, no rule disabled
in `overrides`.

Out of scope and deliberately left: the ten `noUndeclaredDependencies` findings outside `src/**`
(four manifests and the lockfile).

Model: Claude Opus 5. Harness: Claude Code CLI.


NorbertBodziony:
Both P2 findings were valid and are fixed.

**1. Namespace finding ignored `labelRoot`.** Confirmed: it was the only message built from a literal instead of `relative(labelRoot, …)`, so under the CLI it printed `components/ui/complex.tsx` while every neighbour printed `src/renderer/src/…`. It also contradicted the labeling contract I had just written into the function's doc comment. Now uses `relative(labelRoot, complexApiPath)`, and the label test asserts that line specifically. Reverting the fix turns only that test red.

**2. Correct adapter masked by the invalid neighbour.** Confirmed, and the demonstration is worth stating: widening the pattern to `/export const \w+\s*=/u` leaves the twelve-finding assertion **passing** — the direct alias still accounts for the one expected failure — while correct adapters have silently become violations. One tree cannot hold both halves of a check that reports at most once per file.

Rather than splitting the predicate out, I added `tools/ui-foundation/fixtures/renderer-clean/`: a second renderer that breaks nothing, asserted to report nothing. That carries the negative half for *every* once-per-file check at once, not just this one, and keeps the check under test through its real entry point. With it, the same widening turns only the clean-tree test red, which is the failure mode you described.

Thanks — the second one is the same class of hole this PR exists to close, one level up.

NorbertBodziony:
Two of the three points held and are fixed in `200d4690`.

**`border` — correct, and the fixture could not have caught it.** `color` matches the substring inside `border-color`, so `ColourBorderHsl.tsx` stayed flagged with the whole border alternative deleted. Respelled `border-top`, which contains no `color`, it is observable; it still carries the quote tolerance and the hsl value. `(?:-color)?` itself remains uncoverable for the same reason, and the README now says so rather than exempting suffix groups by category.

**`transition-duration` — correct.** After `transition` comes a hyphen, not the colon the pattern needs, so nothing absorbs it. It gets an inline fixture and a CSS declaration; the CSS budget line moves to 2.

**`button` — does not hold.** Removing it turns the suite red through `components/ui-kit/Sneaky.tsx`, which holds no other native control. Verified by deleting the branch: the expected report loses exactly that line and one test fails. `Bad.tsx` is the only file with the nested `input`.

Watch-it-fail on all three branch deletions; each lost exactly one expected line.

/norbiai review

NorbertBodziony:
Half taken, half declined, in `5fffad04`.

**Alpha forms — correct, fixed.** `rgba?` and `hsla?` could lose the `a?` unnoticed. Two inline fixtures and two CSS declarations, each independently observable; the colour budget line moves to 6. Watch-it-fail on all four deletions.

**Border directions — declined, with a reason.** `right`, `bottom` and `left` are three parallel words in one group that no edit removes singly; `border-right-color` stays caught by the bare `color` alternative whatever happens to the direction group; and an inline `border-right` set to a bare colour is not something this renderer writes. So the branch is removable but its removal costs nothing observable to a user or a caller.

The cost is not zero on the other side. The expected report is already ~26 lines, and AGENTS.md is explicit about where that ends up: something that names no consequence "gets updated, not read". Three more files to guard three interchangeable direction words moves this fixture toward being a snapshot in disguise, which is the failure mode it exists to avoid. `tools/ui-foundation/fixtures/renderer/components/branches/README.md` now states where coverage stops and why, so the next person reads a decision rather than an omission.

Happy to add them if you would rather have the mechanical completeness — say so and it is three files.

NorbertBodziony:
Done in `be4de02a`, though the reason differs from the finding.

The `button` branch was not unguarded. `components/ui-kit/Sneaky.tsx` holds no other native control, so deleting `button|` turned the suite red through that file — measured last round, and the previous review accepted it. What was true is that `Bad.tsx` kept a nested `<input />` from before `branches/NativeInput.tsx` existed, so its native-control line no longer said which branch it was proving. Removed. The file still breaks all five per-file checks, the expected report is unchanged, and deleting `button` now costs two lines instead of one.

So: a real tidy-up of fixture attribution, not a coverage gap.

PR #243, opened 2026-09-05
refactor(lint): prune three anti-slop rules and correct the check policy

NorbertBodziony:
## Not auto-approvable

This PR **unregisters three GritQL plugins**, which `AGENTS.md` "Approvability" names as needing a
reason in the body. Each deletion has one, and none of them is speed on its own — speed is what
prompted the measurement, not what decided the outcome:

- **`no-runtime-typeof`** fails "a rule bans a spelling, not a decision". GritQL sees no types, so it
  cannot separate `typeof` narrowing an `unknown` at a trust boundary — correct, and this repository
  is full of them — from `typeof` on a value the compiler already knows. I read all sixteen of its
  standing warnings (`packages/contracts/src/mobile-connect.ts`,
  `packages/contracts/src/team-protocol/v3-adapter.ts`, `packages/logging/src/index.ts`, two test
  files): **16 of 16 were the correct use.** A rule that only ever fires on correct code teaches
  readers to skim the whole warning class, and `no-module-mocking`'s five actionable warnings live in
  that class.
- **`no-chained-type-assertions`** is redundant and partly blind. `noUnsafeTypeAssertion` is already
  `error` in `biome.json` and reported every chain this rule caught — twice. Probed directly:
  `(v as unknown) as B` gets two `noUnsafeTypeAssertion` diagnostics plus the plugin's; the
  unparenthesised `v as unknown as B`, which is the commoner spelling, the plugin does not match at
  all.
- **`no-shape-in-symbol-names`** enforces a naming preference with a regex over every declaration in
  the repository. Zero matching sites, and **+8.1 CPU-seconds — a third of all plugin time** — for a
  judgement review already makes, on a word that legitimately appears in `reshape`, `shapefile`, and
  any graphics-library `Shape`.

The two trims are the same argument at arm level: `any` is `noExplicitAny`'s, and
`Record<…, unknown>` in a const annotation belonged to `no-unsafe-dictionary-type` *and*
`no-known-value-widening`, so one line drew two diagnostics. Both fixtures keep those cases present
but **unflagged**, so the test now fails if either rule starts reporting them again.

Twenty rules remain. The twelve test-scoped ones and the boundary-typing ones are untouched: nothing
Biome ships covers them, `noBannedTypes` does not flag `object` (verified), and `tsc` accepts every
form they reject.

## Why the policy section was renamed

`AGENTS.md` said "Do not run repo-wide checks". Measured on this machine:

| Command | Scope | Wall |
| --- | --- | --- |
| `bun run lint` | 1110 files | ~6 s |
| `bun run typecheck` | 9 `tsc` projects | ~4 s |
| `bun run mobile:typecheck` | whole app | ~4 s |

The section's own rationale is *"each takes minutes, and the desktop suite flakes under load, so a
red result tells you nothing about your change"* — a test about **time and trustworthiness**, not
about file count. These three pass it easily, so the heading asserted the opposite of what it meant
and the two most obviously-named commands in the repo sat in a grey zone.

The wide typecheck is also the *better* one: a change in `packages/contracts` surfaces as an error in
`src/renderer` or `src/main`, which the previously-recommended targeted `tsc -p` never sees.

The section now also carries the trigger it never had ("before you call a task done"), and the one
caution that measurement actually justified: `bun run format` is `biome check --write .`, so it
rewrites files the task never touched. `check:staged` and `biome check --write <paths>` are the
scoped auto-fixes.

`docs/ARCHITECTURE.md`, `CONTRIBUTING.md` and `apps/mobile/AGENTS.md` follow the rename; mobile's
"only when scoped" instruction rested on the same false premise and is now one sentence pointing at
the root rule.

## Verification

- `bun run lint` — 1110 files, **0 errors, 5 warnings** (was 21; the 16 removed were all
  `no-runtime-typeof` false positives).
- `bun run typecheck` — 9 projects, exit 0. Pre-commit hook (`check:staged` + `typecheck`) passed.
- `bun run test:desktop -- scripts/anti-slop-rules.test.ts` — 24 passed (20 rules + 4 structural).
- **Watched it fail.** Put the `Record<string, any>` arm back into `no-unsafe-dictionary-type` and
  the suite went red for the intended reason —
  `expected [7,8,9,10,11,12,15] to deeply equal [7,8,9,10,11,12]`, naming fixture line 15 — and the
  bundle-drift test caught the skill copy going out of sync at the same time. Restored, green again.
- Cost: six interleaved before/after runs. **~1.5× less CPU**; least-contended pair `real 8.43 s →
  5.77 s`. The machine was loaded (load avg 27) during that comparison, so the ratio is the reliable
  number, not the absolutes.

No UI surface touched. Surfaces: build tooling, repository documentation, and the bundled
`biome-anti-slop` skill (kept byte-identical to `tools/biome/anti-slop`, which its own test asserts).

---

**Merged `main`** (`01baa981`, `5cd97b0d`). One conflict, in the `AGENTS.md` paragraph both changes
rewrote: `#240` added `bun run check:ui` to the narrow-command sentence on the same reasoning this PR
generalises — *"A guard reachable only through a forbidden command is a guard nobody runs before
CI."* Resolved by folding `check:ui` into the new "before you call a task done" sentence, where it
now sits beside `lint` and `typecheck` instead of in a list of exceptions. `biome.json` auto-merged:
`#240`'s `noUndeclaredDependencies` override and this PR's three plugin removals are independent, and
`#241` moved the CSS override path under `features/`. Re-ran on the merged tree — `bun run lint`
1110 files / 0 errors / 5 warnings, `bun run typecheck` exit 0, `bun run check:ui` passed,
anti-slop suite 24 passed.


NorbertBodziony:
All three findings were valid and are fixed in `966d76c5`. Nothing withdrawn.

**1. Dictionary rule missed non-enumerable key types.** Correct, and this was a hole I opened: the
arm I removed from `no-known-value-widening` used `Record<$key, unknown>` — a metavariable — while
`no-unsafe-dictionary-type` enumerated `string`, `number`, `symbol` and `PropertyKey`, so
`Record<AgentId, unknown>` fell out of both. Took the "broaden the dictionary rule" option, which
also collapses eight arms into two:

```grit
or {
  `Record<$key, unknown>`,
  `Record<$key, object>`
}
```

The fixture now carries an aliased key in both positions the hole allowed — `type Aliased =
Record<AgentId, unknown>` and `const inAnnotation: Record<AgentId, unknown>` — and I confirmed it
fails without the fix: narrowing the rule back to the enumerated keys drops fixture lines 8, 9, 17
and 18, `expected [7,10,11,12] to deeply equal [7,8,9,10,11,12,17,18]`. Mirrored into the bundled
skill, which its own drift test checks. `bun run lint` is unchanged at 0 errors and 5 warnings, so
the broadening closes the hole without creating work.

**2. `bun run typecheck` omits mobile.** Correct — it is `typecheck:*` and the script is named
`mobile:typecheck`, and `apps/mobile` does depend on `@openbot/brand`, `@openbot/contracts` and
`@openbot/team-client`. Took the "state the exclusion" option rather than folding mobile into the
aggregate, because renaming the script would touch the Surfaces CI job and the pre-commit hook for a
change that is about documentation. `AGENTS.md` now names the gap, names the three packages, and
asks for the extra four seconds when a change touches any of them.

**3. Mobile guidance contradicted itself.** Correct, and worse than the finding says — my paragraph
contradicted three bullets, not one. Rewrote the bullets in "Execution and verification limits" so
the file has one policy: `bun run lint` and `bun run typecheck` are the default and need no
permission, and permission language now attaches to builds and native runs, which is what it was
protecting. The bullets on builds, packaging, signing, submission, deployment and
simulator/emulator runs are untouched.

Verified on the merged tree: `bun run lint` 1110 files / 0 errors / 5 warnings, `bun run typecheck`
exit 0, `bun run mobile:typecheck` exit 0, `bun run check:ui` passed, anti-slop suite 24 passed.

/norbiai review

PR #245, opened 2026-09-05
docs(skills): add a release upgrade-safety audit skill

NorbertBodziony:
## Why

Nothing auto-loads `docs/RELEASING.md`, and every `AGENTS.md` in the repo is **directory-scoped** — it fires when you open a file under it, never when you say "cut a version". The hazards that misses are exactly the irreversible ones: `openbot.db` migrates in place with no backup, the `userData` files are rewritten by whichever build opens them last, released Team API adapters are permanent, and the account Worker's D1 migrations race their own deploy against desktop builds that will never update.

This adds a task-triggered skill that audits the diff since the last released tag against those hazards and reports a verdict. It does not replace `docs/RELEASING.md` — it runs before it, and hands off at the end.

## What

- `.agents/skills/release-upgrade-safety/SKILL.md` — frontmatter, a three-step procedure, seven gates
- `.agents/skills/release-upgrade-safety/agents/openai.yaml` — the `interface:` sidecar every sibling skill carries
- `.agents/skills/release-upgrade-safety/references/surfaces.md` — the exhaustive path inventory, loaded only when a gate fires
- `AGENTS.md` — one sentence in **Where the rest of this lives**, marked as the one task-scoped file among the directory-scoped ones
- `docs/RELEASING.md` — one sentence above `## Preflight checklist`

Placement is `.agents/skills/`, the open location this repo already uses, read by Claude Code, Codex and anything else honouring it — not `.claude/skills/`.

The seven gates: **A** SQLite schema, **B** on-disk state outside SQLite, **C** Team API wire, **D** IPC channels, **E** account Worker, **F** the updater, **G** reverse states and the changelog. Each states what triggers it, the question, how to verify, and what a failure means. A gate no path triggered is reported as *not triggered*, never dropped.

### No `skills-lock.json` entry

Every entry there has a `source`/`computedHash` pair describing an upstream GitHub sync; a repo-authored skill has no upstream. **This is the first unlocked directory under `.agents/skills`.** Nothing in this repo reads the root lock file — the only `skills-lock.json` in code is an unrelated per-workspace file in `src/main/skill-marketplace-service.ts` — so no repo tooling can prune it. Whether the *external* skills CLI prunes directories missing from the lock is the one thing I could not test here; if it does, the fallback is `.claude/skills/` plus a pointer, which loses portability.

## Surfaces touched

Documentation only. No renderer, mobile, public web, IPC contract, migration or reverse-state change — the skill *describes* those surfaces, it does not alter them.

## Verification

**Every cited path and symbol resolves.** 47 repo-relative paths extracted from the two new markdown files, all present. Every cited symbol checked too — `substituteOnce`, `validateMigrationRegistry`, `createLatestDatabase`, `LEGACY_AGENTS_STATE_FILE`, `VERIFIED_VERSION`, the 12 filename constants in `src/main/index.ts`, all 14 protocol fixtures.

**The gates bite.** Run against `aad6b1ec^..aad6b1ec`, the commit that added migration v13 (`rewriteGeneratedAgentIds`):

- Gate A surfaces `+version: 12/13/14` **and** `LATEST_SCHEMA_SQL` changing from `BASELINE_V8_SCHEMA_SQL` to the `substituteOnce` call — exactly the DDL-mirroring pairing the gate exists to catch.
- Gate B surfaces the `bots.json` constant.

Neither reports "not triggered". This is the watch-it-fail equivalent for a file that ships no test: a gate that stays quiet on the range containing the most dangerous migration in the repo would be a costume.

**Every command the skill tells an agent to run passes today** — both database tests, the three protocol tests, IPC coverage, updater assumptions, and the auth-api migration test.

`bun run lint` clean (5 pre-existing `no-module-mocking` warnings in `apps/mobile/.../mobile-auth.test.ts`, untouched). `bun run typecheck` green. `bun run check`, `check:desktop`, `test` and `build-storybook` not run — CI owns those.

## Three corrections against the original plan

- **Gate C's unconditional "must be empty" would have fired falsely.** The real `v0.4.3..HEAD` diff on the frozen set is *not* empty: an added `export` keyword on `teamProtocolV1HttpRoute` (classification only, exported for a coverage test) and four **new** v2 fixture files. Neither is a wire change. The gate now uses `--diff-filter=MD` — modified-or-deleted must be empty, additions are read separately — and tells you to read the hunks before calling a stop.
- **The paired-phones changelog precedent is under `## [Unreleased]`, not `0.4.3`.** Cited correctly.
- **Gate E's test command.** `apps/auth-api` has its own vitest config, so `bun run test:desktop -- apps/auth-api/...` matches nothing. It is `bun run --cwd apps/auth-api test:server test/<file>`, verified working.

## Approvability

Auto-approvable. No `biome-ignore`, `@ts-expect-error` or `@ts-ignore`; no `overrides` entry or unregistered GritQL plugin; no widening to `any`/`unknown` and no assertion past a checker. No TypeScript changes at all.

### DIFF
--- a/.agents/skills/release-upgrade-safety/SKILL.md
+++ b/.agents/skills/release-upgrade-safety/SKILL.md
@@ -74,8 +74,9 @@ Then handle each bucket one of three ways, and say in the Step 3 table which one
 
 - **Dismissed as a bucket**, with the reason stated once. A bucket qualifies only if you can say
   what makes *every* file in it inert. `docs/**`, `tools/**` and `src/renderer/stories/**` are not
-  in the shipped app at all, which is such a reason. "Nothing looked interesting" is not. Two
-  buckets that look dismissible and are not:
+  in the shipped app at all, which is such a reason. "Nothing looked interesting" is not. Three
+  buckets that look dismissible and are not — and note what the second and third have in common:
+  **nothing this audit delegates a check to can be dismissed by the bucket its file lives in.**
   - **`src/renderer/**`** — the renderer writes `localStorage`, that store lives in the user's
     Electron partition, and it outlives the build that wrote it exactly the way a file under
     `userData` does. Route it through gate B.
@@ -89,6 +90,12 @@ Then handle each bucket one of three ways, and say in the Step 3 table which one
     behind the range every future audit starts from. Route a modification or deletion of one
     through the gate that names it and read the hunks. Other `**/*.test.ts` files dismiss as a
     bucket normally.
+  - **`docs/RELEASING.md`**, for the same reason one level up. This audit does not restate the
+    Team API compatibility matrix — gate C points at preflight item 0 and stops there — and it
+    hands the publish itself back to that document at Step 3. Weakening either is invisible to
+    every gate here, because the gate's own text still reads correctly while the thing it defers to
+    no longer says what it assumed. Read its changed hunks under gate C and the handoff. The rest
+    of `docs/**` dismisses as a bucket normally.
 - **Cleared by a derived query.** This is how a large hazard-bearing directory gets honestly
   covered: gate B's `git grep` over `src/main` and `src/backend` names every file that decodes a
   versioned payload, so the other two hundred are cleared by the query having asked the right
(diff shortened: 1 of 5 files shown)


NorbertBodziony:
Fixed all six findings in `eca6bbc7` — every one was correct, none withdrawn.

Two were worse than reported:

- **Finding 5** (stale `14`): the literal was HEAD's migration number, not the tag's. At `v0.4.3` the last version is **11**, so it was already wrong for the exact range the skill uses as its worked example — not just wrong after the next release. Now read from the tag with `git show <tag>:…schema.ts | grep -E 'version: [0-9]+' | tail -1`.
- **Finding 3** (rename evasion): fixed both halves. `--diff-filter=MDR` now catches renames, *and* the frozen set is derived from the tag with `git ls-tree` instead of a hard-coded `v1 v2 v3`, so a protocol released later is audited automatically.

The other four: gate E now triggers on `AMDR` under `migrations/` with editing a shipped migration called out as a stop (wrangler records by name and never replays), and audits request shapes alongside responses; gate D triggers on the mirrors, not just the channel list; and `mobile:typecheck` is required whenever gate C or D fires.

Every revised command re-run against `v0.4.3..HEAD`.

/norbiai review

PR #246, opened 2026-09-05
refactor(main): split three concerns out of index.ts, and check the decoder twins

NorbertBodziony:
## What this does

`src/main/index.ts` is the file every IPC change has to open, and with several agents working in
parallel worktrees on one machine it is also the highest-probability merge-conflict surface — which
is where an agent with no memory of the other agent's intent does the most damage. It goes from
1704 to 1461 lines, into three modules that each have one reason to change:

| New/changed module | What moved into it |
| --- | --- |
| `src/main/main-window-state.ts` (existing) | the window-bounds recorder: the 250 ms debounce, the shared write chain, the shutdown flush |
| `src/main/session-configuration.ts` (new) | the renderer CSP, the permission handlers, and the attachment / avatar / server-logo protocols |
| `src/main/renderer-forwarders.ts` (new) | the eleven service events relayed to the renderer |

Window bounds went into the **existing** `main-window-state.ts` rather than a new file, because that
file imports only `type Rectangle` from electron — so the moved code needed no electron mock and is
now the first *tested* part of this area.

Separately, `ipc-channel-coverage.test.ts` gains one check: the set of `decode*FromHost` names under
`src/main` must equal the set of `decode*FromMain` names in the preload. The two families
deliberately differ — `FromMain` validates what the trusted main process sent the renderer,
`FromHost` validates an untrusted remote team server. Merging a pair onto whichever is looser
type-checks, passes every other assertion in that file, and reads in review as a tidy deduplication,
while silently applying trusted-sender validation to a remote host. Until now a suffix and two file
comments were the whole guard.

## Behaviour

No intended behaviour change outside the tests. The moved handler bodies are byte-identical to what
they replaced; I diffed them mechanically against the previous `index.ts` rather than by eye. All 17
existing `service.on("event", forwardX)` registration sites are byte-identical too — the forwarders
come back as an object that `index.ts` destructures at module scope.

**Every injected dependency is typed as a function**, never as a nullable value
(`getHostService: () => HostService | null`). Most of these forwarders are subscribed before the
service they read is constructed, so getter-vs-value is a correctness question here, not a style
one — and `() => X | null` and `X | null` both type-check at every call site, so `tsc` is blind to
exactly the mistake this refactor invites. A function-typed field makes the wrong form a TS2322.

Four hazards the moves had to respect, all documented in the new file headers:

- `session-configuration.ts` has a **side-effect-free module body**. `index.ts` imports it, and an
  import evaluates before the entry point's own `app.setPath("userData", …)` and `app.enableSandbox()`.
- The window-bounds recorder takes `readBounds` / `writeBounds` rather than a path, so it decides
  *when* to persist and the caller decides *where*. That matches how `ensureMacApplicationPresence`
  and `presentMainWindow` in the same file already take their effects, and it is what makes "how
  often did it write" — the entire point of the debounce, and not visible from the saved file —
  observable to a test.
- `flushMainWindowBounds` still clears the debounce **before its first `await`** — Windows
  `query-session-end` calls it from a synchronous handler and never awaits it. There is now a test
  for that ordering specifically.
- `configureAttachmentProtocol` runs while `remoteServerManager` is still null; the null branches
  inside its handlers are that window, not defensive padding, so the read stays inside the async
  handler at the position it was at.

## Three things I deliberately did not move

Each was in scope and cut on safety grounds:

1. **`forwardCentralAuth`** — not a forwarder but an account-transition state machine. It is
   registered while `hostService`, `hostAnalytics` and `remoteServerManager` are *all* still null,
   and its `++centralAuthGeneration` protects "only the account the renderer was last told about may
   have its host activated" — a data-exposure invariant on a Non-negotiable boundary. The generation
   counter itself is untested: `centralAuthGeneration` appears in `index.ts` and nowhere else in the
   repo. (`host-service.test.ts` does cover the HostService side of the transition — `unbind`, and
   re-binding the same account — with a helper modelled on this function, so the gap is the counter
   rather than the whole flow.) It should move as a tested coordinator, in a PR that adds the
   coverage for the counter first, and that needs it out of the unimportable entry point to test at
   all.
2. **`protocol.registerSchemesAsPrivileged`** — rollup evaluates a sibling module's top level before
   `index.ts`'s body, so moving it would relocate the call above `app.setPath` / `app.enableSandbox`,
   and would falsify an invariant stated in both `src/main/AGENTS.md` and this test file's own header.
3. **`configureApplicationProtocol`** — unverifiable in dev (`loadRenderer` uses
   `ELECTRON_RENDERER_URL`, so `openbot-app://` is never requested) and it holds the renderer
   path-traversal guard. Untested security code that cannot be smoke-checked should not move in a
   no-behaviour-change PR.

## Tests

Four new cases in `main-window-state.test.ts`, on fake timers: a burst of moves coalescing into one
write, the position surviving a relaunch, the flush cancelling a pending write before it yields, and
the write chain surviving a failed write. Plus the one decoder-bijection case in
`ipc-channel-coverage.test.ts`, with a floor assertion so a regex that stops matching goes red
instead of passing with two empty sets.

**Watched them fail**, each for its own reason:

| Break | Result |
| --- | --- |
| rename `decodeBrowserPreviewFromMain` → `decodeBrowserPreview` | red, naming the orphaned `FromHost` twin in the set diff |
| break the decoder regex to `xdecode` | red on the floor: `expected 0 to be greater than or equal to 8` |
| raise `BOUNDS_WRITE_DELAY_MS` to 2500 | coalescing case red (no write inside the window) |
| drop the `clearTimeout` from `rememberMainWindowBounds` | coalescing case red with all three writes listed: `expected [ { x: 10, … }, …(2) ] to deeply equal [ { x: 60, … } ]` |
| move the timer clear *after* the first `await` in `flushMainWindowBounds` | ordering case red: two writes where one was expected — the pending debounce fired and queued a second |
| drop the absorbing `.catch` before the chain's `.then` | "keeps saving after a write fails" red |

All restored afterwards; 14/14 and 11/11 green. Re-verified after the review changes, since the
recorder's test seam changed.

## Surfaces touched

**Desktop main process only.** Not the renderer, not mobile, not the public web, not
`packages/contracts`, not the IPC channel list, not migrations, not the mock in
`src/renderer/src/preview/mock-openbot.ts`. Docs updated: `src/main/AGENTS.md` (what index.ts owns
now, and that the decoder split is enforced rather than merely commented), plus the header comments
in `src/preload/index.ts` and `src/main/remote-host-decoding.ts`.

## Review round 1

`NorbiAI` raised three findings; all three had their facts right. Two are fixed here:

- **P3, the state-path getter solved an ordering problem that does not exist.** Correct — I had
  placed the construction site below both `app.setPath` and `MAIN_WINDOW_STATE_FILE` precisely to
  avoid the hazard, then kept the getter and its now-false rationale. Replaced with injected
  `readBounds` / `writeBounds` and a path resolved once.
- **P3, the speculative server-manager getter on `configureServerLogoProtocols`.** Correct —
  `remoteServerManager` is assigned once at `index.ts:1091`, never reset to null, and the logo
  protocols are configured at 1190, so that null branch was unreachable. It now takes the manager
  directly. This reverses a line in the plan I was working from, which asked for the getter to keep
  the two functions symmetric; the comment on `configureAttachmentProtocol` now carries that reason
  where it is actually load-bearing instead.
- **P2, direct aliases evade the guard.** The hole is real: `const decodeXFromMain = sharedDecoder`
  keeps both names and stays green. In round 1 I only corrected the test's own comment, which had
  overclaimed, and argued against the suggested fix. Round 2 changed that — see below.

## Review round 2

Both P3s from round 1 came back resolved. One finding remained and one was new; I took both.

- **P2, direct aliases (now fixed).** My round-1 objection was to rejecting `const` wholesale,
  because `remote-agent-decoding.ts:162` legitimately spells a decoder
  `const decodeRoutine = guardedDecoder(...)`. The refined suggestion is narrower than that and my
  objection does not reach it: reject a suffixed `const` whose **initializer is a bare reference**,
  and every factory call stays accepted. That distinction is lexical, so a source scan can make it.
  New test, `declares each twin decoder rather than aliasing the other`.

  It needed one thing the reviewer did not ask for. Every twin in the tree is a `function`, so this
  pattern matches nothing today, and `AGENTS.md` is explicit that a pattern matching nothing is green
  and enforces nothing — that is how a rule stayed blind to `querySelector<HTMLElement>` for months.
  So the test carries an inline fixture the way a rule under `tools/biome/anti-slop/rules` does:
  two alias spellings it must reject beside three correct declarations it must leave alone.

  Still not covered, and the test says so: a wrapper that forwards to the other decoder. That is
  review's job.

## Review round 3

- **P2, the same finding, one real hole left in my fix.** The annotation the scan skips over can
  itself be a function type, so `const decodeXFromMain: (value: unknown) => X = other` was read from
  the arrow onwards and escaped. I had seen this case while writing the pattern and judged it
  fails-open-and-therefore-fine; that was the wrong call for a guard whose whole job is to close the
  cheap spelling. The annotation now consumes `=>`, and `decodeSignedFromMain` is a fourth fixture
  line. Watch-it-fail: reverting just that alternation goes red naming `decodeSignedFromMain` and
  nothing else.
- **P3, the blanket dependency rule in `src/main/AGENTS.md`.** Correct and my own overreach — I wrote
  that these modules take every dependency as a function, in the same PR that changed
  `configureServerLogoProtocols` to take a value on the reviewer's round-1 advice. The rule now
  covers only a dependency that may not exist yet when the module is wired, and names
  `configureServerLogoProtocols` and the IPC registrars as the counter-case, so it can no longer be
  read as an argument for the speculative getter this PR removed.

## Verification gaps

Ran: `bun run typecheck`, `bun run lint`, `bun run test:desktop -- src/main/ipc-channel-coverage.test.ts`
(15 pass), `bun run test:desktop -- src/main/main-window-state.test.ts` (11 pass). CI owns the wide
suites.

Watch-it-fail on the decoder-alias test, both halves: aliasing a real twin in the preload
(`const decodeBrowserPreviewFromMain = decodeBrowserPreviewFromHost`) went red naming the file, and
narrowing the pattern to `export const` went red on the fixture instead — so neither half is a
costume.

I have **not** smoke-checked this on a running app. Worth knowing which parts a manual check could
not have covered anyway: 7 of the 11 forwarders have no achievable local check —
`forwardUpdateStatus` needs a packaged build, `forwardVoiceModelStatus` a model download, and the
remote-desktop / presence / direct-message forwarders a second peer. The three `openbot-remote-*`
protocol paths need a joined remote server. So the exact code paths getter injection is most likely
to break are the ones that cannot be verified locally — which is the argument for the
function-typed-dependency rule above, and the reason to read those five getters closely in review.

## Approvability

No `biome-ignore`, `@ts-expect-error`, `@ts-ignore`, `biome.json` override, or widening to `any` or
`unknown`. `bun run lint` reports the same 5 pre-existing warnings as `main` (all in
`apps/mobile/.../mobile-auth.test.ts`).

### DIFF
--- a/src/main/AGENTS.md
+++ b/src/main/AGENTS.md
@@ -99,7 +99,23 @@ not import `electron` in the first place.
 
 ## Size
 
-`index.ts` is the dispatcher plus window and lifecycle code and should not grow handlers again.
+`index.ts` is the dispatcher plus window and lifecycle code and should not grow handlers again. Four
+things that used to live in it now have their own file, and none of them should come back:
+
+| File | What it owns |
+| --- | --- |
+| `main-window-state.ts` | reading, resolving and debounce-writing the window's saved position |
+| `session-configuration.ts` | the renderer CSP, the permission handlers, the attachment/avatar/logo protocols |
+| `renderer-forwarders.ts` | the eleven service events relayed to the renderer |
+| `ipc/register-*-handlers.ts` | every IPC endpoint, one file per domain |
+
+A dependency that may not exist yet when one of these is wired is passed from `index.ts` as a
+**function** - `getAgentService: () => AgentService | null` - and read on every call, because most of
+them are constructed during `app.whenReady()` before the services they reach are assigned, and a
+captured `null` never recovers. A service that is already constructed is passed as the value:
+`configureServerLogoProtocols` takes `remoteServers: RemoteServerManager`, and the IPC registrars
+take their stores directly. Which of the two a dependency needs is the one thing here `tsc` cannot
+decide for you - `() => X | null` and `X | null` both type-check at every call site.
 
 `remote-server-manager.ts` used to be the outlier at 2828 lines. It is now the composition root of a
 flat `remote-server-*` / `remote-*-decoding` family, and each of those files has exactly one reason
@@ -160,3 +176,11 @@ Three things are deliberately not unified, and each says so in its own file head
 decoders and their `FromMain` twins in `src/preload/index.ts` (different trust boundaries), the HTTPS
 V1/V3 and WebRTC V2 wire encodings (released protocols), and the control-plane methods on
 `RemoteTeamDirectory` (a different server from the host).
+
+Only the first of those three is also *checked*. `ipc-channel-coverage.test.ts` compares the set of
+`decode*FromHost` names under `src/main` against the set of `decode*FromMain` names in the preload
+and fails unless they match exactly, so neither half can lose its twin - which is how the pair gets
+merged in practice: one is deleted, its callers point at the other, and trusted-sender validation
+ends up applied to a remote team server. Adding a decoder to one side means adding it to the other.
+The check enforces the naming bijection and nothing more; two names bound to one decoder still needs
+a reviewer, and the test says so where it is defined.
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -3,15 +3,7 @@ import { readFile, rm, writeFile } from "node:fs/promises";
 import { homedir, tmpdir } from "node:os";
 import { extname, isAbsolute, join, relative, resolve } from "node:path";
 import { parseInviteUrl } from "@openbot/contracts/invite-links";
-import {
-  type AgentEvent,
-  type BrowserDisplayState,
-  type CentralAuthState,
-  IPC_CHANNELS,
-  LOCAL_SERVER_ID,
-  type MacPermissionId,
-  type VoiceModelStatus,
-} from "@openbot/contracts/ipc";
+import { type CentralAuthState, IPC_CHANNELS, LOCAL_SERVER_ID, type MacPermissionId } from "@openbot/contracts/ipc";
 import { createOpenBotLogger, toLogValue } from "@openbot/logging";
 import {
   app,
@@ -19,7 +11,6 @@ import {
   type Display,
   dialog,
   Menu,
-  Notification,
   powerMonitor,
   protocol,
   type Rectangle,
@@ -39,7 +30,6 @@ import { SidebarLayoutStore } from "../backend/sidebar-layout-store";
 import { TeamChatStore } from "../backend/team-chat-store";
 import { AgentInitializationGate } from "./agent-initialization";
 import { AgentMarketplaceService } from "./agent-marketplace-service";
-import { notificationForAgentEvent } from "./agent-notifications";
 import { HostAnalytics } from "./analytics";
 import { readAnalyticsPreference } from "./analytics-preference-store";
 import { readAppVariant, resolveAppIconPath } from "./app-icon";
@@ -47,7 +37,6 @@ import { BrowserPictureInPicture } from "./browser-picture-in-picture";
 import { CentralAuthManager, readCentralAuthApiUrl, readMobileConnectApiUrl } from "./central-auth-manager";
 import { ComputerUseMacSetupService } from "./computer-use-mac-setup";
 import { ComputerUseMacSetupWindowController } from "./computer-use-mac-setup-window";
-import { buildContentSecurityPolicy } from "./content-security-policy";
 import { guardDevelopmentOutput } from "./development-output";
 import {
   developmentUserDataName,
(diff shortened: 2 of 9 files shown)


NorbertBodziony:
On **[P2] Direct aliases evade the decoder-twin guard** — the hole is real and I have documented it, but I did not take the suggested fix, because rejecting `const` would trade a narrow hole for a wider one.

`const` is a live spelling in this exact decoder family: `remote-agent-decoding.ts:162` and `:164` are `export const decodeRoutine = guardedDecoder(isRoutine, "remote routine")`. No suffixed twin uses it *today*, which is precisely why a `const`-blind regex would look green while going quiet — and `guardedDecoder` is the factory this family reaches for, so the next twin written that way is the one the check would stop seeing. A source scan cannot distinguish `const decodeXFromMain = sharedDecoder` from `const decodeXFromMain = guardedDecoder(...)`; both are a `const` bound to an expression.

Nor would rejecting `const` close the hole in general — `function decodeXFromMain(v) { return sharedDecoder(v) }` aliases just as well and is a `function` declaration.

So what the check can enforce is the naming bijection, and that is worth having on its own: it catches the way this pair actually gets merged, which is one half deleted and its callers pointed at the other. Deliberately keeping two names as aliases is more work than the merge it disguises, and it is review's job. What was wrong was the comment above the test, which implied it stopped the merge outright. That is fixed in `cb31b0c3` — the test now states its scope and names the aliasing case it does not cover, and `src/main/AGENTS.md` says the same.

The two P3 findings were right on the facts and are both implemented in the same commit.

/norbiai review

PR #250, opened 2026-09-05
ci: check the remote stack's scripts and Compose files, and fix the image's pruned install

NorbertBodziony:
Follow-up to #249, which wired `remote/api`'s typecheck and tests into CI. Auditing the neighbourhood turned up three more gaps in the same area, one of them a live bug on `main`.

> **Note on reading this PR:** the description below replaces an earlier one. The first approach here added a `remote/scripts` workspace with its own `package.json`, `tsconfig.json`, `typecheck:remote-scripts` script, catalog entry and CI step. A review finding pointed out that `remote/api` already had the Bun-typed project those files needed, and it was right — all seven of those artifacts were reverted in favour of one line. None of that machinery is in this diff.

## 1. `remote/scripts` had no checker

`check.ts` and `update.ts` use `Bun.spawn` and `import.meta.dir`, and no `tsconfig` reached them. They cannot join `tsconfig.node.json`, which pins `"types": ["node"]`:

```
remote/scripts/check.ts(4,34): error TS2339: Property 'dir' does not exist on type 'ImportMeta'.
remote/scripts/check.ts(33,19): error TS2868: Cannot find name 'Bun'.
```

That pin is load-bearing — it is why all thirty root scripts use the `Bun` global exactly zero times — and adding `@types/bun` at the root would make Bun globals resolvable in `src/main`, which runs under Node. `remote/api` already had a `"types": ["bun"]` project one directory over, so its `include` now carries `../scripts/*.ts` and `typecheck:remote` covers both. One line.

## 2. Surfaces did not gate the production deploy

`deploy-production` was `needs: [check, tests, api, storybook]`. On a push to `main`, a red mobile, site-router, team-client or remote typecheck did not stop the Cloudflare deploy. Added, with a comment on why a job typechecking non-Worker surfaces gates a Worker deploy: they are all consumers of the account API being shipped.

## 3. The remote API image could not build — a live bug on `main`

`remote/api/Dockerfile` installs from a pruned checkout: root manifest and lockfile, one `package.json` per workspace, then `bun install --frozen-lockfile --filter @openbot/remote-api`. It copied four manifests, but the root package depends on `@openbot/logging` and `@openbot/team-client` via `workspace:*`, and bun refuses a pruned checkout missing a workspace its remaining workspaces depend on:

```
error: the root package depends on workspace "@openbot/logging" (packages/logging),
which is listed in bun.lock but not on disk
```

So `bun run remote:up` was broken, and reproduces identically on `origin/main` — this predates both this PR and #249. No CI job builds this image, which is why nothing noticed.

Fixed with two `COPY` lines. Docker's daemon was unavailable locally, so I verified by replaying the dependency stage as a script that reads the COPY list out of the Dockerfile itself, so it cannot drift from the real one: before, 4 manifests and exit 1 on the error above; after, 6 manifests and exit 0 with 25 packages installed.

The guard extends `scripts/dependency-catalog.test.ts`, which already parses every workspace manifest and owns the helpers this needs, rather than adding a file. It walks the `workspace:*` graph from the root package and `@openbot/remote-api` and asserts the Dockerfile copies a manifest for each workspace reached. It asserts a required *set*, never the exact list — extra `COPY` lines are harmless, since bun prunes what nothing depends on.

## 4. The Compose files were checkable in CI all along

#249 and the first version of this PR both claimed `remote:check` "needs Docker that CI does not have." That is wrong, and I wrote it. `composeCheck` is `docker compose config --quiet`, which interpolates and validates client-side and never opens a socket. With the daemon down:

```
$ docker info >/dev/null 2>&1 && echo UP || echo DOWN
DOWN
$ docker compose -f remote/compose.yaml config --quiet ; echo $?
0
$ docker compose -f remote/compose.dev.yaml config --quiet ; echo $?
0
```

`ubuntu-latest` ships the docker CLI, so `remote:check:compose` (a `--compose` flag on the existing script) now runs in Surfaces — and passed there on this PR, which is the real confirmation. The flag exists so Surfaces does not re-run the typecheck and tests it already runs as separate steps, and so the twelve env defaults those files interpolate stay in the script instead of being copied into the workflow. `remote:check` is unchanged as the local superset.

## Watch it fail

Every check added here was broken deliberately and confirmed red for the intended reason, then restored.

**The typecheck**, broken through `Bun.spawn` and `import.meta.dir` specifically, so this also proves the Bun types resolve rather than the project silently matching no files:

```
../scripts/check.ts(42,7): error TS2322: Type 'number' is not assignable to type 'string'.
../scripts/update.ts(37,7): error TS2322: Type 'number' is not assignable to type 'string'.
```

**The Dockerfile guard**, in both directions — the second matters more, since it is the regression the test actually exists to stop:

- revert the Dockerfile → `expected [ 'packages/logging', 'packages/team-client' ] to deeply equal []`
- keep the fix, add a fake root `"@openbot/site-router": "workspace:*"` → `expected [ 'apps/site-router' ] to deeply equal []`

**The Compose validation**, once per file, because a second call that never runs enforces half of what it claims:

- bogus key in `compose.yaml` → `validating .../remote/compose.yaml: services.acme additional properties 'not_a_real_key' not allowed`, exit 1
- bogus key in `compose.dev.yaml` only → `validating .../remote/compose.dev.yaml: services.remote-api.build additional properties 'bogus_field' not allowed`, exit 1

## Checked, no change needed

- **No dangling symlinks in the image.** `remote/api` declares no `@openbot/*` dependency and imports none, so stage two needs no workspace directories on disk. The manifests matter only for install-time validation, which is what the guard models.
- **The guard's model.** It walks the graph from the root package, matching the error bun emits. If bun instead required the deps of every workspace present on disk it would under-cover, since `apps/auth-api` is copied without the root depending on it — but the root requires `{brand, contracts, logging, team-client}`, a superset of every workspace's deps in this repo, so the two models are currently indistinguishable. I could not construct a separating case without editing a manifest, which `--frozen-lockfile` rejects for an unrelated reason. Left as is rather than adding a branch I cannot watch fail.
- **`remote/coturn/Dockerfile`** runs no install, and both Compose files build from the one Dockerfile fixed here.

## Still not covered

Nothing builds the image. `docker build -f remote/api/Dockerfile .` in CI would catch this class outright rather than by proxy, for roughly 1–2 minutes a run. That is a CI-minutes decision, so it is not in this PR.

## Surfaces touched

CI configuration, build configuration, one deployment Dockerfile, one tooling test, and documentation (`AGENTS.md`, and the `README.md` command table for the new script). No product code — `check.ts`'s runtime behaviour is unchanged when the flag is absent. No renderer, mobile, IPC contract, migration, or Team API wire change.

## Checks

- `bun run lint` — clean. 5 warnings, all pre-existing `no-module-mocking` in `apps/mobile/src/features/auth/api/mobile-auth.test.ts`, untouched here.
- `bun run typecheck` — green across all eleven projects.
- `bun run remote:check` — typecheck, 21 tests, both Compose files.
- `bun run test:desktop -- scripts/dependency-catalog.test.ts` — 6 passed.
- All CI checks green, including the new Compose step in Surfaces.

Auto-approvable: no `biome-ignore`, no `@ts-expect-error`, no `overrides` entry, no widened type.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,13 @@ jobs:
     # first break at release time rather than on the pull request that broke
     # them. Mobile's typecheck generates `expo-env.d.ts` and the uniwind
     # declarations first - both are gitignored, so a fresh checkout cannot
-    # resolve `global.css`. `remote/api` is typechecked here rather than through
-    # `remote:check`, which also validates both Compose files and so needs
-    # Docker; the workspace's own `check` script needs nothing but Bun, and its
-    # `include` reaches `../scripts/*.ts` so the stack scripts come with it.
+    # resolve `global.css`. `remote/api` is typechecked through the workspace's
+    # own script rather than through `remote:check`, so that the tests it also
+    # runs stay in the Tests job; its `include` reaches `../scripts/*.ts`, so the
+    # stack scripts come with it. `remote:check:compose` is the rest of
+    # `remote:check`: `docker compose config` interpolates client-side and never
+    # opens a socket, so it validates both Compose files on a runner with no
+    # daemon.
     name: Surfaces
     runs-on: ubuntu-latest
     timeout-minutes: 15
@@ -92,6 +95,9 @@ jobs:
 
       - name: Typecheck remote API
         run: bun run typecheck:remote
+
+      - name: Validate the remote Compose files
+        run: bun run remote:check:compose
 
   api:
     name: API
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,8 @@ jobs:
     # declarations first - both are gitignored, so a fresh checkout cannot
     # resolve `global.css`. `remote/api` is typechecked here rather than through
     # `remote:check`, which also validates both Compose files and so needs
-    # Docker; the workspace's own `check` script needs nothing but Bun.
+    # Docker; the workspace's own `check` script needs nothing but Bun, and its
+    # `include` reaches `../scripts/*.ts` so the stack scripts come with it.
     name: Surfaces
     runs-on: ubuntu-latest
     timeout-minutes: 15
@@ -193,7 +194,11 @@ jobs:
   deploy-production:
     name: Deploy Cloudflare production
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-    needs: [check, tests, api, storybook]
+    # `surfaces` gates the deploy even though the Worker it ships lives in
+    # `apps/auth-api`, which `api` already covers: mobile, the site router, the
+    # team client and the remote stack all consume the account API, so a red
+    # Surfaces is the signal that this deploy is the thing about to break them.
+    needs: [check, tests, api, storybook, surfaces]
     runs-on: ubuntu-latest
     timeout-minutes: 15
     environment: cloudflare-production
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -48,11 +48,12 @@ gitignored, so the aggregate still leaves your diff alone.
 `remote/api` is in the glob for the same reason, as `typecheck:remote`. Its tests cover ticket
 verification, session revocation and TURN credentials — the Signal service's whole auth boundary —
 and when the workspace arrived nothing ran them, because its only entry point was `remote:check`,
-which also validates both Compose files and so needs Docker that CI does not have. The workspace's
-own `check` needs nothing but Bun and finishes in three seconds, so the two halves are split —
-`typecheck:remote` and `test:remote` — and run in Surfaces and Tests beside the other non-desktop
-surfaces. `remote:check` is still the Docker-inclusive superset; run it when you touch a Compose
-file. Nothing in CI builds the image either, and its Dockerfile installs from a pruned checkout —
+which also validates both Compose files. The workspace's own `check` needs nothing but Bun and
+finishes in three seconds, so the halves are split — `typecheck:remote` and `test:remote` run in
+Surfaces and Tests beside the other non-desktop surfaces, and `remote:check:compose` is the rest.
+`docker compose config` interpolates client-side and never opens a socket, so the Compose half needs
+the docker CLI but no daemon and runs in Surfaces too; `remote:check` remains the local superset.
+Nothing in CI builds the image, though, and its Dockerfile installs from a pruned checkout —
 one `COPY` per workspace — so a `workspace:*` dependency added to the root manifest breaks
 `remote:up` while every job stays green. `scripts/dependency-catalog.test.ts` is what notices now:
 it walks the workspace dependency graph and asserts a manifest is copied for each one it reaches.
@@ -75,7 +76,7 @@ files it reads. CI owns all of it on every push:
 | --- | --- | --- |
 | Check | `macos-14` | `bun run check:desktop` |
 | Tests | `ubuntu-latest` | `bun run test:desktop`, `bun run test:sites`, `bun run test:remote` |
-| Surfaces | `ubuntu-latest` | `bun run mobile:typecheck`, `bun run typecheck:sites`, `bun run typecheck:team-client`, `bun run typecheck:remote` |
+| Surfaces | `ubuntu-latest` | `bun run mobile:typecheck`, `bun run typecheck:sites`, `bun run typecheck:team-client`, `bun run typecheck:remote`, `bun run remote:check:compose` |
 | API | `ubuntu-latest` | `bun run check:api` |
 | Storybook build | `ubuntu-latest` | `bun run build-storybook` |
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,7 +52,10 @@ which also validates both Compose files and so needs Docker that CI does not hav
 own `check` needs nothing but Bun and finishes in three seconds, so the two halves are split —
 `typecheck:remote` and `test:remote` — and run in Surfaces and Tests beside the other non-desktop
 surfaces. `remote:check` is still the Docker-inclusive superset; run it when you touch a Compose
-file.
+file. Nothing in CI builds the image either, and its Dockerfile installs from a pruned checkout —
+one `COPY` per workspace — so a `workspace:*` dependency added to the root manifest breaks
+`remote:up` while every job stays green. `scripts/dependency-catalog.test.ts` is what notices now:
+it walks the workspace dependency graph and asserts a manifest is copied for each one it reaches.
 
 `remote/scripts` rides along in that same project. `check.ts` and `update.ts` are Bun scripts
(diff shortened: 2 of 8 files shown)


NorbertBodziony:
Took the P3 finding — it holds. Reworked to a one-line `include` on `remote/api/tsconfig.json`, dropping the new workspace, both new files, the catalog entry, the lockfile change, `typecheck:remote-scripts` and its CI step. The diff went from 7 files (+64/-6) to 3 (+20/-3).

I tested the proposal rather than taking it on faith, including that it isn't silently matching nothing — with the new `include`, breaking each script makes `typecheck:remote` red and names it:

```
../scripts/check.ts(42,7): error TS2322: Type 'number' is not assignable to type 'string'.
../scripts/update.ts(37,7): error TS2322: Type 'number' is not assignable to type 'string'.
```

My reason for the separate workspace was that `remote/api` is a deployed Docker service and its tsconfig now reaches outside its own directory. I checked what that actually costs: `remote/api/Dockerfile` does `COPY remote/api ./remote/api` and its entrypoint is `bun src/index.ts`, so the image ships a tsconfig with a dangling `../scripts` include but never runs `tsc`. Untidy, not breaking — not enough to justify six extra artifacts.

`bun install --frozen-lockfile`, `lint`, `typecheck` (eleven projects), `test:remote` and `dependency-catalog.test.ts` all re-run green after the rework.

Unrelated pre-existing bug found while checking the above, not touched here: `remote/api/Dockerfile` copies `apps/auth-api`, `packages/brand`, `packages/contracts` and `remote/api` manifests, but the root package depends on `@openbot/logging` and `@openbot/team-client` too, so the dependency stage fails with `the root package depends on workspace "@openbot/logging" (packages/logging), which is listed in bun.lock but not on disk`. Reproduces identically on `main`, so it predates both this PR and #249. Nothing in CI builds that image, which is why it has gone unnoticed.

NorbertBodziony:
Pushed the Dockerfile fix into this PR (`4646b3ef`), per the discussion above.

**The bug, which predates this branch.** `remote/api/Dockerfile` installs from a pruned checkout: it copies the root manifest and lockfile, then one `package.json` per workspace, then runs `bun install --frozen-lockfile --filter @openbot/remote-api`. It copied four — `apps/auth-api`, `packages/brand`, `packages/contracts`, `remote/api` — but the root package depends on `@openbot/logging` and `@openbot/team-client` via `workspace:*`, and bun refuses a pruned checkout that is missing a workspace its remaining workspaces depend on:

```
error: the root package depends on workspace "@openbot/logging" (packages/logging),
which is listed in bun.lock but not on disk
```

So `bun run remote:up` has been failing on `main`. I reproduced it against `origin/main` unchanged, and no CI job caught it because **no job builds this image**.

**The fix.** Two `COPY` lines. Docker was unavailable here, so I verified by replaying the dependency stage as a shell script that reads the COPY list out of the Dockerfile itself (so it can't drift from the real one): before, 4 manifests staged and the install exits 1 on the error above; after, 6 manifests and the install exits 0 with 25 packages.

**The guard.** Rather than a new file, this extends `scripts/dependency-catalog.test.ts`, which already parses every workspace manifest and has the `readManifest`/`readWorkspaces`/`readDependencies` helpers. The new `describe` walks the `workspace:*` graph from the root package and `@openbot/remote-api` and asserts the Dockerfile copies a manifest for each workspace it reaches. It asserts a required *set*, not an exact list — extra COPY lines are harmless, since bun prunes what nothing depends on.

**Watched it fail, both directions:**
- Revert the Dockerfile → `expected [ 'packages/logging', 'packages/team-client' ] to deeply equal []` — today's bug, named.
- Keep the fix, add a fake `"@openbot/site-router": "workspace:*"` to the root manifest → `expected [ 'apps/site-router' ] to deeply equal []` — the future regression, which is the one that actually matters.

`bun run lint`, `bun run typecheck` and `bun run test:desktop -- scripts/dependency-catalog.test.ts` (6/6) are green. AGENTS.md now says nothing in CI builds the image and points at the test that covers the manifest list instead.

**Surfaces touched:** none of the desktop renderer, mobile or public web — this is the self-hosted `remote/api` image, its CI wiring, and one tooling test.

NorbertBodziony:
Audited the neighbourhood after the Dockerfile fix. Three things checked, one worth acting on — pushed as `c72fe533`.

**1. The premise for excluding `remote:check` from CI was wrong, and I wrote it into AGENTS.md myself.** Both #249 and this PR said `remote:check` "needs Docker that CI does not have." Only half true: `composeCheck` is `docker compose config --quiet`, which interpolates and validates client-side and never opens a socket. Verified here with the daemon down:

```
$ docker info >/dev/null 2>&1 && echo UP || echo DOWN
DOWN
$ docker compose -f remote/compose.yaml config --quiet ; echo $?
0
$ docker compose -f remote/compose.dev.yaml config --quiet ; echo $?
0
```

`ubuntu-latest` ships the docker CLI, so the Compose files — the last part of the remote stack nothing checked — can run in Surfaces. Added `remote:check:compose` (a `--compose` flag on the existing script) and one step. The flag exists so Surfaces doesn't re-run the workspace typecheck and tests it already runs as separate steps, and so the twelve env defaults those files interpolate stay in the script instead of being copied into the workflow. `remote:check` is unchanged as the local superset. AGENTS.md and the `ci.yml` comment now state the actual constraint.

**Watched it fail, and watched *both* calls fail** — a second validation that never runs is a step that enforces half of what it claims:
- bogus key in `compose.yaml` → `validating .../remote/compose.yaml: services.acme additional properties 'not_a_real_key' not allowed`, exit 1
- bogus key in `compose.dev.yaml` only → `validating .../remote/compose.dev.yaml: services.remote-api.build additional properties 'bogus_field' not allowed`, exit 1
- unflagged `remote:check` still runs typecheck + 21 tests + both files

**2. No dangling-symlink risk in the final image.** `remote/api` declares no `@openbot/*` dependency and imports none, so stage two — which copies only `remote/api` plus the two `node_modules` trees — needs no workspace directories on disk. The manifests are needed for bun's pruned-checkout validation at install time only, which is exactly what the new test models.

**3. The guard's model matches reality today, and I checked the case where it might not.** It walks the graph from the root package, matching the error bun actually emits. If bun instead requires the deps of *every* workspace present on disk, it would under-cover, since `apps/auth-api` is copied without the root depending on it. But the root requires `{brand, contracts, logging, team-client}` and every workspace in the repo depends only on a subset of those, so the two models are currently indistinguishable — I could not construct a case separating them without editing a manifest, which `--frozen-lockfile` rejects for an unrelated reason. Left the model as is rather than adding a branch I cannot watch fail.

`remote/coturn/Dockerfile` runs no install, and both compose files build from the one Dockerfile already fixed — nothing else to change there.

**Still not covered, and deliberately not done here:** nothing builds the image. `docker build -f remote/api/Dockerfile .` in CI would have caught this class outright rather than by proxy, at maybe 1–2 min a run. That's a CI-minutes call, so it's yours — happy to add it.

`lint`, `typecheck` (eleven projects), `remote:check` and `dependency-catalog.test.ts` (6/6) all green.

PR #251, opened 2026-09-05
refactor(contracts): split ipc-conversation.ts by domain

NorbertBodziony:
`packages/contracts/src/ipc-conversation.ts` held ~15 unrelated domains in 1906 lines. The problem
was not size but **the distance between an interface and the runtime guard that validates it at the
trust boundary**:

| Type | Guard | Gap |
| --- | --- | --- |
| `AgentSummary:164` | `isAgentSummary:1637` | 1473 lines |
| `SidebarLayoutSnapshot:463` | `isSidebarLayoutSnapshot:1876` | 1413 lines |
| `AttachmentSummary:600` | `isAttachmentSummary:1730` | 1130 lines |
| `QueueSnapshot:685` | `isQueueSnapshot:1675` | 990 lines |

That distance was not sloppiness. Lines 1471–1896 were one 425-line block serving `isAgentEvent` — a
24-variant switch reaching into ~10 domains — so each guard sat beside its *caller* rather than its
*type*.

## The split

Fourteen domain modules, plus a package-internal `ipc-bounded-values.ts` for the `INPUT_LIMITS`-aware
primitives. Those deliberately do **not** go in `runtime-values.ts`, which is limit-free and imported
by renderer, backend and preload alike.

Imports flow strictly downward. `ipc-agent-events.ts` is the only fan-in module and nothing imports
it back, so no cycle is possible — `isAgentEvent` imports the guards it switches on from the eleven
modules below it rather than restating their rules.

## This is a pure move

`ipc-conversation.ts` was never in the package `exports` map. All 355 importers go through the
`@openbot/contracts/ipc` barrel, and only 8 files referenced the module by path — all relative, all
inside the package. **No call-site churn outside `packages/contracts`.**

Verified mechanically against `origin/main` rather than by eye: no exported name is missing, none is
duplicated, and the public surface grows by exactly **five guards** that now have to cross a module
line — `isAgentApproval`, `isAgentPromptQuestion`, `isAgentRuntimeSnapshot`, `isBrowserTakeoverRequest`,
`isConversationSnapshot`. The six bounded-value helpers stay package-internal.
Nothing else is added, removed or renamed. `ipc-dynamic-island.ts` and `ipc-team-host.ts` appear in
the diff for a one-line import repoint each and nothing more.

### Two things left alone on purpose

`isAgentApprovalPermissions` and `isRuntimeApprovalPermissions` are two guards over one type with
different path limits; they land in `ipc-approvals.ts` and `ipc-agent-runtime.ts` respectively. Same
for `isPathList` / `isRuntimePathList`. `src/backend/agent-store.ts:1108` has an unrelated third
`isMarketplaceSource`; untouched.

## Verification

| Check | Result |
| --- | --- |
| `bun run test:contracts` | 126 tests / 14 files green — `ipc.test.ts` **unedited** |
| `bun run typecheck` | ten projects green, incl. renderer, mobile, storybook |
| `bun run lint` | 0 errors (5 remaining warnings are pre-existing `apps/mobile` module-mocks) |
| `bun run test:desktop -- src/main/ipc-channel-coverage.test.ts` | 15 green — no guard leaked into preload |

`ipc.test.ts` (742 lines, 40 cases over `isAgentEvent`, `isAgentSummary`, `isQueueSnapshot`,
`isAttachmentSummary`, the conversation event markers) imports the barrel, so **green there without
editing the test is the verification a pure move needs**: every name still resolves and every guard
still behaves identically.

## What this does not fix

A `value is T` predicate need not check every field, and `tsc` cannot see that it skipped one. Adding
a field to an interface and forgetting the guard still compiles, and the field is still silently
dropped at the renderer-to-main boundary. The split makes that easier to notice by putting the guard
beside its type; it does not prevent it.

An earlier revision of this PR added a `satisfies Record<keyof Required<T>, true>` field manifest
above each guard. It was dropped: a manifest is satisfied by adding one line to a key list, so it
converts a silent omission into a mechanical box-tick rather than into validation — and a filled-in
manifest *looks* like coverage, which is worse than no mechanism at all. That cost ~293 lines and 51
public barrel exports for a prompt, not a guarantee.

The mechanism that would actually close it is a completeness test: take a valid sample, delete one
key at a time, assert the guard rejects. `ipc.test.ts:637` already does this for a single key of
`AgentSummary`. That cannot be satisfied by editing a list — the only way to green is real
validation. Generalizing it across the guarded types is worth its own PR.

## Surfaces touched

`packages/contracts` only. No renderer, mobile, web, migration or `mock-openbot.ts` change — the mock
consumes `OpenBotDesktopApi` from `ipc-desktop-apis.ts`, whose import list is repointed but whose
exported types are byte-identical. **No wire protocol change**: `team-protocol/v1-adapter.ts` and
`v2-adapter.ts` import the same `AgentEvent` and `isAgentEvent`, only from a new path.

No `biome-ignore`, `@ts-expect-error`, `@ts-ignore`, `biome.json` override, or widening to
`any`/`unknown`.


NorbertBodziony:
The finding's facts are all correct — 51 manifests, 51 exported aliases, ~293 lines, and no proof
that any guard checks the field it just forced you to list. I'm not disputing any of that. But I
think it measures the mechanism against a goal the mechanism explicitly disclaims, and I'd like the
P3 reconsidered on that basis.

**What it enforces is narrower than "guard completeness."** The claim is only that the declared key
list cannot drift from the interface. That is a real, checked property, and it is the one the
`satisfies` gives you in both directions — verified on this branch by breaking it twice:
`ipc-team-host.ts(288,3)` TS2741 after adding a field to the *base* `TeamMemberSummary`, and
`ipc-dynamic-island.ts(396,3)` TS2353 after adding a key the interface never declared. Both errors
land on the manifest line, directly above the guard body.

**"No consumers" is the intended shape, not an oversight.** The exported `…Field` alias exists
because `noUnusedLocals` reports an unread module-scope `const` as TS6133. It is a type alias, so it
erases at build; the object is built once per module at import and never per validation. That was the
deliberate trade against the alternative — a `{ id: isString, agentId: isIdentifier }` validator
table, which *would* prove guard completeness but walks an object on every IPC message.

**Why a prompt is worth the weight here.** The split already puts each guard beside its type, which
helps a reader who is in the guard file. It does nothing for the person editing the interface, who
adds a field and never scrolls to the guard — and that is who causes this bug. The failure is silent
(the guard approves an object missing the field, downstream reads `undefined` typed as `boolean`) and
it is on the renderer-to-main trust boundary. Converting a silent failure into a compile error at the
exact moment of the edit is the value; it is not proof, and the PR body and
`packages/contracts/AGENTS.md` both say so rather than overclaiming.

I also tested the type-only formulation that would drop the 51 runtime objects
(`type Exact<T, K extends Record<keyof Required<T>, true>> = K`). It catches the missing-key
direction but not the stale-key direction — a type-argument constraint is not an excess-property
check — and the workaround surfaces the error at a contrived use site with a message naming the
wrong key. `satisfies` is the better form.

/norbiai review

NorbertBodziony:
Accepted — the manifests are gone. The branch is rebuilt as a single commit `048c18e5` containing the domain split alone.

What settled it wasn't the enforcement gap on its own, but who the mechanism was aimed at. The manifest's whole value was "it makes you go look at the guard," and that premise is weak when most edits here are made by an agent: a compile error is something to clear, not to reflect on. A filled-in manifest then reads as coverage while proving nothing — the same failure mode that got `no-runtime-typeof` deleted for teaching readers to skim its warning class.

Removed: 51 manifests, 51 `…Field` aliases, 15 convention comment blocks, the `packages/contracts/AGENTS.md` section, and the two interfaces extracted only to carry a manifest (`AgentMarketplaceSource`, `DynamicIslandApprovalDetail`) — both restored byte-faithfully from `origin/main`. `AGENTS.md` is now identical to main.

What's left is a pure move: no exported name missing, none duplicated, and the public surface grows by exactly six guards that have to cross a module line. `ipc.test.ts` is unedited and green.

The completeness gap stays open, and the mechanism that would close it is a test, not a type: valid sample, delete one key at a time, assert the guard rejects. `ipc.test.ts:637` already does that for one key of `AgentSummary`. Generalizing it is a follow-up PR.

NorbertBodziony:
Took the `isQueueDelivery` half of this. Its only caller is `ipc-queue.ts:79`, 44 lines below it in the same module — nothing crosses a module line, so the `export` was a slip in the mechanical split rather than a cost of it. Removed in `599ff59f`; ten typecheck projects still green, which is the proof nothing outside was reading it. The public surface now grows by five guards, not six.

Leaving the other five exported. Each has a real cross-module caller — `isAgentEvent` in `ipc-agent-events.ts` switches on all of them, and `isBrowserTakeoverRequest` and `isConversationSnapshot` have second callers in `ipc-agent-runtime.ts` and `ipc-conversations.ts`. The suggested alternative — an explicit allowlist in `ipc.ts` re-exporting only the pre-PR names — would mean the barrel stops being `export *` and becomes a hand-maintained list of ~400 symbols that drifts silently the first time someone adds a type. That is a larger and more failure-prone surface than the five guards it hides, and it is exactly the class of hand-maintained key list this PR just deleted 293 lines of. These five are guards over already-public types; exporting them leaks no new shape.

PR #256, opened 2026-09-06
refactor(ipc): let the compiler own the main-process channel mirror

NorbertBodziony:
## What this is

`packages/contracts/src/ipc-channels.ts` declares 191 channels, and `packages/contracts/AGENTS.md`
described three files that mirror the list by hand. Only one of the three was actually unchecked.

The mock (`mock-openbot.ts`) and the preload bridge are both annotated `: OpenBotDesktopApi`, so
`tsc` already links them in both directions. The **main process** was the real gap: it registered a
handler by naming a channel string, and nothing said the sequence was complete. A channel declared
and never registered compiled, shipped, and surfaced as `No handler registered for
'agent:list-models'` the first time a user opened the feature. A 565-line static scan of the source
text was what noticed — a test that had to re-implement enough of the language to find the calls.

This replaces that scan with the type checker.

## How

**`packages/contracts/src/ipc-endpoints.ts`** (new) gives the flat channel list structure: which
group each channel belongs to, and whether it is a `request` the renderer invokes or an `event` the
main process sends. 20 groups, one per registrar; no wire value changes, and the file reads its
values from `IPC_CHANNELS` so a typo is already an error at the reference.

A type-level assertion holds the two files together at zero runtime cost:

```
error TS2344: Type '{ channelsMissingFromEveryGroup: "app:brand-new-thing"; }'
  does not satisfy the constraint 'true'.
```

**`src/main/ipc/define-ipc-group.ts`** (new) turns a registrar from a sequence of calls into a value.
Each of the sixteen registrars now *returns* `Pick<IpcGroupHandlers, …>` instead of registering, and
`src/main/index.ts` spreads all sixteen into one `registerIpcGroups` call. Neither entry point takes a group object — both are handed a group *name* and read the endpoints out of
`IPC_ENDPOINTS`, so no channel outside the manifest can reach a registration. `GroupHandlers<G>` is a
mapped type over the group's request endpoints, so:

| Mistake | Diagnostic |
| --- | --- |
| endpoint with no handler | `TS2741: Property 'prepareModel' is missing … but required in type 'GroupHandlers<…>'` |
| handler for an endpoint that does not exist | `TS2561: 'prepareTheModel' does not exist … Did you mean 'prepareModel'?` |
| group with no registrar | `TS2741: Property 'telemetry' is missing … but required in type 'IpcGroupHandlers'` (at `index.ts`) |
| channel in no group | `TS2344` above |

Four constructors bind a handler — `handler`, `payloadHandler`, `eventHandler`,
`authorizedHandler`. Each is a closure that calls `handleTrusted` /
`handleTrustedWithEvent` with its *real* decoder and payload type, so the rule that a payload cannot
be bound without a decoder for it survives intact — there is no constructor that pairs one with a raw
`unknown`. No assertion past the checker anywhere in the change.

**`src/main/ipc-channel-coverage.test.ts`** loses four assertions and the whole main-side scan. What
is left is what no type reaches: the preload (its API object is shaped for the renderer — nested,
renamed, decoding results — so nothing pairs a method with an endpoint) and the event senders
(ordinary calls wherever the event happens). Two assertions are new. One reads the manifest rather
than source text: no channel may be named by two groups, which `ipcMain.handle` would crash on at
launch and set equality says nothing about. The other keeps the registration primitives reachable
from one module — `handleTrusted` and `handleTrustedWithEvent` take a `string` channel, because
`ipcMain` does, so a direct call anywhere else would be a privileged handler on a channel no group
declares. It carries a positive control, since a scan whose pattern goes blind is green either way.

## Trust boundary

`src/main/trusted-ipc.ts` is unchanged except for exporting four types it already had.
`define-ipc-group.ts` is now its only caller, and the coverage test is what holds that. The existing "no `ipcMain` outside `trusted-ipc.ts`"
and "every registration inside it names the sender check" assertions are untouched and still pass,
and `src/main/ipc/define-ipc-group.test.ts` (new, 2 tests) asserts behaviourally that a handler bound
through the new indirection still rejects an untrusted renderer and lands on the channel its endpoint
declares.

`authorizedHandler` preserves the ordering `handleTrustedWithEvent` already had: sender identity is
settled before anything the caller sent is decoded.

## Conversion check

The one mistake neither the types nor the tests can see is an implementation bound to the wrong
endpoint key — that compiles. So every registration was mapped back to `b819884f`: each old
`handleTrusted*(IPC_CHANNELS.x, …)` call against the channel its new binding resolves to through
`IPC_ENDPOINTS`.

173 registrations before, 173 bindings after, the two channel sets identical, and 162 of the
implementations byte-identical once whitespace is normalized. Of the 11 that differ, 5 are a trailing
comma from reformatting. The other 6 are `dynamic-island-handlers.ts`, which was rewritten by hand to
hoist `fromMainRenderer` / `fromOverlayRenderer`; each of its seven endpoints keeps the renderer-id
set and the rejection message it had before — `setPreference` and `publishPresentation` on
`mainRendererIds`, `getPresentation`, `performAction`, `performHaptic` and `setInteractive` on
`overlayRendererIds`, and `getPreference` on the union, inline as before.

## Surfaces touched

- **IPC contracts** (`packages/contracts`) — one new file, one barrel export. No wire value changed.
- **Main process** (`src/main`) — 16 registrars renamed (`register-x-handlers.ts` → `x-handlers.ts`,
  `registerXIpcHandlers` → `xIpcHandlers`) and converted; `index.ts` dispatcher; coverage test.
- **Documentation** — `packages/contracts/AGENTS.md`, `src/main/AGENTS.md`, and a new
  `src/main/ipc/AGENTS.md` with the four constructors and a table of what enforces what.

Not touched: the renderer, the preload, `mock-openbot.ts`, mobile, sites, remote, the database.

## Not in scope

The preload is still a hand-written mirror held by a source scan. Closing it by type would mean
reshaping `OpenBotDesktopApi` around `IPC_ENDPOINTS`, which moves every renderer call site — a much
larger change than this one, and a separate decision.

## Checks

`bun run lint` (5 warnings, all pre-existing in `apps/mobile`), `bun run typecheck` (all 11
projects), and `bun run test:desktop` on `ipc-channel-coverage`, `define-ipc-group`, `trusted-ipc`,
`input-parsers` and `dynamic-island-window` — 67 passed.

### Watch it fail

Every new and changed assertion was broken first and confirmed red for the right reason:

| Break | Result |
| --- | --- |
| endpoint dropped from a group | `TS2741` naming `prepareModel` |
| stray key added to a group | `TS2561` naming `prepareTheModel` |
| group added with no registrar | `TS2741` at `index.ts` naming `telemetry` |
| channel added to `IPC_CHANNELS`, in no group | `TS2344`, `channelsMissingFromEveryGroup: "app:brand-new-thing"` |
| preload stops invoking a request channel | *invokes exactly the request endpoints* → 172 vs 173 |
| endpoint kind flipped request → event | all three set assertions red |
| the only sender of an event channel removed | *sends exactly the event endpoints* → `authEvent` missing |
| one channel named by two groups | *declares each channel in exactly one endpoint group* → 1 finding |
| binder stops skipping event endpoints | *registers each request endpoint …* red |
| binder registers on the endpoint name, not its channel | both binder tests red |
| sender check deleted from `handleTrusted` | *keeps a bound handler behind the sender check* red |

Model: `claude-opus-5`, harness: Claude Code.


NorbertBodziony:
Accepted — `eventPayloadHandler` had no caller. Removed in 41342328, along with its row in `src/main/ipc/AGENTS.md`; the section is now "The four ways to bind a handler" and says why the unauthorized event-plus-payload form is deliberately absent (every endpoint needing both needs the sender check, and `handleTrustedWithEvent` still carries the overload for when one does not).

/norbiai review

NorbertBodziony:
Half of this is right, half is not — 1279b41a takes the half that is.

`GroupedChannel` and `DeclaredChannel` are now private. Nothing outside the file names them, and `ChannelCoverage` references both, so `noUnusedLocals` is satisfied.

`ChannelCoverage` has to stay exported, and the compiler is what says so. Unexporting it is `src/ipc-endpoints.ts(294,6): error TS6196: 'ChannelCoverage' is declared but never used.` — `tsconfig.base.json` sets `noUnusedLocals: true`, and a type-level assertion is by construction referenced by nothing.

Folding it into `IpcEndpoints` does not work either. `GroupedChannel` maps over `keyof IpcEndpoints`, so `IpcEndpoints = ChannelCoverage extends true ? … : …` is circular; breaking the cycle by mapping over `typeof IPC_ENDPOINTS` instead leaves the conditional resolving to the failure object when coverage breaks, so `IpcEndpoints` becomes `never` or `{ channelsMissingFromEveryGroup: … }` and every group, registrar and `registerIpcGroups` call in `src/main` errors at once. Today the same break is exactly one diagnostic, which I re-checked by deleting `transcribe: request(IPC_CHANNELS.voiceTranscribe)` from the `voice` group:

```
src/ipc-endpoints.ts(297,43): error TS2344: Type '{ channelsMissingFromEveryGroup: "voice:transcribe"; }' does not satisfy the constraint 'true'.
```

The export now carries a comment saying it is an assertion, not API, and that nothing should import it.

/norbiai review

NorbertBodziony:
Fixed properly in beb9ada6 — the alias is gone rather than hidden, so there is nothing coverage-only on the package surface and no export list to keep in step with the module.

```ts
type CoveredEndpoints<Coverage extends true> = typeof IPC_ENDPOINTS & Record<never, Coverage>;

export type IpcEndpoints = CoveredEndpoints<SameChannels<GroupedChannel, DeclaredChannel>>;
```

The constraint is the assertion, checked where the argument is written. `Record<never, Coverage>` is `{}`, so `IpcEndpoints` is the manifest and nothing else — including when coverage fails, which is what keeps the break to one diagnostic instead of every registrar in `src/main` erroring against a poisoned `IpcEndpoints`. Re-checked by deleting `transcribe: request(IPC_CHANNELS.voiceTranscribe)` from the `voice` group: `ipc-endpoints.ts(298,45): error TS2344: Type '{ channelsMissingFromEveryGroup: "voice:transcribe"; }' does not satisfy the constraint 'true'.`, plus the expected `TS2353` on the now-extra `voice-handlers.ts` key.

`export *` stays. Replacing it with a hand-listed set of names is the mirror this PR exists to delete, and it would be the one file in the `ipc.ts` barrel that works differently.

`GroupedChannel`, `DeclaredChannel`, `SameChannels` and `CoveredEndpoints` are all private now; the two AGENTS.md rows that named `ChannelCoverage` name the `IpcEndpoints` declaration instead (03d8e2a3).

/norbiai review

NorbertBodziony:
Valid — fixed in 8dceb804. `registerIpcGroup` takes a group *name* now and reads the group out of `IPC_ENDPOINTS` itself, exactly as `registerIpcGroups` already did:

```ts
export function registerIpcGroup<Name extends keyof IpcEndpoints>(
  name: Name,
  handlers: GroupHandlers<IpcEndpoints[Name]>,
): void;
```

So there is no longer a parameter through which a caller-made group reaches `ipcMain.handle` — from either entry point, the channel a handler binds to comes from the manifest and nowhere else. `src/main/ipc/AGENTS.md` gains that row in the enforcement table.

The test drives the real `voice` group instead of a synthetic `test:*` one — it is the smallest group with all three shapes (a request taking nothing, a request taking a payload, an event). Watch-it-fail, three ways:

- bind events too (drop the `kind !== "request"` guard) → both tests red
- bind to the endpoint's key instead of its channel → `expected undefined to be 'status'`
- drop the sender check in `handleTrusted` → `expected [Function] to throw an error`

/norbiai review

NorbertBodziony:
Valid — closed in d47dd6ab. `ipc-channel-coverage.test.ts` now fails on the *name* `handleTrusted` or `handleTrustedWithEvent` anywhere in `src/main` outside `ipc/define-ipc-group.ts` and the wrapper module itself, which is the same shape as the `ipcMain` assertion beside it: a name has to appear to be imported, called or aliased at all.

It reads names rather than the import path because `readSource` strips string literals — the first version matched `from "…/trusted-ipc"` and was silently green against a real violation. The positive control caught it and stayed in: `expect(TRUSTED_WRAPPERS.test(readSource(BINDER_MODULE))).toBe(true)`, so a pattern that goes blind fails instead of passing.

Watch-it-fail, both directions:

- add `handleTrusted("voice:undeclared", () => null)` to `ipc/voice-handlers.ts` → `expected [ Array(1) ] to deeply equal []`, naming the file
- rename the pattern to something no source has → `expected false to be true` on the fixture line

`src/main/AGENTS.md` and `src/main/ipc/AGENTS.md` say what enforces it (baf3067a).

/norbiai review

PR #258, opened 2026-09-06
refactor(main): extract a composition root from index.ts

NorbertBodziony:
## What

`src/main/index.ts` was 1464 lines and the highest-churn file in the repo. Inside it,
`app.whenReady().then(async () => …)` spanned ~500 lines as an anonymous callback that built
20+ services and assigned them into **27 module-scope `let`s**.

Because construction had no named ordering, anything wired *before* the service it reads had to
take a **function** instead of a value — the footgun `src/main/AGENTS.md` already documents, since
`() => X | null` and `X | null` both type-check at every call site. Two concrete costs: **23
unreachable guards** throwing `"The account service is not ready."` for a manager assigned exactly
once, and a shutdown sequence of thirteen `?.`-guarded lines in a hand-maintained order that
nothing linked to construction.

The callback moves **whole** into `createApplicationServices()` in a new file. Not "split into
stages" — one function, because `let x: T | null` assigned and then passed to a non-null parameter
compiles only through control-flow narrowing inside a single function body. A split would break
every one of those; a single function makes the hazard disappear rather than mitigating it.

| File | Lines | Role |
| --- | --- | --- |
| `application-services.ts` | 560 | new — the composition root, construct-only |
| `main-window.ts` | 402 | new — every `BrowserWindow`, renderer URLs, menu, bounds recorder |
| `development-remote-bootstrap.ts` | 138 | new — the dev-only `OPENBOT_DEV_REMOTE_ROLE` account and connection |
| `teardown-registry.ts` | 79 | new — the only genuinely new logic |
| `teardown-registry.test.ts` | 57 | new — 4 cases |
| `session-configuration.ts` | +50 | gained `configureApplicationProtocol`; `remoteServers` is now a value |
| `index.ts` | **1464 → 595** | dispatcher, lifecycle, deep links, shutdown |

Module-scope `let`s: **27 → 11**. (The plan predicted 13; `mainWindow`/`mainWindowLoad` became
fields of a `const` holder instead.)

`registerIpcHandlers` and all 16 registrar calls **stay in `index.ts`**, so `DISPATCHER_MODULE` in
`ipc-channel-coverage.test.ts` and its assertion are untouched. No existing test file changes.

## Two deliberate behaviour changes

Both are in `teardown-registry.ts` and both are why the file exists:

1. **Every step is caught individually.** `prepareForShutdown` had no `try`, so a rejecting
   `remoteServers.stop()` skipped the host, the WebRTC bridge and the agent service below it —
   and `.finally(() => app.quit())` quit anyway, leaving provider CLIs as orphans.
2. **A step registered after `runAll` still runs.** `before-quit` does not abort the `whenReady`
   callback (pre-existing), so construction keeps going during teardown. A service built in that
   window would otherwise leak a listening socket or a child process; its step is chained onto the
   run in progress instead of dropped.

Shutdown order is **preserved byte-for-byte**. Steps carry an explicit ordinal rather than running
LIFO, because today's order is largely *construction* order and not its reverse: the browser host
is destroyed before the picture-in-picture window holding it, and the provider runtimes stop before
the agent service that owns them. Reversing it would move the browser's cookie and state flush
behind a potentially slow agent stop — a real data-loss window, not taken here.

## Cleanups this unlocked

- 21 `throw "The account service is not ready."` + 2 null-ish fallbacks → direct arrow calls. The
  `TeamWebRtcClientTransport` literal drops from 53 lines to ~17.
- `getRemoteServerManager` is the one accessor that converts to a value; both `!remoteServers`
  branches in `session-configuration.ts` and their doc comment are deleted.
- `browserAlreadyDestroyed` / `destroyBrowserForShutdown` deleted — `BrowserHost.destroy()` is
  memoized. Verified idempotent likewise: `UpdateService.stop()`,
  `BrowserPictureInPicture.destroy()`, `DynamicIslandWindowController.destroy()`,
  `MacHapticFeedback.destroy()`.
- `IpcHandlerDependencies` deleted; `registerIpcHandlers` takes `ApplicationServices` directly.
  After the merge with `main` it does so through #256's `registerIpcGroups`, whose group coverage
  is compiler-checked - so that half of the conflict resolution is verified by `tsc`, not by eye.

Accessors that **stay** functions are the genuinely lifetime-nullable ones, each with a comment
saying so: `getMainWindow`, `takePendingInvite`, `resolveOwner`/`resolveAgent`, `getLocalHostId`,
`getSignedInUser`, the three `getRemoteDesktop*` options.

## Surfaces touched

Desktop main process only. No renderer, mobile, web, `remote/api`, IPC contract, migration or
schema change — hence no `mock-openbot.ts` counterpart. Docs updated: `src/main/AGENTS.md` (Size
table, the function-versus-value paragraph) and
`.agents/skills/release-upgrade-safety/references/surfaces.md`, which named `index.ts` as the home
of the file-name constants.

## Verification

`bun run lint` (5 warnings, all pre-existing `no-module-mocking` in `apps/mobile`) and
`bun run typecheck` (11/11 projects) are green. Targeted tests:

```
teardown-registry     4 passed
ipc-channel-coverage 14 passed
host-service         12 passed
main-window-state    11 passed
```

**Watch-it-fail**: all four `teardown-registry` cases were broken and confirmed red for their own
reason before restoring. One caveat worth stating rather than hiding — *"runs each step once when
shutdown is requested twice"* only goes red when **both** the `#closed` memo guard and
`this.#steps = []` are removed. It protects the property, not a single line. Reshaping it to fail
on one line would make it test less.

`createApplicationServices()` **cannot be credibly unit-tested**: it would need a mocked `electron`
plus 20 real services opening SQLite, spawning processes and binding listeners. The test would
assert its own mocks. `tsc` covers the type-level half; the ordering half has no automated check,
so the manual gate below is the real one.

### Manual gate, run against a real dev instance

- **Cold start, restored session** (`OPENBOT_DEV_REMOTE_ROLE=host`) — account, local host server,
  agents, attachments and the Dynamic Island surface all up; zero errors in the main log.
- **Close all windows → `activate` rebuild** — new window target id, fully functional. This is the
  check for the stale-window hazard and the single lazy `getServices()` handle that replaced eleven.
- **Quit after startup** — exit code 0 and the `grok`/`codex` provider processes died. They are
  reparented rather than killed by a dying parent, so their death *is* the evidence the registry's
  provider step ran.
- **Quit during startup** — **no orphaned providers**, which is exactly what the late-push rule was
  added for. The process itself stayed alive on `dialog.showErrorBox`; that `.catch` block is
  byte-identical to `main`, so it is pre-existing and out of scope.
- **`OPENBOT_DEV_REMOTE_ROLE=client`** — connected to the host and issued remote calls.

The client run logged two `RemoteProtocolError`s (`agent:list-models`, `browser:set-visible`). Not
assumed pre-existing — the same scenario was rerun on `33bd4971`: **2 errors on base, 2 on this
branch, same handlers, same cause** (`decodeAgentModelOptions`, `ensureCompatibility` — neither
touched here).

## Known limits

- ~~**Risk 2 is narrowed, not eliminated.**~~ **Fixed in `0bbc4bfa`** after review flagged it P1.
  `createApplicationServices` returns `appliedAccount` — the state it read and bound the local host
  to — and the entry point adopts it as the active analytics/remote principal, then replays the
  current state through `forwardCentralAuth` when status or principal id moved on during
  construction. Conditional on purpose: an unconditional replay would repeat `syncRemoteHosts()` on
  every signed-in cold start, since `remoteServers.initialize()` already syncs. Not fixed by moving
  the `on("changed")` subscription later, which would drop the `centralAuthGeneration` bookkeeping.
- **A teardown step registered after `runAll` has returned is unawaited** (review P2, accepted).
  Ordinal placement and awaiting during an active drain are fixed; closing the last gap needs
  construction to abort on `before-quit`, which is the separate commit below.
- Twelve services are never torn down at all today (`store`, `teamStore`, `mailbox`, `analytics`,
  `hostedSites`, …). Registry pushes go only where a teardown call already existed; adding one is
  out of scope.
- Adding a `shutdownStarted` check at construction boundaries so `before-quit` can abort the
  callback is a separate commit.

## Approvability

Auto-approvable: no `biome-ignore`, no `@ts-expect-error`/`@ts-ignore`, no `biome.json` overrides
or unregistered plugins, no widening to `any`/`unknown`, no assertion past a checker.

Model: Claude Opus 5. Harness: Claude Code CLI.


NorbertBodziony:
Both findings are real. Fixed in `0bbc4bfa`, then merged with `main` in `b224d795`.

**P1 — account changes lost during construction.** Correct, and it is the plan's Risk 2 widened rather than narrowed: `forwardCentralAuth` reaches the host, the remote servers and analytics only through `services`, so a sign-in settling while `remoteServers.initialize()` awaits was dropped with nothing to replay it.

Took the first of the two suggested remedies. `createApplicationServices` now returns `appliedAccount` — the state it read at `application-services.ts:414` and bound the local host to. After `services = built`, the entry point adopts that account as `activeAnalyticsPrincipalId` / `activeRemotePrincipalId` (construction attributed the flushed analytics to it, so it *is* the active principal, and those two decide whether the next account change clears analytics and disconnects the previous sessions), then replays the current state through `forwardCentralAuth` when status or principal id has moved on since.

Deliberately conditional. An unconditional replay would repeat `syncRemoteHosts()` on every signed-in cold start — `remoteServers.initialize()` already calls `#syncWebRtcHosts()` — which is a behaviour change this PR should not carry. Not fixed by moving the `on("changed")` subscription later, which would drop the `centralAuthGeneration` bookkeeping.

**P2 — late teardown steps.** Both sub-claims were right, and the second contradicted the module doc, so the doc was not the thing to fix.

Did not take the seal signal: `before-quit` awaiting construction would block a quit behind a startup that can poll for 30s in `OPENBOT_DEV_REMOTE_ROLE=client`. Instead the drain now holds its remaining steps in `#pending`, and a `push` arriving while it runs is spliced into that array at its ordinal position rather than appended to the tail. That fixes the ordering claim, and because the drain `shift`s from the same array, such a step is now inside the promise `before-quit` awaits — closing the first sub-claim for the window that matters.

What remains genuinely best-effort is a push arriving after `runAll` has returned. Closing that needs construction itself to abort on `before-quit`, which it does not do today (pre-existing, and named as a separate commit in the plan). The module doc now says exactly this instead of claiming a guarantee it cannot keep.

`teardown-registry.test.ts` covers the ordinal placement — the existing late-push case was widened rather than a new one added. Watched it fail first: with `push` restored to tail-chaining it reports `[browser, agents, remote servers]` against the expected `[browser, remote servers, agents]`.

**Merge with `main`.** `#256` landed the compiler-checked `registerIpcGroups` mirror and renamed `ipc/register-*-handlers.ts`. Resolved to main's shape and the new module names, keeping this branch's `ApplicationServices` parameter, `windowHolder`-backed `getMainWindow` and non-optional `analytics`. The service imports sitting beside the ipc ones on main stay dropped — they live in `application-services.ts` now. `registerIpcGroups` type-checks the group coverage, so the resolution is compiler-verified.

Local: `lint` clean (5 pre-existing `apps/mobile` warnings), `typecheck` 11/11, and `teardown-registry`, `ipc-channel-coverage`, `ipc/define-ipc-group`, `host-service`, `main-window-state` all pass.

/norbiai review

NorbertBodziony:
Agreed on the remaining P2, and not disputing it — no `/norbiai review` here.

A push arriving after `runAll` has returned is unawaited, and neither remedy fits this PR:

- **Sealing** makes `before-quit` wait for construction to finish. `OPENBOT_DEV_REMOTE_ROLE=client` polls for the host connection file for 30s before throwing, so a quit during that window would hang for up to 30s.
- **Aborting construction** is the right fix, but it changes the startup failure path, which this PR deliberately leaves alone. The plan scoped it to a separate commit for that reason.

Recording the shape so the follow-up does not have to rediscover it: `teardown.push` is already the abort boundary. It is called immediately after every service that owns a resource is created, so it needs no new stage boundaries in what is deliberately one function — a `push` on a closed registry can tear the just-built service down and abort the rest of construction from there. The catch to get right is `createApplicationServices` rejecting: the entry point currently surfaces a startup failure with `dialog.showErrorBox`, and a quit-during-startup must not raise a modal on the way out. That distinction is the actual work, and it is why it is not bundled here.

Until then the module doc states the limit rather than claiming the guarantee.

All eight checks green on `b224d795`.

NorbertBodziony:
Smoke run on the merged head `b224d795` — the plan's manual list, which is the real gate here since `createApplicationServices` cannot be credibly unit-tested. Driven against a dev instance with `dev:automation` and CDP.

| Check | Result |
| --- | --- |
| Cold start, restored session | Both windows built. Renderer fully loaded: server list, chat list, conversation, attachments, routines. Signed in as `openbot-dev-host`, local host bound and `[pressed]`, email redacted in the a11y tree. |
| Cmd+Q after startup | Apple Event → `before-quit`. App `stopped with code 0`, no teardown errors, instance registry cleaned. |
| Orphans after quit | Zero. No surviving child of the main process, nothing left under the worktree. |
| Close all windows → dock activate | `window-all-closed` correctly did not quit. `activate` rebuilt the main window under a **new** CDP target id, and the rebuilt window snapshots with the host still bound — so the holder nulled and `ensureMainWindow` rebuilt, and IPC reaches the new window. |
| Quit **during** startup | Quit fired 471 ms after the Electron main process appeared, inside construction. Clean exit code 0, no hang, no error modal, zero orphans. |

The two `mach_port_rendezvous` ERROR lines in that last run come from a Chromium helper noticing its parent exited first (`No rendezvous client, terminating process (parent died?)`) and self-terminating — a consequence of quitting mid-launch, not an app teardown failure.

**Honest gaps.** The cold start ran against a signed-in profile, so `appliedAccount` was `signed_in` and the reconcile took the no-change branch — the *racing* branch, where the account settles after construction reads it, is still not deterministically reproducible and remains unexercised. Signed-out cold start and the two `OPENBOT_DEV_REMOTE_ROLE` runs were done pre-merge and not repeated; #256 touched IPC registration only, not those paths.

PR #259, opened 2026-09-06
refactor: the four maintainability findings left after #255–#258

NorbertBodziony:
Round two of the maintainability audit. #255–#258 landed the first two findings; these are the four that were left, one commit each.

## What is in here

**`783408d0` — index the two missing directory guides.** `AGENTS.md` promised five directory guides and the repo has seven: `src/main/ipc/AGENTS.md` and `apps/mobile/AGENTS.md` were not in the table, so nothing pointed an agent at them before it edited an IPC handler or the Expo app. Also corrects "the one channel list with its three mirrors" — since #256 the compiler owns one of them, so it is a manifest and two mirrors.

**`06faf699` — delete the unreachable CSS and add the scan that finds it.** 194 rules across 7 stylesheets named classes that appear in no markup anywhere in the repo, plus 69 selectors inside surviving rules and 6 `:is()` groups. CSS is the one renderer surface with no compiler, so this is the class of dead code that only grows. `scripts/ui-foundation-check.ts` now reports every declared class that is not reachable from any `.ts`/`.tsx`/`.html` under `src` or `apps`, understanding template-literal class families (`agent-row-agent-status-${kind}`) so a generated name is not a false positive. It runs in 0.17 s inside the existing `check:ui`.

`conversation.css` 8389→8066 lines, `app-shell.css` 4122→3150, `primitives.css` 2005→1990, `auth.css` 1451→1416, `settings-modal.css` 862→845, `server-settings-modal.css` 490→485, `action-menu.css` 118→107.

**`6107ba1b` — record the scan in `AGENTS.md`** beside the `data-testid` budget it sits next to.

**`6ad2c92a` — collapse the three connect channels into one.** Codex, Claude and Grok each had their own IPC channel, preload binding, desktop API entry, main handler, service method, runtime method and renderer wrapper — seven parallel copies, and every renderer call site carried a `provider === "codex" ? … : provider === "claude" ? … : …` chain. `app:connect-provider` takes the provider id. The CLI half of the sign-in moved to `BuiltInProviderDriver.cliLogin` (argv, env, timeout), the seam that already owns per-provider behaviour; a driver without one signs in over the app-server protocol, which is Codex and only Codex. `ProviderRuntime` keeps one map of pending CLI logins instead of two fields and one map of bundled executables instead of three, so start/complete/fail/cancel exist once each. −138 lines net.

**`9d4bfe16` — fill the three marketplace holes in the preview mock.** `skills`, `hostedSites` and `marketplaceAgents` were the only surfaces `mock-openbot.ts` answered with an empty list or a thrown "unavailable in preview mode", so the marketplace read as permanently empty in Storybook and the preview, and `SkillsMarketplaceModal` — the largest renderer component at 1862 lines — had no story. The mock now serves all three from `preview/fixtures.ts` and holds state, with every reverse present: install/uninstall a skill, publish/replace/delete a site, install a marketplace agent with its skills and `marketplaceSource`. New `SkillsMarketplaceModal.stories.tsx` covers discover, skill detail, installed, my submissions and the agent marketplace at two widths.

## Surfaces touched

Desktop renderer, IPC contracts (`packages/contracts`) and their second implementation `src/renderer/src/preview/mock-openbot.ts`, `src/main` IPC, `src/preload`, `src/backend`, and `AGENTS.md`. No migration, no Team API wire change, no change to what leaves the machine.

## Watch it fail

Done for the new dead-CSS scan, with six deliberate mutations of `ui-foundation-check.ts` against the fixture trees. Five went red for the reason intended. The sixth — dropping `cx(` from the class-region openers — stayed green, so I measured it: all 78 `cx()` calls in the renderer are already inside a `class={…}`, meaning the opener widened no region and enforced nothing. It is deleted, along with its fixture class and the never-matched `className`/`classList=` openers, rather than left as an unexercised branch. `tools/ui-foundation/fixtures/renderer` gained a stylesheet that breaks the check twice beside four correct classes it must leave alone, and `renderer-clean` carries the negative half.

Before deleting any rule I also grepped all 121 dead names across the whole repo independently of the scan; one apparent hit (`file-icon`) was a substring of `file-preview-file-icon` and is genuinely dead.

The changed tests in `provider-runtime.test.ts` and `App.onboarding.test.tsx` are existing tests updated to the single channel, not new ones.

## Checks

`bun run lint`, `bun run typecheck` (all eleven projects) and `bun run check:ui` are green. Targeted: `provider-runtime.test.ts` (8), `agent-service.providers.test.ts` (11), `ipc-channel-coverage.test.ts` (14), `App.onboarding.test.tsx` (15). The new story's five play paths and the mock's install/uninstall, publish/delete and agent-install reverses were verified against throwaway test files that were then deleted — nothing in the repo runs play functions, so this was the only way to check the selectors are real.

No screenshots: the CSS commit has no intended visual change (every deleted rule was unreachable), and the story is the new UI surface itself.

## Approvability

No `biome-ignore`, no `@ts-expect-error`, no `biome.json` override, no widening to `any` or `unknown`, no assertion past a checker. The five standing Biome warnings are the pre-existing module mocks in `apps/mobile/src/features/auth/api/mobile-auth.test.ts`.

One standing disagreement with the reviewer, left as-is for you to decide. It asks the dead-CSS
scan to collect class names only from class-bearing expressions instead of from every
identifier-shaped word in `src` and `apps`, because a stale name in a comment, a test or an `id`
currently keeps its rule out of the report. That false negative is real and deliberate: it leaves a
dead rule in the file, which is the debt the check chips away at, while the narrow scan's failure
mode is telling someone to delete a rule that styles live UI, with no compiler behind the deletion.
Renderer classes are built by `cx()`, template literals, arrays, hoisted constants and props, so the
narrow scan would report live styling as dead. The permissive half of the scan is already restricted
to `class={…}` regions, and `unreachable.css` documents what that restriction costs
(`.identifier-panel`, named by a template-literal `id`). Say the word and I will tighten it.


NorbertBodziony:
Fixed in `62805747`. The finding was right on both halves: the mock always appended a new agent and ignored `input.agentId`, and it recorded invented `routineIds` without creating the routines, so Release Manager installed claiming two managed routines that `listRoutines` did not have.

It now mirrors `AgentMarketplaceService.install` — reuse and update the named agent, refuse one whose `marketplaceSource.listingId` is a different listing, create the listing's routines with real ids in the timezone the caller passed, and drop the routines the previous version installed. `createRoutine` and the install path now share one record builder.

Verified against a throwaway test (deleted): a second install of the same listing updates in place and leaves the agent count and routine count unchanged with fresh ids; installing over an agent from another listing, or over a non-marketplace agent, rejects; and the created routines carry the fixture's `active` flags.

NorbertBodziony:
Both new findings addressed, one of the two halves of the CSS finding pushed back on.

**[P2] Marketplace updates erase user-installed skills — fixed.** `marketplaceAgents.install` now mirrors `AgentMarketplaceService.install`: it drops only the skills the *previous* install of this listing added (`existing.marketplaceSource.skillIds`) that the new version no longer carries, and keeps everything else. Verified with a throwaway test — install a listing, install an unrelated skill onto the same agent, reinstall the listing, assert the unrelated skill survives and the managed one is not duplicated. It fails with `expected [ 'skill-release-notes', … ] to include 'skill-design-audit'` without the fix.

**[P2] Dead-CSS scan — declaration side fixed.** You are right that the declaration regex read `.name` from anywhere in the file. Confirmed empirically against the fixture tree: a `url("panel-sprite.webp")` reported `.webp`, and a comment mentioning `.commented-out-class` reported that, both as unreachable rules to delete — a latent build failure the first time someone adds a sprite or leaves a stale comment. Declarations now come from selector preludes only (comments stripped, then text since the last `{`, `}` or `;`), and both spellings are permanent negative fixtures in `unreachable.css`. Mutating the checker back to the whole-file scan turns `scripts/ui-foundation-check.test.ts` red with exactly those two lines.

**[P2] Dead-CSS scan — usage side, respectfully disagreeing.** Collecting every identifier-shaped word from markup is deliberate, not an oversight, and narrowing it to "class-bearing syntax" would make the check worse. The two error directions are not symmetric: a false negative leaves a dead rule in the file, which is the debt this check exists to chip away at; a false positive tells someone to delete a rule that is styling live UI, and no compiler catches that. Renderer class names are built by `cx()`, template literals, conditional objects, arrays, constants hoisted to module scope, and props threaded through several components — a scan restricted to `class={…}` regions would miss every one of those and report live styling as dead. The prefix-family half of the scan is already restricted to `class={…}` regions precisely because it is the permissive branch, and `unreachable.css` documents the one rule that costs (`.identifier-panel`, named by a template-literal `id`). Excluding tests would also be a behaviour change rather than a tightening: a class named only from a test is still a class the scan should not ask anyone to delete. Happy to be argued out of it if there is a case where the word scan hides a real dead rule that a narrower one would catch.

/norbiai review

PR #260, opened 2026-09-06
refactor(remote): put every client on one shared remote-stack contract

NorbertBodziony:
## Why

OpenBot has a second client of the remote stack (Cloudflare auth API → Signal service → WebRTC to a
host) and a third is planned: the desktop-as-client, the Expo app, and a web client. The question
that started this was whether to extract a shared package for that communication.

The answer is that the package already exists and is already the right shape — `@openbot/contracts`
for the wire, `@openbot/team-client` for behaviour. The work was not creating a package; it was
closing the specific gaps where a client had written the same thing out by hand instead.

The same shape validates independently in `pingdotgg/t3code`, which splits `packages/contracts`
(wire only) from `packages/client-runtime` (connection supervision) with all three apps importing
both and reimplementing nothing.

## What changed

Five commits, each independently shippable.

**`d317b956` — mobile onto the shared Team API route table.** `packages/contracts/src/team-api-routes.ts`
was created to kill bare Team API path literals; its header records that *"until this table every
path was a bare string literal on both of them … Nothing linked the copies, so a rename that missed
one half broke every remote server silently."* `apps/mobile` had zero references to it and
hand-wrote the whole table. It was the un-linked third copy the table exists to prevent.

**`edd544b7` — one Signal wire contract.** `SIGNAL_PROTOCOL_VERSION`, `SignalClientMessage`,
`SignalServerMessage`, `IceServer`, `RemoteTicketClaims` lived in `@openbot/remote-api`, which no
client imports, so four copies existed. The two desktop copies were also *weaker*: both ended in
`.loose()` over a flat bag of optionals, so `sdp`, `candidate` and `resumeToken` were
`string | undefined` on every branch and the compiler could not tell a `ready` from an
`ice-candidate`. Adopting the shared contract is a correctness improvement there, not just DRY.

This also ties `protocolMinimum: 2` — a bare literal in the issuer at
`apps/auth-api/src/server/remote-control-plane.ts` — to the constant the verifier in `remote/api`
checks. Issuer and verifier now share a type.

**`5032b271` — one decoder for the account API's session responses.** `RemoteSession` and
`RemoteSessionTicket` plus their decoders now live in `@openbot/contracts/remote-control-plane`, and
both clients run the same ones. This includes the Signal URL scheme check, which is the last place
that can refuse to put a bearer ticket on the wire in the clear.

**`b683c144` — keep the shared packages browser-safe.** Both are already node-free in non-code
paths, so this locks in a property that holds today rather than demanding a cleanup, and it is what
makes the future web client a drop-in. Plus `team-client`'s root export no longer re-exports all of
`@openbot/contracts/team-protocol`, which dragged every frozen adapter into any consumer of the
root; the three call sites that relied on it now name the subpath they need.

**`b707d6bf` — derive the TURN refresh interval from the service's TTL.** The service minted
credentials with a one-hour TTL and both peers refreshed at forty-five minutes, as three unlinked
literals. Shortening the TTL would have left both clients refreshing on the old schedule with
credentials that had already expired — and a peer only reads TURN when it needs the relay, so the
connection keeps working for everyone except the users behind a symmetric NAT. The interval is now
computed from the TTL rather than restated, so the two cannot drift. No value changes.

## Two things deliberately not merged

**Validation is not shared, on purpose.** `remote/api` keeps its zod schema for the untrusted
*client* input it accepts — byte limits, identifier patterns, a closed discriminated union — and
`signal-protocol/decode.ts` carries the guards a client runs over the *service's* output. One
implementation per trust direction. Neither is the other's mirror and neither should grow into one.
`remote/api`'s existing hardening tests are untouched.

**`@openbot/contracts` stays zero-runtime-dependency.** It is in the graph of a Cloudflare Worker, a
React Native app and an Electron renderer, and `apps/auth-api` has no zod at all — adding zod to
contracts would push it into the Worker's bundle. The new decoders use the existing
`runtime-values` guards in the house style.

The frozen per-version Team API codecs and adapters are untouched. Nothing here collapses a protocol
version.

## The packaging trap this had to fix

Depending on `@openbot/contracts` makes it `remote/api`'s first workspace dependency, and the
Dockerfile was not ready. The install stage was fine, but the **runtime stage never copied
`packages/contracts/src`** — contracts is consumed as raw TypeScript, so the node_modules symlink
resolved into a directory holding only a `package.json`. The image builds, the install succeeds, and
the container **crashes on first import at startup**.

Nothing caught this. No CI job builds the image, and `scripts/dependency-catalog.test.ts` asserted a
*manifest* is copied for each reachable workspace, which contracts already satisfied — it was
written for a different failure (bun refusing the pruned install), which is a build-time error. This
one is a runtime error in the shipped container.

So this PR adds the runtime-stage `COPY` and extends that test with a second assertion: for every
workspace reachable from `@openbot/remote-api`, the runtime stage copies its *source*, not just its
manifest.

## Surfaces touched

Desktop renderer (`src/renderer`), desktop main (`src/main`), mobile (`apps/mobile`), the account
Worker (`apps/auth-api`), the self-hosted Signal service (`remote/api`), and the shared packages
(`packages/contracts`, `packages/team-client`).

No migrations, no schema change, no persisted-state change. **No wire change** — every phase is a
type or module move, with no field added, removed or re-meant, so no new protocol version is needed.
IPC channels are untouched, so `src/renderer/src/preview/mock-openbot.ts` needs no matching edit;
`RemoteDesktopIceServer` became a type alias for the same shape it already declared.

## Watched them fail

Per AGENTS.md rule 2, four times:

1. Neutering the Signal URL scheme check in the shared decoder turned `central-auth-manager.test.ts`
   red on "accepts a private-LAN Signal URL for local Mobile Connect development". The mobile suite
   stayed green — it only asserts the positive case — so the shared rule is covered once, at the
   desktop.
2. Deleting the new `COPY packages/contracts` line turned `scripts/dependency-catalog.test.ts` red
   naming `packages/contracts`, then restored it.
3. `node:crypto` and `node:events` probes in both shared packages turned `bun run lint` red with
   exactly two `noNodejsModules` errors, and a probe in a `.test.ts` beside them correctly did not
   fire.
4. Setting the shared TURN TTL to thirty minutes turned `remote/api/test/tokens.test.ts:77` red on
   the minted credential's username — `1788705189` for the expected `1788706989` — which is the
   assertion that reaches the wire.

## Checks

`bun run lint` (5 pre-existing module-mock warnings, none from this branch), `bun run typecheck`
(11 projects), `bun run check:ui`, `bun run test:remote` (21), `bun run test:api` (158),
`bun run test:desktop -- packages src/main src/renderer/src/features/team` (85 files, 664 tests).
CI owns `check:desktop` and the rest.

### Not run — needs a decision

`bun run build` plus a packaged launch, and `bun run remote:check` plus a container start. Both are
outside the permitted fast set, and the container half is the direct verification for the Dockerfile
fix above. The test added in `edd544b7` is what stands in for it; a real image build would be worth
doing before this merges.

## Approvability

No `biome-ignore`, `@ts-expect-error`, `@ts-ignore`, `any`, or assertion past a checker anywhere in
the diff. The new `biome.json` `overrides` entry **raises** `correctness/noNodejsModules` from its
default `warn` to `error` for the two shared packages — it enables a rule rather than disabling one,
so it is not the escape hatch the approvability rule is about. Precedent is the existing `noConsole`
override, which enables a rule for a path subset the same way.

## Phase B of the plan is deliberately absent

The plan this came from had a sixth piece: migrate the client role of
`src/renderer/src/features/team/team-webrtc.ts` onto `createRemoteTeamPeer`, and unify ed25519 on
`@noble`. I did not do it, and I do not think it should be done as written.

`team-webrtc.ts` is a transport-only bridge worker — raw data channels over a MessagePort, no
knowledge of the app protocol. `createRemoteTeamPeer` is a whole remote-team client: ed25519
handshake, v2/v3 framing, an RPC map, event sequencing, session lifecycle, and no raw-channel egress
at all. Adopting it moves the desktop's protocol stack out of `src/main` and into a sandboxed hidden
`BrowserWindow`, across the trust boundary AGENTS.md calls non-negotiable, and takes
`team-webrtc-client-transport.ts` (759 lines), the bridge envelope, the host peer and the gateway
with it. The multiplexed host cannot survive it either: it builds each child peer by spreading the
client `PeerState`, and the shared peer has no host concept to spread from.

Unifying ed25519 would be a downgrade rather than a dedup. The security-relevant half is already
shared — all four handshake sites call `teamProtocolV2AuthenticationTranscript` from contracts. What
is left is six lines of Node's native ed25519 in `src/main` against ninety lines of hand-rolled
SPKI/PEM parsing plus a JS implementation. The stated reason for the swap ("the shared peer must
work in a browser") only applies if the peer runs in `src/main`, which the first point rules out.

The drift that was Phase B's evidence — the TURN constants — is fixed in `b707d6bf` at a small
fraction of the risk. Happy to scope the rest as its own branch if you want it anyway.

---


NorbertBodziony:
Split the runtime rebuild out of this PR into #261, per the review's scope finding. Moved: the macOS
dylib bundler and its verifier check, the workflow path/guard update, the `recipeVersion` 8 → 9 bump
that unpins the runtime, and the `fast-uri` lockfile bump that unblocks the rebuild. Nothing product
-side moved — this branch is unchanged apart from those six files, and #261 has exactly what came
out.

That also clears the P1 on `native-runtime.lock.json`: this PR no longer touches the lock, so merging
it cannot leave `main` unable to package. #261 stays a draft until the publish job's pin commit can
be folded into it.

Still open here from the review: the fail-closed suggestion on `remote-agent-decoding.ts`. Leaving it
as-is — the partial accept is deliberate and the file argues the case at the decoder. Treating one
unrepresentable model id as a protocol error took the whole server offline (agents, browser, desktop)
until an explicit reconnect, which is what `claude-fable-5-1[1m]` did.

NorbertBodziony:
Finding 1 is right and is fixed in `1c831b5e`. Moving the renderer onto the shared decoder dropped a
bound its own schema had carried — `sessionExpiresAt` was `z.number().int().positive()` and became a
plain integer — so zero and negatives decoded and were forwarded across the renderer-to-main
boundary, where `team-webrtc-bridge.ts` requires a positive integer and parses inside the
MessagePort listener. `peer-ready` now takes a `timestamp` guard; `sdpMLineIndex` keeps plain
`integer`, since zero is the first m-line rather than a malformed one. Added
`packages/contracts/src/signal-protocol/decode.test.ts` and watched it fail with `integer` restored
at the call site.

Finding 2 I think is wrong, and the reason is that `ensureCompatibility` does not merely fail one
call. `src/main/remote-server-client.ts:253-271`: when a `protocol_error` is on record it throws
**without touching the network** on every later call, until an explicit reconnect passes `refresh`.
`decodeOrProtocolError` on the model list is one of the paths that records it
(`remote-server-client.ts:432-436` → `#connections.reportError`), and lines 141, 153, 185 and 211
show agents, browser and remote desktop all pass through the same gate. So a whole-array rejection
does not degrade the model menu — it takes the entire server offline until the user reconnects by
hand. That is what `claude-fable-5-1[1m]` did.

The bracket-compatible guard addresses the id we saw, not the class. Model ids come from provider
CLIs that neither end of this wire controls, so a host on a newer CLI can always offer one this
build has no way to spell; a menu is the one payload here where an unrepresentable entry is a
missing option rather than a host talking nonsense, and an entry that fails the guard could not have
been selected anyway. A payload that is not a list at all still throws — the array itself is
validated, only its members are filtered.

Finding 3 stands and is the developer's call, not mine to act on unasked. The native-packaging half
did move out, to #261.

/norbiai review

NorbertBodziony:
Finding 1: you are right and I withdraw the rebuttal. `packages/contracts/AGENTS.md:22` says a
malformed known payload fails closed as `protocol_error`, and I argued the consequence instead of
the rule. Fixed in `9bcc4165` — both decoders validate the whole array again, and the Team API test
now asserts the contract (request rejects, connection records `protocol_error`) rather than partial
acceptance. Watched it fail with the filter restored: "promise resolved instead of rejecting".

Worth recording where the real fix sits, because the severity did not go away: failing closed here
takes agents, browser and remote desktop offline together until an explicit reconnect, so
`isAgentModel`'s charset is now the only thing between a legitimate provider-CLI id and a bricked
connection. That is why a model id is an opaque token there rather than a list of the models this
build knows, and `ipc.test.ts` pins the bracketed form. Said so in the comments on both decoders and
on the guard.

Finding 2 stands and is the developer's to decide — I am not splitting further unasked.

/norbiai review

NorbertBodziony:
Finding 1 is real and is fixed in `666c8b4d`. Confirmed the trace: `createSession` throws before a
session exists, so nothing calls `closeSession`, `#runtimeState` stays cached, `#ensureRuntime`
returns it untouched, and `SunshineMoonlightRuntime.start` returns the state it already holds — so
the process that was refused never restarted and "then try again" could not work. The gateway now
drops the runtime alongside the refusal, so the next attempt is a fresh Sunshine that reads the new
grant. It waits for the last session to close when one is live: that runtime is shared, and ending
another member's session is not this member's failed create's business, which `closeSession` already
covers.

`#stopRuntime` gathers the three lines the two existing teardowns each spelled out — a cleared
`#runtimeState` beside a live `#runtime` is exactly the state that latched this.

New test in `remote-screen-gateway.test.ts` is the reverse of the refusal test beside it. The fake
now models the real latch: a runtime is denied for its whole life and granting shows up as the next
one started. Watched it fail without the teardown — the second create rejects with the same
"Grant screen recording on the host, then try again."

Finding 2, the scope split: the developer has decided to keep this PR whole and merge it as one. So
it is not being split, and it is not an open question.

/norbiai review

NorbertBodziony:
Finding 1 is real and is fixed in `e989f508`. Confirmed both halves of it: `failPeer` only posts a
`peer-error`, and `classifyTransportError` (`remote-server-connection-status.ts:113`) reads
`webrtc_error` as `network_unavailable` with `retryable: true` — so the client reconnected into the
same service to be sent the same bytes, and the socket stayed open so the next frame arrived on a
connection whose state the dropped one was supposed to set.

Worth noting this catch is not new — `main` has the same `.catch((error) => failPeer(state, error))`
around `signalMessageSchema.parse`. What made the path live is this PR: the old schema was a
`.loose()` bag of optionals that threw for almost nothing, and the shared decoder throws for every
malformed known frame. So the finding belongs here.

Decode failures now report `protocol_error` and disconnect the peer.
`classifyTransportError` already suspends reconnection for that code and records the server
`incompatible`, so nothing downstream needed changing. The host gateway ignoring it is right —
`#recoverSignal` re-mints a ticket, which cannot fix a wire disagreement.

The outer catch keeps `failPeer` and now means only what handling a frame this peer *did* read can
throw — an ICE or SDP operation the browser refused, which is a connection failing and which a
reconnect can still fix. Splitting the two is the point.

New test in `team-webrtc.test.ts` sends a `ready` frame carrying an empty `resumeToken` and asserts
`protocol_error`, one `peer-disconnected`, and a closed socket. Watched it fail with the decode
failure routed back to `failPeer` — `code: "webrtc_error"`. The bridge start-up is now shared
between the two tests instead of duplicated.

Finding 2, the scope split: the developer has decided to keep this PR whole. That is settled, not an
open question.

/norbiai review

NorbertBodziony:
Both findings from the last round were real. Fixed in `d57fe236` and `2e7cd50f`.

**P1 — negative ICE indices.** `remote/api/src/protocol.ts:70` takes `sdpMLineIndex` as
`z.number().int().nonnegative().nullable()`; the client decoder took any integer, so `-1` decoded,
reached `addIceCandidate`, and came back as a browser refusal that `failPeer` reports as
`webrtc_error` — `network_unavailable`, retryable, socket still open. Exactly the path `e989f508`
closed for every other malformed frame. `decode.ts` now has an `mLineIndex` guard bounded at `< 0`
rather than the timestamp's `<= 0`, because zero is the first m-line; the test asserts both halves.
Watched it fail with the field back on plain `integer`: "expected [Function] to throw an error",
with the zero case still green.

**P2 — the HTTPS event stream left behind.** Confirmed, and slightly wider than reported: the same
line also drops *every* non-WebRTC entry in a released build (`keepOtherTransports` is off), and
`removedHostIds` reported neither, so neither was ever forgotten. `RemoteEventStream` has one
teardown — `forget` — and `ensure` branches on the transport it finds without aborting an HTTPS
controller already running for that id, so both transports deliver the same events and the stale
socket can mark a healthy host offline.

`reconcileWebRtcHosts` now returns `staleTransportHostIds`. It is a second list rather than more ids
in `removedHostIds` because the caller does something different with it: no `transport.disconnect`,
since these entries were never on the WebRTC transport and where the id survives the only thing a
disconnect could name is the entry that just replaced them. `#syncWebRtcHosts` clears their
connection state before `replaceServers`, because after it `ensure` answers for the new entry. Two
existing tests carry the report rather than a new one; watched both fail with the push removed
("expected [] to deeply equal [ 'host' ]").

Ran `bun run test:desktop` on `decode.test.ts`, `remote-server-host-directory.test.ts` and
`remote-server-manager.test.ts`, plus `bun run lint` and `bun run typecheck` (all 11 projects —
`typecheck:remote` and `typecheck:contracts` both green, which is what the shared decoder change
needed).

/norbiai review

NorbertBodziony:
P1 confirmed and fixed in `00c2b9ef`. It is the same defect `e989f508` closed in the desktop
renderer, on the surface I failed to carry it to — `packages/team-client` is what mobile runs, and
AGENTS.md "Hit every surface" names mobile explicitly. Traced the whole chain before fixing:
`handleSignal` decodes → the chain's `.catch` → `failPeer` → a plain `offline` update →
`mobile-workspace-context.tsx:632` → `recovery.offline` → `scheduleRetry`, which is five attempts
ten seconds apart and then one every two minutes for as long as the app is open.

**The peer** now splits the decode from the handling exactly as the renderer does. A decode failure
reports `code: "protocol_error"`; the outer `.catch` keeps `failPeer` and now means only what
handling a frame this peer *did* read can throw — an ICE or SDP operation the browser refused. The
synchronous `try` around `JSON.parse` is gone with it: a frame that is not JSON is the same
disagreement about the wire, not a different one.

**The recovery controller** gains `suspend`, which stops the loop rather than joining it, and mobile
routes `protocol_error` there. On reverse states — it is reversible three ways, so it is not a state
the app cannot leave: `refresh`, switching servers (the controller is per-server and is rebuilt),
and returning to the foreground. The last one is deliberate and the one a user reaches without
knowing it exists: the desktop they left to update is the reason the frame was unreadable, and it
starts one attempt, not a loop.

`"Signal returned an invalid message."` joins `SAFE_CONNECTION_ERRORS` — a fixed protocol string
with no user data. A `JSON.parse` message, which *can* echo frame bytes, is not in the set and still
redacts to `"SyntaxError."`.

Watched both new tests fail: the peer test with the classification dropped (`code` missing from the
update), the recovery test with `suspend` doing what `offline` does ("expected 2 to be 1"). Ran
`test:desktop` on `remote-peer.test.ts` and `remote-recovery.test.ts`, plus `lint`, `typecheck` (all
11, `typecheck:mobile` and `typecheck:team-client` included) and `check:ui`.

On the P3: the repository owner decided to keep #260 whole and merge it as one. That is settled and
not a finding I can close by argument — please treat it as accepted rather than re-raising it.

/norbiai review

PR #261, opened 2026-09-06
build(runtime): rebuild the remote desktop runtime so it starts without Homebrew

NorbertBodziony:
Split out of #260, which was carrying this alongside the shared-client extraction. Nothing here
touches product code; #260 keeps the rest.

## The shipping bug

`otool -L` on the Sunshine binary inside the released **OpenBot 0.2.0** DMG:

```
/opt/homebrew/opt/miniupnpc/lib/libminiupnpc.21.dylib
/opt/homebrew/opt/openssl@3/lib/libssl.3.dylib
/opt/homebrew/opt/openssl@3/lib/libcrypto.3.dylib
```

CMake linked the build runner's Homebrew by absolute path. On any Mac without Homebrew and both
formulae, dyld cannot start Sunshine — and nothing downstream says so: the app still reports remote
control as available, the session still opens, and no frame ever arrives. `web-server` and
`streamer` in the same release are clean, which is why this went unnoticed.

## What is here

- `scripts/mac-runtime-dylibs.ts` walks the load commands transitively, copies every non-portable
  library into `Frameworks`, and rewrites the paths with `install_name_tool`. It runs **before**
  `codesign`, because rewriting a load command invalidates the signature over it.
- `scripts/verify-remote-desktop-runtime.ts` checked presence and checksums only, so it was green on
  the broken release. It now fails on any load command outside `@…`, `/usr/lib` or `/System`.
- The workflow's path triggers and its recipe-bump guard now count the new bundler, so a change to it
  cannot ship against a runtime built before it existed.
- `recipeVersion` 8 → 9 with `artifactRelease`/`releaseArtifacts` dropped. The input digest
  deliberately does not cover the build script, so without this bump the pinned artifacts stay
  reusable and the fix reaches nobody.
- `fast-uri` 3.1.5 → 3.1.7 in the Sunshine fork's lockfile (`NorbertBodziony/Sunshine@cde2ca69`, and
  the recorded patch says the same). Four advisories landed against 3.0.0–3.1.5 and
  `npm audit --audit-level=high` runs before anything compiles, so both builds died in a minute. It
  is a dev-only dependency of `ajv` under the web UI's vite tooling and never shipped, but bumping it
  rather than narrowing the gate to `--omit=dev` keeps the audit able to fail on a compromised build
  tool, which is the case it is there for.

### Bundling creates a licence obligation, so the notices ship too

Copying those three libraries into the runtime is what makes them OpenBot's to redistribute — until
this branch they stayed on the machine that linked them and the distribution owed nothing. The build
wrote two GPL licence files and no more. miniupnpc is BSD-3-Clause, whose binary clause is explicit
about reproducing the notice, and OpenSSL 3 is Apache-2.0.

`REDISTRIBUTED_LIBRARIES` names each library allowed into the runtime and the notice it obliges,
keyed by the name before the version so a Homebrew bump from `libssl.3.dylib` does not silently drop
the entry. The licence text is taken from the tree that was actually linked against, so it describes
the binary in the archive rather than whatever the formula says later. Verification reads the shipped
tree rather than a list: a library that starts being bundled arrives as a failure naming it, not as a
release quietly missing a licence. Unlisted means the build stops.

### The pin job could not do what this PR asked of it

`pinRemoteDesktopRuntime` derives the input digest from the lock it is handed and refuses a manifest
that does not match, so the job only works where the checked-out recipe is the one the release was
built from. It hardcoded `ref: main`, which holds for the push-to-main flow it was written for and
fails for the one case that needs it most — a branch that changes the recipe, published before
merging so `main` never carries an unpinned lock. It is now gated on `main` rather than taught to
check out the branch: pinning from a branch would open a lock PR whose diff carries that whole
branch, and pushing the commit onto someone's PR branch from CI races their local work. The fold is
manual, which the flow already assumed without writing down, and `docs/RELEASING.md` now has the
sequence.

### Running the pin found it deleting the provider runtimes

`native-runtime.lock.json` answers to two schemas: `agent-runtime-lock.ts` owns `codex`, `claude`
and `grok`, and `native-runtime-lock.ts` declares only `cloudflared` and `remoteDesktop`. Zod keeps
what a schema names and drops the rest — correct while reading, destructive on the way back out —
and `pin-remote-desktop-runtime.ts` serialized the parsed lock straight over the file. Every
provider pin went with it: eighty lines of assets, digests and licence hashes that
`src/main/provider-runtime-manager.ts:78` parses to bundle the CLIs.

Nothing downstream caught it. The pin written was correct, and the job's guard is
`[[ "$changed" != "native-runtime.lock.json" ]]` — it counts the files a pin changed, not the keys
it lost, so the lock could drop three whole sections with every check green. The script runs twice
per publish: once in `pin` to open the lock PR, and once in `verify-published` before it packages
the app, so a published runtime went through it either way.

The pin now edits the document it read and copies out everything else, and the job formats the
result: `JSON.stringify` puts every array on its own lines and Biome collapses the short ones, and
`pin` commits with no hook to notice, so the lock PR used to open with `lint` already red.

## Merging

`main` must never see this recipe without its pins: while the lock is unpinned,
`prepare-remote-desktop-runtime.ts` throws and takes every `package`, `dist:mac`, `dist:win` and
release run with it. So the recipe and the pin land together, in this branch — published from the
branch by `workflow_dispatch`, then folded in by hand for the reason above. **Both are here now**:
`native-runtime.lock.json` carries `artifactRelease` and both target entries, 16 added lines and no
deletions beyond `"releaseArtifacts": {}`.

Input digest: `a6e49e7ac064584df4e4bc3dd9531b98e99fcfa1bf0f690ee1557ac351d60ab9`.

### `Verify published darwin-arm64` fails, and it is not this branch

The dispatch that published the release fails that job twice, identically:
`A second OpenBot instance did not exit` from `scripts/verify-macos-package.ts:243`. The runtime is
not implicated — the three bundled dylibs report `--validated` inside the packaged app,
`Sunshine.app` satisfies its Designated Requirement, and every step through the macOS smoke test
passes. It cannot fail this PR (`publish` is gated `!= 'pull_request'`, so the job never runs here)
and it will not fail `main` on merge (the release for this digest exists, so `build`, `publish` and
`verify-published` all skip). The job has never executed in the last 20 runs of this workflow, so
there is no green baseline: `verify:package` passes in `release.yml`, which builds signed and
notarized, and this job runs it against
`electron-builder --dir --config.mac.identity=null`. Pre-existing, and left for its own change
rather than folded in here.

## Verification

`bun run lint` and `bun run typecheck` (all 11 projects) green. The `fast-uri` bump was confirmed
against a clean `npm audit --audit-level=high` on the patched tree, and the archive at the new fork
commit was checked to contain 3.1.7 before its `sourceSha256` was recorded.

The linkage fix is verified by CI: the same build that produces the artifacts runs the load-path
check that failed against 0.2.0. The licence resolution was exercised against the real Homebrew trees
— `miniupnpc/2.3.3` and `openssl@3/3.6.3`, the same OpenSSL version the runner reports — confirming
it finds `LICENSE` and `LICENSE.txt`, collapses `libssl` and `libcrypto` to one notice, and throws
by name on a library that is not on the list.

The pin regression has a test in `native-runtime-lock.test.ts` asserting against the other owner: a
pinned document must still satisfy `parseAgentRuntimeLock`, unchanged. Watched it fail — restoring
the old one-line serialization turns it red with three `expected object, received undefined`, one
each for `codex`, `claude` and `grok`, leaving the other eight green. The pinned lock was then
re-checked through both loaders.


NorbertBodziony:
### [REMAINS][P1] runtime artifacts unpinned — fixed in `22e913ff`, but the pin needed a fix of its own first

`native-runtime.lock.json` now carries `artifactRelease` and both target entries for
`remote-desktop-runtime-a6e49e7ac064584df4e4bc3dd9531b98e99fcfa1bf0f690ee1557ac351d60ab9`.

Running the pin is what surfaced the bug in `c96606c5`, which is the more important of the two
commits.

**`pin-remote-desktop-runtime.ts` was deleting the provider runtimes.** `native-runtime.lock.json`
answers to two schemas: `agent-runtime-lock.ts` owns `codex`, `claude` and `grok`, and
`native-runtime-lock.ts` declares only `cloudflared` and `remoteDesktop`. Zod keeps what a schema
names and drops the rest — correct while reading, destructive on the way back out — and the script
serialized the parsed lock straight over the file. Every provider pin went with it: eighty lines of
assets, digests and licence hashes that `src/main/provider-runtime-manager.ts:78` parses to bundle
the CLIs.

Nothing downstream caught this. The pin it wrote was correct, and the job's own guard

```bash
changed="$(git diff --name-only)"
if [[ "$changed" != "native-runtime.lock.json" ]]; then
```

counts the files a pin changed, not the keys it lost — so the lock could drop three whole sections
with every check green. The script runs twice per publish: once in `pin` to open the lock PR, and
once in `verify-published` before it packages the app, so a published runtime went through it either
way.

The fix edits the document it read and copies out everything else. It also formats the result:
`JSON.stringify` puts every array on its own lines and Biome collapses the short ones, and `pin`
commits with no hook to notice, so the lock PR opened with `lint` already red.

The test is in `native-runtime-lock.test.ts` and asserts against the other owner — a pinned document
must still satisfy `parseAgentRuntimeLock`, unchanged. Watched it fail (AGENTS.md rule 2): restoring
the old one-line serialization turns it red with three `expected object, received undefined`, one
each for `codex`, `claude` and `grok`, and leaves the other eight tests green.

The resulting lock diff is 16 added lines and no deletions beyond `"releaseArtifacts": {}`.

### One thing this PR does not fix

`Verify published darwin-arm64` fails in the dispatch, twice, identically:
`A second OpenBot instance did not exit` from `scripts/verify-macos-package.ts:243`, 27s in — both
the 15s and 10s waits elapsed. It is not the runtime: the three bundled dylibs report `--validated`
inside the packaged app and `Sunshine.app: satisfies its Designated Requirement`, and steps 1–9,
including `Verify published runtime` and the macOS smoke test, all pass.

I could not attribute it to this PR, and I do not think it is ours:

- The job has **never executed** in the last 20 runs of this workflow — skipped or cancelled every
  time. There is no green baseline to regress from; this dispatch is its first real run.
- It cannot fail on this PR: `publish` is gated `if: github.event_name != 'pull_request'`, so
  `verify-published` never runs here.
- It will not fail on `main` when this merges: the release for digest `a6e49e7a` already exists, so
  `build` is skipped (`release_exists != 'true' || workflow_dispatch`), and `publish` and
  `verify-published` skip with it.

I checked the obvious suspect and it is innocent: the corrupted lock above cannot explain it.
`ProviderRuntimeManager` is constructed in an async post-ready path
(`application-services.ts:298`), and a second instance exits at `src/main/index.ts:447` before that
runs.

So it is a latent, pre-existing failure that bites whoever next bumps `recipeVersion`, not a blocker
here. Flagging it rather than folding an unrelated Electron single-instance investigation into this
PR.

PR #273, opened 2026-09-07
feat(dev): generate local Auth API secrets when a fork has no .env.keys

michaldudek:
Public checkouts cannot decrypt the tracked dotenvx files, so bun run dev now writes gitignored contributor keys and an overlay instead of failing closed.

When trying to run `bun run dev` on a fresh clone (following setup instructions) I stumbled on:

```
2026-09-06T20:44:32.032Z ERROR [dev-services] Missing or empty .env.keys. Add it to the local checkout so Codex can copy it through .worktreeinclude.
```

Screenshot 2026-09-07 at 12 20 07

## What changed

`bun run dev` no longer requires the author's ignored `.env.keys`. A public fork generates gitignored contributor keys and `apps/auth-api/.env.dev.local` (local login code, ticket JWKS, placeholder SMTP secret) so the Auth API and Signal service can start. Tracked production ciphertext is left alone. JSON secrets are single-quoted so dotenvx decrypts them as JSON, not `\"`-escaped text.

Surfaces: `apps/auth-api` local env loading, local Signal dotenvx path, README / CONTRIBUTING / Auth API docs. Not renderer, mobile, IPC contracts, or migrations.

## Verification

  - `bun run test:desktop -- scripts/development-secrets.test.ts scripts/prepare-dev-environment.test.ts scripts/dev-services.test.ts`
  - `bun run lint` and `bun run typecheck`
  - Watch-it-fail: overlay generation without `AUTH_EXPOSE_DEVELOPMENT_CODE=true`; JSON.stringify quoting left `\\` in JWKS and dotenvx could not parse it
- [x] Surfaces touched are named under "What changed": renderer, mobile, `apps/auth-api`, the three `--openbot-*` palettes, IPC contracts and `mock-openbot.ts`, reverse states, migrations and the latest schema, documentation
- [x] Relevant manual smoke test completed, or not applicable — `dotenvx run --strict` on the overlay parses ticket JWKS; `bun run dev` injects the local env and starts the API

## Risk and security impact

None for the renderer-to-main trust boundary. Contributor secrets stay gitignored and cannot decrypt the tracked production files. Local development exposes the one-time login code in the API response (`AUTH_EXPOSE_DEVELOPMENT_CODE`), matching the existing tracked `.env.dev` intent.

### DIFF
--- a/.worktreeinclude
+++ b/.worktreeinclude
@@ -1,4 +1,6 @@
 # Local dotenvx decryption key used by the development Auth API.
 .env.keys
+# Contributor-generated Auth API secrets when the tracked .env.dev cannot be decrypted.
+apps/auth-api/.env.dev.local
 # Local dotenvx decryption key used by the Remote production stack.
 remote/.env.keys
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,8 +27,9 @@ belongs.
 
 A fresh worktree has no `node_modules`, and every command below dies with `biome: command not found`
 or `Cannot find module '@openbot/logging'` until it does. Run `bun install --frozen-lockfile` first,
-or `bun scripts/prepare-dev-environment.ts`, which adds the local D1 migration and asserts the Bun
-version and the `.env.keys` `.worktreeinclude` carries across.
+or `bun scripts/prepare-dev-environment.ts`, which adds the local D1 migration, asserts the Bun
+version, and either copies `.env.keys` through `.worktreeinclude` or generates contributor-local
+keys and `apps/auth-api/.env.dev.local` when the author's file is not present.
 
 Before you call a task done, run the narrowest test for what you touched, then `bun run lint` and
 `bun run typecheck` — plus `bun run check:ui` if you touched `src/renderer`, which scans the whole
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,9 @@ bun run check
 bun run dev
 ```
 
+`bun run dev` generates ignored local Auth API secrets when `.env.keys` is missing, so a public
+fork can run the app without the author's dotenvx key.
+
 The supported toolchain is pinned in `package.json`. Use stable Bun 1.4.0, TypeScript 5.9, Vite 7, and the
 existing Biome configuration. Biome is the only lint and format tool. Do not add a second linter,
 Prettier, a second state library, or a UI kit without first discussing the architectural cost.
--- a/README.md
+++ b/README.md
@@ -110,6 +110,10 @@ bun run codex:doctor
 bun run dev
 ```
 
+A fork without the author's ignored `.env.keys` file is fine: `bun run dev` generates a local
+keypair and `apps/auth-api/.env.dev.local`. Those files stay gitignored. Do not commit them.
+Local login returns the one-time code in the API response (`AUTH_EXPOSE_DEVELOPMENT_CODE`).
+
 `codex:doctor` checks the CLI version, App Server handshake, ChatGPT login, and Computer Use plugin
 without starting a model turn.
 
@@ -135,6 +139,7 @@ Use `bun run dev:seed --dry-run` to inspect the target and fixture counts withou
 | Command | Purpose |
 | --- | --- |
 | `bun run dev` | Start the local Auth API, Signal service, and Electron client with renderer HMR on its app profile. |
+| `bun run env:bootstrap` | Generate ignored local `.env.keys` and `apps/auth-api/.env.dev.local` when the author's key is not present. |
 | `bun run preview` | Preview the built Electron client with the green preview icon. |
 | `bun run dev:api` | Start the TanStack Start API and its local D1 database on `127.0.0.1:3100`. |
 | `bun run api:start` | Build and preview the Cloudflare Worker locally. |
--- a/apps/auth-api/README.md
+++ b/apps/auth-api/README.md
@@ -13,6 +13,10 @@ private keys stay in the ignored root `.env.keys` file. Dotenvx decrypts the
 selected file only in process memory. The explicit development flag returns the
 code in the API response. It never writes the code to logs.
 
+A checkout without `.env.keys` (a public fork) generates contributor-local secrets
+on `bun run dev`: a new `.env.keys` and gitignored `apps/auth-api/.env.dev.local`.
+Do not commit either file. They cannot decrypt the tracked production ciphertext.
+
 ```bash
 bun run api:migrate:local
 bun run dev:api
@@ -29,7 +33,7 @@ bun run env:validate:dev
 bun run env:validate:prod
 ```
 
-Commit `.env.dev` and `.env.production`. Never commit `.env.keys`.
+Commit `.env.dev` and `.env.production`. Never commit `.env.keys` or `.env.dev.local`.
 
 ## Email delivery
 
--- a/apps/auth-api/package.json
+++ b/apps/auth-api/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "dotenvx run -f .env.dev -fk ../../.env.keys -- vite dev",
+    "dev": "bun ../../scripts/auth-api-dev.ts",
     "start": "bun run build && bun run preview",
     "build": "vite build",
     "preview": "dotenvx run -f .env.production -fk ../../.env.keys -- vite preview --port 3100",
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "remote:check:compose": "bun remote/scripts/check.ts --compose",
     "remote:env:validate": "cd remote && bin/dotenvx validate -f .env.production -fk .env.keys",
     "remote:update": "remote/bin/dotenvx run --overload -f remote/.env.production -fk remote/.env.keys -- bun remote/scripts/update.ts",
+    "env:bootstrap": "bun scripts/development-secrets.ts",
     "env:encrypt:dev": "bun run --cwd apps/auth-api env:encrypt:dev",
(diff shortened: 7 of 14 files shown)


NorbertBodziony:
Thanks for this — the diagnosis is right and this is the approach I want. The ordering in `prepareDevelopmentEnvironment` is the detail I'd have got wrong: running `ensureDevelopmentSecrets` *after* `bun install --frozen-lockfile` is what keeps dotenvx available on the normal path. The keypair is correct too (P-256/ES256 matches `importJWK(..., "ES256")` in `remote-control-plane.ts`, and the `kid` lines up with `REMOTE_TICKET_KEY_ID` so `parseJwks` finds it).

One change to merge, and it's a deletion.

## Drop the encryption on the generated overlay

`apps/auth-api/.env.dev.local` should be plaintext. It's gitignored twice over (`.env.*` and `*.local`) and never committed, and the decryption key would sit in `.env.keys` on the same disk right next to it — so encrypting it buys nothing. dotenvx encryption is there to make a file safe to *commit*; this file isn't one.

More to the point, the encryption is what's causing the two problems in the PR:

- **NorbiAI's P2 is a direct consequence.** `resolveOverlaySecret` returns `undefined` both when dotenvx is missing and when decryption fails, so `isRemoteTicketJson(undefined, undefined)` is `false` and the whole overlay gets rewritten. If the file is plaintext, "cannot decrypt" stops existing as a state and the bug goes away by construction rather than needing a tri-state. Keep `contributorOverlayNeedsRepair` — just have it read the values straight off the file and drop `resolveOverlaySecret` and `dotenvxExecutable`.
- **The fallback in `encryptContributorDevelopmentSecrets` writes a key paired with nothing.** When dotenvx isn't found it writes `DOTENV_PRIVATE_KEY_DEV_LOCAL=<randomBytes(32)>`, which matches no public key, and leaves the overlay in plaintext anyway. `assertDevelopmentSecrets` then passes because the file is merely non-empty, so nothing surfaces — and it never self-heals, because on the next run the plaintext values parse as valid JSON and `needsRepair` is `false`.

So: delete `encryptContributorDevelopmentSecrets`, `dotenvxExecutable`, `resolveOverlaySecret` and the `DevelopmentSecretEncryptor` option, which also lets the tests drop their injected `encrypt` stub.

The thing I most want out of this is that **nothing in the dev bootstrap path writes `.env.keys` at all.** That file holds `DOTENV_PRIVATE_KEY_PRODUCTION`, which decrypts the tracked Cloudflare DNS token, TURN shared secret and session secret, and there is no backup of it. Right now `encryptContributorDevelopmentSecrets` is the only function that touches it and every test injects a stub over it, so its real behaviour — including whether `dotenvx encrypt -fk` appends to the keys file or rewrites it — is untested. Removing the write is a better answer than adding a test for it.

Once bootstrap stops creating a key file, `assertDevelopmentSecrets` needs to accept *either* my `.env.keys` or a generated overlay, instead of requiring the former.

Keep the single-quote serialization — `dotenvx run -f` still has to parse the plaintext file, so your `JSON.stringify` fix is still load-bearing. Keep the `.worktreeinclude` entry too.

## One small addition

Log which env file `developmentAuthEnvFile()` picked. It silently prefers `.env.dev.local` over the tracked `.env.dev`, so if an overlay ever gets generated on a checkout that *does* have `.env.keys`, real dev values stop being used with nothing saying so. `dev-services` already logs port substitutions; this is the more consequential one.

## Not yours

Two things this PR exposed that I'll handle separately, so don't feel you need to touch them:

- `scripts/set-auth-smtp-secret.ts` writes the same app password into `.env.dev` and `.env.production`, so the tracked dev file carries a live mail credential. That's why you needed the `unused-local-development` placeholder — `readSmtpConfig` is all-or-nothing and the other four SMTP vars are already plaintext in `wrangler.jsonc`. I'm moving the password out of dev and making a missing password mean "not configured" instead of "incomplete", which will let the placeholder go.
- After that, `.env.dev` has nothing left in it worth encrypting, and the dev keypair can go entirely.

Make those two changes and I'll merge.

NorbertBodziony:
Hold off on my previous comment — don't spend time on the overlay's encryption. Talking it through, there's a simpler destination and it makes your PR smaller rather than larger.

## Where this is going

Instead of a second file that shadows the tracked one, **`apps/auth-api/.env.dev` stops being tracked and gets generated on every machine, mine included.** One file, one path, no precedence rule — and no separate fork code path, so the setup I use every day is the same one you exercise.

The only thing standing in the way was that `.env.dev` carried a live credential: `set-auth-smtp-secret.ts` wrote the same Private Email app password into dev and production. I've fixed that on my side, since it needs the mailbox — email delivery is now simply off in local development. That's also what retires your `unused-local-development` placeholder: the reason a placeholder was needed at all is that `wrangler.jsonc` sets four of the five SMTP variables in the top-level `vars` that local `vite dev` reads, so removing just the password left the partial set `readSmtpConfig` rejects. The four names now join `LOCAL_RUNTIME_KEYS` and the dev file blanks all five, so nothing is configured and delivery returns `null`. I'll push that shortly; rebase onto it.

## What your PR becomes

Keep, unchanged — this is the part that's actually yours:

- `createContributorDevelopmentSecrets` and the P-256/ES256 keypair generation
- `serializeEnvFile`, including the single-quote fix — `dotenvx run -f` still parses the file, so that's still load-bearing
- moving `ensureDevelopmentSecrets` after `bun install --frozen-lockfile` in `prepareDevelopmentEnvironment`
- the `assertDevelopmentSecrets` message fix, and the README / CONTRIBUTING / Auth API docs

Drop entirely:

- `scripts/auth-api-dev.ts`, `developmentAuthEnvFile()` and the overlay precedence rule
- the `.worktreeinclude` entry
- all the encryption machinery — `encryptContributorDevelopmentSecrets`, `dotenvxExecutable`, `resolveOverlaySecret`, the `DevelopmentSecretEncryptor` option, and the `encrypt` stub the tests inject

`scripts/dev-services.ts` and `apps/auth-api/package.json` then need no changes at all, and NorbiAI's P2 finding disappears with `resolveOverlaySecret` rather than needing a tri-state.

Changes:

- The generator writes `apps/auth-api/.env.dev` directly, in plaintext, only when the file is absent. Nothing in the bootstrap path should write `.env.keys` — that file holds the production private key and there is no backup of it.
- Add the five blank SMTP variables to the generated set alongside what you already generate.
- Do the untracking in your PR, so the generator and the untracking land together and a fresh clone is never left without the file: `git rm --cached apps/auth-api/.env.dev`, then drop the `!apps/auth-api/.env.dev` negation from `.gitignore` so `.env.*` covers it. `--cached` leaves existing working copies on disk, so nobody's setup breaks. `env:encrypt:dev` and `env:validate:dev` go too.
- Keep the repair check — with a plaintext file it just reads the values, so "cannot decrypt" stops existing as a state.
- Still worth the log line when the generator writes the file.

I'll rotate the mail password and drop `DOTENV_PRIVATE_KEY_DEV` from my `.env.keys` on my side.

Sorry for the reversal — your PR is what surfaced that the dev file never needed to be encrypted in the first place, which is the more useful outcome.

NorbertBodziony:
Thank you — and an apology, because this landed in the wrong order.

Your PR is what found this. The error in your screenshot was the real state of the repo for anyone outside my machine, and nobody had hit it because we all had `.env.keys` sitting in the checkout already. Working through your diff is also what surfaced the thing that mattered more than the fix itself: `.env.dev` was tracked *and* encrypted *and* carried a live Private Email app password, so it needed a key it should never have needed in the first place. You went looking for a way to start the app and found a credential leak.

Then I implemented the destination I described in my second comment myself instead of handing it back to you to rebase onto. That was yours to finish and I took it. Sorry — that's on me, not on anything about your PR.

**What merged**

- #279 — takes the app password out of `.env.dev`, adds the four SMTP names to `LOCAL_RUNTIME_KEYS` so the file can blank all five, and email delivery is simply off in local development
- #282 — `apps/auth-api/.env.dev` is untracked and generated per checkout, in plaintext, on first `bun run dev`

Which is my second comment on this PR, near enough line for line.

**What of yours is in it**

- `scripts/development-secrets.ts` and `scripts/development-secrets.test.ts` — your file names and your split between generation and the bootstrap script
- the P-256/ES256 pair with `kid` matched to `REMOTE_TICKET_KEY_ID` so `parseJwks` finds it, same construction
- generate only when the file is absent, never clobber a file someone already has — your repair check, minus the state that made it hard
- the `.worktreeinclude` entry, which I told you to drop and then added back for the generated file

**What didn't carry over, so you're not left guessing**

- Your single-quote serialization. The merged version writes the JWK bare — `REMOTE_TICKET_PRIVATE_JWK={"kty":"EC",...}` — and dotenvx parses to end of line. Your quoting was load-bearing for a value making a round trip through `dotenvx encrypt`; with no encryption left there's no round trip to survive. It was right for the design it was in.
- The ordering. You put `ensureDevelopmentSecrets` *after* `bun install --frozen-lockfile` because dotenvx had to exist by then, and that was correct. Plaintext needs no dotenvx, so it moved back *before* the install — which is what lets it use node builtins only and run on a clone with no `node_modules` at all.

**One real piece of yours that didn't land**

Neither `README.md` nor `CONTRIBUTING.md` says anything about this today. Your PR added the fork paragraph to both and I didn't carry it across, so a contributor still has to find out by running `bun run dev` and watching it work. The shape is different enough that it's a rewrite rather than a rebase — one generated `apps/auth-api/.env.dev`, no `.env.dev.local`, no `env:bootstrap`, no generated `.env.keys`, and the note is now "this just works" rather than "here's the fork path" — but it's the same paragraph doing the same job, and it should have your name on it. Want to open it?

I'll close this one as superseded unless you'd rather keep it open.

michaldudek:
👌

PR #278, opened 2026-09-07
fix(claude): register update_profile in the Claude MCP tool mirror

michaldudek:
Codex and Grok receive OPENBOT_DYNAMIC_TOOLS, but Claude gets a hand-written in-process MCP server that never declared update_profile, so a Claude agent was instructed to use a tool it could not call and correctly refused.

The Claude test now holds the mirror to the catalogue: the set of registered tool names must equal OPENBOT_DYNAMIC_TOOLS minus ask_user (Claude uses the SDK's AskUserQuestion). Watched it fail naming update_profile before the fix.

## What changed

A Claude agent told to rename itself or change its title answered that `update_profile` did not exist in its environment, and it was right. Codex and Grok receive the tool through `OPENBOT_DYNAMIC_TOOLS`; Claude gets a hand-written in-process MCP server in `claude-client.ts` `#createOpenBotServers`, and that list never declared `update_profile`. The developer instructions promised a tool one provider could not call (confirmed: works on Luna, fails on Sonnet).

- `src/backend/claude-client.ts`: `update_profile` registered in the Claude mirror, delegating to the same `openbot` handler the other providers use. The handler in `agent-service.ts` is provider-agnostic and is unchanged.
- `src/backend/claude-client.test.ts`: the assertion that let this drift through read `_registeredTools` and checked `expect.arrayContaining` against a hand-typed list that never included `update_profile`. It now holds the mirror to the catalogue: the set of names Claude registers must equal `OPENBOT_DYNAMIC_TOOLS` minus a named exclusion set (`ask_user`, which Claude serves through the SDK's `AskUserQuestion`). Set equality, so both a missing tool and a Claude-only tool go red.

Surfaces touched: desktop backend only (the Claude provider tool surface). No IPC contracts, `mock-openbot.ts`, renderer, mobile, `apps/auth-api`, palettes, migrations, reverse states or documentation change.

This is the first of three stacked branches (next: `update_profile` gains `model`/`reasoningEffort` plus `list_models`; then `create_agent`).

## Verification

- [x] Narrow checks for what changed: `biome check` on the two files, `bun run lint`, `bun run typecheck`, and `bun run test:desktop -- src/backend/claude-client.test.ts` — CI owns the full suite
- [x] Watched the test fail before the fix: it went red naming exactly `update_profile` as the missing entry, then green after adding the tool
- [x] Manual smoke test: ask a Sonnet agent to rename itself — not run here, left for the developer

## Risk and security impact

None new. Claude gains reach to an `openbot` tool Codex and Grok already have, routed through the same provider-agnostic handler and the same `INPUT_LIMITS` on name, title and description. No change to permissions, IPC, filesystem access, persistence, browser behavior or network access.

Model: Claude (Cursor agent). Harness: Cursor.

Screenshot 2026-09-07 at 13 05 34


NorbertBodziony:
LGTM

PR #282, opened 2026-09-07
Generate the development env file per checkout

NorbertBodziony:
## What changed

#279 has merged and this is now rebased onto `main`, so the diff below is this change alone.

`apps/auth-api/.env.dev` was committed encrypted, so running the app needed the maintainer's `DOTENV_PRIVATE_KEY_DEV`. `README.md:105` promises clone → install → `bun run dev`, and for anyone outside the org that died at `prepare-dev-environment.ts:48` with `Missing or empty .env.keys`. #273 works around this by encrypting a contributor overlay; this removes the reason for an overlay instead.

Every value in that file is derivable on the machine that runs it, so `scripts/development-secrets.ts` writes one on first run:

| Key | Generated as |
| --- | --- |
| `REMOTE_TICKET_PRIVATE_JWK` / `_PUBLIC_JWKS` | fresh P-256 pair, `kid` = `openbot-remote-1` to match the `REMOTE_TICKET_KEY_ID` that `wrangler.jsonc` pins, since `parseJwks` rejects a JWKS without the active key |
| `SKILLS_ADMIN_TOKEN`, `SITE_REPORT_HASH_SECRET`, `REMOTE_AUTH_WEBHOOK_SECRET` | `randomBytes(32).toString("hex")` — over the 32-byte floor `strongSecret` in `remote/api/src/config.ts` enforces, and the webhook secret is read from this same file by both the Worker and the Signal service, so one value pairs both halves |
| `AUTH_EXPOSE_DEVELOPMENT_CODE` and the five blank SMTP names | from #279 |

The file is now untracked (`git rm --cached`, and the `!apps/auth-api/.env.dev` negation is gone from `.gitignore`), and it is never overwritten — an existing file keeps its secrets, so your encrypted copy goes on working with no migration step. `.worktreeinclude` carries it so a worktree reuses the machine's dev identity instead of generating a second one.

Generation runs before `bun install`, which is why the script imports node builtins only and writes to stdout rather than using `@openbot/logging` — neither resolves on a fresh clone. `dev-services.ts:184` already calls the preparation, so `bun run dev` and `bun run dev:api` both cover it.

The launch commands need no change. I expected to have to strip `-fk ../../.env.keys` for a plaintext file and it turns out `dotenvx run` accepts a missing key path and injects normally, exit 0, including under the `--redact --strict` the `remote` service spec uses. So one command shape serves both your encrypted file and a contributor's generated one.

`env:encrypt:dev` and `env:validate:dev` are removed from both manifests — they operated on a file that is now plaintext and untracked. `env:*:prod` is untouched.

Surfaces: local development setup and its docs (`AGENTS.md`, `apps/auth-api/README.md`). Not the renderer, mobile, IPC contracts, reverse states, migrations, or production.

**After this merges, two maintainer steps I can't do:** drop `DOTENV_PRIVATE_KEY_DEV` from your `.env.keys` (nothing reads it any more), and rotate the credentials named in #279.

## Verification

  - `vitest run scripts/development-secrets.test.ts scripts/prepare-dev-environment.test.ts scripts/dev-services.test.ts scripts/dependency-catalog.test.ts` — 35 passed
  - `bun run lint` (1188 files, the same 5 pre-existing `apps/mobile` module-mock warnings, none in touched files) and `bun run typecheck` (11/11)
  - Watch-it-fail, one break per new assertion, each failing only its own test:
    - `namedCurve: "P-384"` → the ES256 sign/verify round-trip fails
    - key ID `openbot-remote-2` → `expected 'openbot-remote-2' to be 'openbot-remote-1'`
    - `randomBytes(8)` → `expected 16 to be greater than or equal to 32`
    - dropped the `existsSync` guard → `expected 'created' to be 'kept'`
    - dropped a blank SMTP name → `expected [ undefined, '', '', '', '' ] to deeply equal [ '', '', '', '', '' ]`
    - moved generation after the install commands → `expected [ false, false ] to deeply equal [ true, true ]`
- [x] Surfaces touched are named under "What changed": renderer, mobile, `apps/auth-api`, the three `--openbot-*` palettes, IPC contracts and `mock-openbot.ts`, reverse states, migrations and the latest schema, documentation
- [x] Relevant manual smoke test completed, or not applicable — generated a file with the script into a scratch root and booted the Worker off it exactly as `dev:api` does (`dotenvx run -f … -fk ../../.env.keys -- vite dev`) with **no `.env.keys` present at all**, on `OPENBOT_API_PORT=3199` against local D1:
  - `⟐ injected env (11)`, Vite ready — the missing key file is a non-event
  - `GET /.well-known/jwks.json` → serves the generated P-256 key with `kid: openbot-remote-1`, so the ticket signer constructed
  - `POST /v1/auth/email/start` → `200` with `developmentCode`; `/verify` → session; `GET /v1/me` → `200`
  - `POST /v1/team-invitations/email` → `503 email_delivery_not_configured`
  - ran `bun scripts/development-secrets.ts` against this checkout's existing encrypted file → `Kept the existing apps/auth-api/.env.dev.`, file unchanged
- [x] No credentials, private data, generated output, or real user files are included — it deletes the last credential-bearing tracked env file

## Risk and security impact

Lowers the value of `.env.keys` to production only, and removes the last shared development secret. Each machine's dev secrets are now unique, so one leaking compromises one checkout.

No IPC, filesystem, persistence, or renderer-to-main change. Production is untouched: `.env.production` stays encrypted, `env:*:prod` and `api:deploy` are unchanged, and the generated file is only ever read by `vite dev`.

One judgement call worth a reviewer's eye: the generated file is `0o600` plaintext on disk. That is the same exposure as the `.dev.vars` and `remote/.env.keys` files already sitting there, and strictly less than a tracked file decryptable with a key copied into every worktree — but it is plaintext, so if you would rather it be encrypted per machine, say so and I will add that instead.


NorbertBodziony:
Follow-up verification, beyond the unit tests in the description — I had only exercised `vite dev` by hand before, not the real script path or the Signal service.

**The actual `bun run dev:api`, from a genuine fork state** (encrypted `.env.dev` deleted, no `.env.keys` on disk):

```
$ bun scripts/dev-services.ts api
Checked 1295 installs across 1518 packages (no changes) [265.00ms]
$ wrangler d1 migrations apply openbot-auth --local
✅ No migrations to apply!
API port 3100 is busy. Using 3199.
$ dotenvx run -f .env.dev -fk ../../.env.keys -- vite dev
⟐ injected env (11) from .env.dev
VITE v7.3.6 ready in 1992 ms
```

Generated file is `-rw-------`, zero `encrypted:` values. Login → `200` with `developmentCode`, `/verify` → session, `/.well-known/jwks.json` → `kid: openbot-remote-1`, invite → `503 email_delivery_not_configured`.

**The Signal service, which loads the same file under `--redact --strict`** — this was the real gap, since `remote/api/src/config.ts` has its own required-config checks that the generated file has to satisfy:

```
⟐ injected env (11) from apps/auth-api/.env.dev
OpenBot Remote API is ready at http://127.0.0.1:8195
```

`/health/ready` and `/health/live` both `200`. `REMOTE_TICKET_PUBLIC_JWKS`, `REMOTE_AUTH_WEBHOOK_SECRET` (64 bytes, over the 32-byte `strongSecret` floor) and the dev-services defaults for `REMOTE_CONTROL_PLANE_URL`, `REMOTE_SESSION_SECRET`, `TURN_SHARED_SECRET` and `TURN_HOST` cover every `required()` field. Both services read the ticket key pair from the one file, so the signer and verifier agree by construction, and the sign/verify round-trip is unit-tested.

**One gap this run exposed, fixed in 0c57d7ff:** generation was silent. A first `bun run dev` wrote secrets with no output at all. `prepareDevelopmentEnvironment` now returns the outcome and both entry points report a fresh file — `dev-services.ts` through its logger, the standalone script through stdout since it runs before `bun install`. Watch-it-fail: forcing the return to `"kept"` gives `expected 'kept' to be 'created'`.

Afterwards the encrypted `.env.dev` was restored (6 `encrypted:` values, `DOTENV_PUBLIC_KEY_DEV` intact) and ports 3199/8195/8196 released. Another session's Signal service on 8090 was left alone.

PR #297, opened 2026-09-07
feat(dev): make worktree dev stacks and ports conflict-free

NorbertBodziony:
## Why

Several worktrees run dev on this machine at once, and the runner shared one set of default ports and one profile with all of them. Port selection was a `bind` probe followed by the real listen seconds later, with nothing written down in between: two `bun run dev` calls started together both saw 5173 free, both claimed it, and both claimed the unsuffixed `OpenBot Dev` profile with it — two Electron clients writing one SQLite file. Stopping a stack meant `pkill -f electron` or `pkill -f bun`, which kills the other worktrees' work mid-write, and `storybook dev -p 6006` in a second worktree silently showed the first worktree's stories.

## What changed

- **The allocation lock.** `scripts/dev-automation/port-allocation.ts` puts read → choose → publish behind one machine-wide lock file (`wx` creation, break on a dead or unparseable or stale holder, release only a lock this process still holds). The critical section is a few probes and one file write; everything slow happens after the release.
- **The stack registry.** `scripts/dev-automation/stack-registry.ts` records every port and pid a dev stack holds, Storybook included. `findAvailablePort` now skips a published port **without probing it** — a stack that has won 5173 has not bound it yet, so a probe would call it free and hand it out twice.
- **`bun run dev:status` and `bun run dev:stop`** (`scripts/dev-stack.ts`) replace the `pkill` patterns. Status prints every stack and instance as JSON, with `worktree`, `orphaned` and per-process `live` flags. Stop defaults to this worktree; `--pid=<supervisor pid>` and `--all` are the explicit ways to touch anyone else. A recorded supervisor is signalled as a *process* and the detached children as *groups*: the supervisor shares its group with the `bun run` that started it, so a group signal goes to that job and fails with ESRCH, which `isUnavailableProcess` reads as "already stopped".
- **Storybook** goes through the same allocator with `--exact-port`, so the port it announces is the port it uses. `--port` is refused, `OPENBOT_STORYBOOK_PORT` moves where the search starts.
- **A second stack in one worktree is refused** on service overlap unless `--force`. `bun run dev --isolated` gives a profile keyed to the worktree path (`developmentInstanceIdForWorktree`), instead of a suffix that moves with whichever renderer port the worktree won that morning.
- **Orphans stay visible.** A record lives while anything it started lives, so a SIGKILLed supervisor does not hand its children's ports to the next worktree; `dev:status` reports them as `orphaned` and `dev:stop` clears them.
- `registry-files.ts` holds the owner-only directory checks, the atomic staged write and the recycled-pid liveness check that both registries need, extracted from `instance-registry.ts` rather than duplicated.

## Surfaces touched

Dev tooling (`scripts/`), the desktop main process (one pure function in `src/main/development-profile.ts` — no IPC, contract, renderer or migration change), and documentation (README command table plus the runner paragraph, AGENTS.md "three ways to hurt yourself" items 2 and 3, `docs/ARCHITECTURE.md` change rule 10). No renderer, mobile, `apps/auth-api` or `remote/api` change.

## Tests

New files for the two boundaries that did not exist before: `scripts/dev-automation/stack-registry.test.ts` (parse, round trip, prune, liveness, orphan, held ports, conflict scope) and `scripts/dev-automation/port-allocation.test.ts` (mutual exclusion, stale and half-written locks, release on failure). Cases added to `scripts/dev-services.test.ts` and `src/main/development-profile.test.ts`.

**Watched each one fail.** 11 mutations, all red for the intended reason:

| Break | Failure |
| --- | --- |
| `heldPorts` ignored in `findAvailablePort` | `expected 5173 to be 5175` |
| `signal-health` dropped from the record | 6 ports instead of 7 |
| supervisor signalled as a group | `expected calls [321, 'SIGTERM']` |
| parser accepts a privileged port | `expected {…} to be null` |
| stack dies with its supervisor | `expected false to be true` |
| conflicts ignore the worktree | `expected [{…}] to deeply equal []` |
| held ports cover only the first stack | `Set{3100,3101,5173}` vs `Set{…,3110}` |
| unparseable record left in place | `expected ['stack-4242.json'] to deeply equal []` |
| lock opened `w` instead of `wx` | `first:enter, second:enter, first:publish` |
| release deletes whichever lock is there | `expected false to be true` |
| 6-digit worktree instance id | `expected null to be '140120'` |

No test waits on the clock: the mutual-exclusion test synchronizes on the second allocator's `onWait` callback, so the first one publishes only after the second has met the lock.

## Verification

- `bun run test:desktop -- scripts/dev-services.test.ts scripts/dev-automation/stack-registry.test.ts scripts/dev-automation/port-allocation.test.ts scripts/dev-automation/instance-registry.test.ts src/main/development-profile.test.ts` → 79 passed.
- `bun run lint` → 8 warnings, all pre-existing `no-module-mocking` in `apps/mobile`. Nothing under `scripts/` or `src/`.
- `bun run typecheck` → 12 projects clean.
- Live check with `bun run dev:api`: the stack published its pid and port 3100, a second `dev:api` was refused with the running stack printed, `dev:stop` exited it with SIGTERM (143), 3100 was released and the registry went empty.

No `biome-ignore`, `@ts-expect-error` or widened type, so nothing here needs a named approvability reason.

Model: Claude Opus 5. Harness: Claude Code in a git worktree.


NorbertBodziony:
Fixed the four review findings. Each one was valid, and the third only in one half.

**1. The lock could be read while it held nothing (P1).** It was created with `wx` and written
afterwards, so a waiter that read it in that window saw an anonymous file, called it abandoned and
broke it - two allocators in the critical section, the exact collision the lock exists to stop. The
holder now goes into a staging file that is published with `link`: atomic, and EEXIST when the lock
is taken. Two related arms came with it - an unreadable lock is breakable only once it has also sat
there past the stale window, and an allocator that runs its wait out beside a live holder now fails
with that holder's pid rather than taking the lock.

**2. `dev:stop` signalled on `kill(pid, 0)` alone (P1).** That says a pid is taken, not by what, and
pids are recycled. `verifyRecordedProcess` now answers `live`, `gone` or `unverified`, and
`unverified` - a pid this machine cannot date - is not permission to signal. Anything the command
refuses to touch is reported, left running and kept in the registry, and the command exits non-zero;
`bun run dev:forget` drops such a record once the developer has dealt with the process. Windows gets
a start time too, from PowerShell.

**3. Half valid.** The blind group signal was real, so a process group that outlived its recorded
leader is now named in the report instead of being signalled - the pgid belongs to nobody at that
point. The retention half was also real and the opposite way round: liveness now counts a surviving
group, because electron-vite exits and the Electron it started keeps the renderer port, and pruning
the record there frees a port that is still bound.

**4. Scope (P2).** `dev:status` reports every stack, since the one holding the port this worktree
wanted is a sibling's. `stop` and `forget` stay on this worktree unless `--pid=` or `--all` names
another.

Tests: five new cases and one changed. Each was watched failing for its own reason before the fix
went back in - `verifyRecordedProcess` in its three states (mutating each arm fails exactly one
case), the two unreadable-lock cases and the live-holder case (breaking the age check or the
deadline throw fails them), the surviving process group (`sh -c 'sleep 30 &'`, a real leader that
exits while its group holds on - it fails when the group check is dropped), and the command scope
(widening `forget` or narrowing `status` fails one case each). No test waits on a clock.

Also verified by hand against planted records on this machine: a live-leader group is stopped and
its record removed with exit 0; a group whose leader is gone is reported, kept and exits 1, and
`dev:forget` then clears it. `bun run lint`, `bun run typecheck` and the four narrow test files pass.

/norbiai review

NorbertBodziony:
The remaining P1 is fixed, both halves of it.

**Recovery is serialized.** `unlink` cannot be made conditional on what the path holds, so two
waiters that read one dead holder both deleted - the first published a lock of its own and started
allocating, the second deleted that replacement and started beside it. `breakStaleLock` now takes a
second file first, and re-reads the lock *under* it before deleting anything, so a waiter that
arrives late finds the replacement and waits for it instead. Nothing ever breaks that second file:
it is held across a few syscalls with no `await` between them, and a leak costs recovery rather than
correctness - allocators wait and then fail with a message naming the file to remove. Recovering it
automatically would need the same serialization one level down.

**A live holder's lock is never taken, at any age.** The `now - acquiredAt > staleMs` arm is gone.
An allocation is milliseconds of work, so a lock held for a minute means something is wrong, but
wrong is not finished - the holder may be stopped in a debugger and about to step into the critical
section - so the developer gets an error naming the pid.

That arm existed for a real case, though, which is now covered by evidence rather than by patience:
`isBreakableLock` dates the holder with `verifyRecordedProcess`, so a pid recycled since the lock was
taken reads as `gone` and is broken, while the process that actually took the lock reads as `live`
however long it holds. Without that, one recycled pid would wedge every dev start on the machine
until somebody removed the file by hand.

Tests: four new cases, each watched failing for its own reason - ignoring the re-read fails "leaves
the replacement", proceeding while the second file is taken fails "waits its turn", restoring the
age arm fails "never takes the lock from a live holder", and dropping the start-time check for a
plain `isProcessAlive` fails "breaks a lock whose holder pid has since been recycled". The
live-holder case runs twice, once with an hour of injected clock, so age cannot be what decides it.

Also checked by hand on this machine: a lock left by a dead pid is broken and allocation proceeds; a
lock left by a dead pid with the recovery file present waits and then fails with the message naming
that file; a normal allocation leaves the registry directory empty. `bun run lint`, `bun run
typecheck` and the four narrow test files pass.

/norbiai review

NorbertBodziony:
Round 3 P1 fixed: `unlink` cannot be made conditional on what the path holds, so the break lock serialized recoverers against each other but not against a plain acquisition. A recoverer that read a missing path (`age === null`) could delete a lock somebody else published in the meantime, and both would allocate.

So the lock no longer has one path. Ownership is a **generation**: taking it means creating `port-allocation.<n+1>.lock` with an exclusive create (staging write + `link`), so the only step that decides who owns it is one the kernel makes atomic, and nothing is ever taken by deleting. A lock whose holder is gone is *superseded*, not removed - the judgement about a dead holder decides only when to move on, never who won. The holder of the highest generation owns it; lower ones are litter the winner clears. `breakStaleLock` and the break lock are gone.

Unchanged: a live holder is never moved past however long it has held it, `unverified` counts as live, a recycled pid counts as gone, and an unreadable file is waited out until the stale window passes.

**Tests.** Ten in `port-allocation.test.ts`, and I watched every one of them fail for its own reason (nine mutations of `port-allocation.ts`, restored byte-for-byte after each):

- exclusive create no longer excludes (EEXIST read as a win) -> the multi-process test printed `enter leave enter leave enter enter`
- a gone holder is never recovered -> the two supersede tests hang out their wait and throw
- an unreadable lock judged litter at once / never -> the two unreadable-lock tests
- age decides instead of the holder's liveness -> "never takes the lock from a live holder"
- no `finally` -> "releases the lock when allocation fails" leaves `port-allocation.1.lock`
- release without the identity check -> "leaves a lock a waiter took over"
- no `discardSupersededLocks` -> three tests on leftovers
- every lock judged abandoned -> the two ordering tests

The new one is multi-process on purpose. Everything from reading the registry to publishing the lock is synchronous, so two allocators *in one process* never interleave there - which is exactly the window this finding lives in. It spawns four `bun` processes against a planted dead-holder lock, so each has to recover before it can allocate, and asserts the enter/leave log strictly alternates.

**Live checks** against the real registry directory, `bun scripts/dev-services.ts --dry-run`: a normal run allocated and left no lock behind; a planted generation-7 lock from a dead pid was superseded (and the litter cleared); a lock held by a live `sleep` was waited out and then refused by name, with the file untouched.

`bun run lint` (8 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, and `scripts/dev-stack.test.ts scripts/dev-services.test.ts scripts/dev-automation/` (115 tests) pass. `docs/ARCHITECTURE.md` now describes the generation scheme instead of the break lock.

/norbiai review

NorbertBodziony:
Round 4 P1 valid, and my previous argument was wrong where you said it was: exclusive create only arbitrates attempts that ask for the *same* generation, and `discardSupersededLocks` let the numbering regress under a plan already made against it. Your interleaving is exact - recover 1, take 2, release 2, C finds an empty directory and takes 1, A resumes with its stale plan for 2, creates it, and sweeps C's live lock away.

The rule is now stated in the file and is the one thing to keep when changing it: **a lock file is removed by nothing but its own live holder, releasing it.** The sweep is gone. A recovered lock is superseded and left standing, which is what pins the numbering - generation 1 stays there, so nobody is handed 1 again.

Release still deletes our own file, and that one case cannot regress the counter: an allocator that read our lock as the highest generation read it as *live*, because we are running, so it waited rather than planning a generation above it. There is no plan our removal could let back down. Anything that is not ours is left alone, as before.

The price is a small file left in the per-user temporary directory per crashed dev start. Nothing reads it - the highest generation is the lock - and leaving it is what makes the numbering monotone.

**Tests.** The invariant is now what the supersede test asserts: inside the critical section the directory holds `["port-allocation.1.lock", "port-allocation.2.lock"]`, and after release just `port-allocation.1.lock`. I watched it fail with the sweep put back - `expected [ 'port-allocation.2.lock' ] to deeply equal [ 'port-allocation.1.lock', ... ]`, plus the two other leftover assertions - and watched "leaves a lock a waiter took over" fail with the identity check removed from release. All 10 in the file pass; the multi-process test now expects the planted abandoned lock to be the one leftover.

Live again with `bun scripts/dev-services.ts --dry-run`: a normal run leaves the registry directory empty, and with a dead holder's lock planted at generation 3, allocation proceeds and generation 3 is still there afterwards while ours is released.

`bun run lint` (8 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, and the narrow suites (115 tests) pass. `docs/ARCHITECTURE.md` states the invariant.

/norbiai review

NorbertBodziony:
Round 5 P1 valid, and it kills delete-on-release too. The step I was reasoning about was the wrong one: the file does not have to be gone when another allocator *lists* it, only when it reads its contents. Listing `1.lock`, then reading nothing there, gives a null holder and a null age, so it is judged abandoned and generation 2 is planned - while generation 1, freed by the delete, goes to the next arrival. Your interleaving is exact.

**No lock file is ever deleted now.** Releasing renames it to `port-allocation.<n>.released`, which gives up the claim in one step while the number stays spoken for, and `lockGeneration` counts both suffixes towards the numbering. A lock a crashed holder left behind is superseded where it lies. That is now the rule at the top of the file: a generation is used once and stays used, because the scan, the read and the create are separate steps and a file that goes away between any two of them takes its number out of the count.

Litter is the trade, and it is a small one: tens of bytes per allocation in the per-user temporary directory, which the system clears. The two registry readers beside it (`stack-registry.ts`, `instance-registry.ts`) both filter `.json`, so nothing else sees these files - checked, and `dev:status` still reports correctly with one present.

**Tests.** A new one pins exactly this: three allocations in a row must take generations 1, 2 and 3, and end with three `.released` files and no `.lock`. "releases the lock when allocation fails" now goes on to assert the next attempt gets in *at generation 2*. Watched both fail two ways - with release deleting the file again, and with `.released` files excluded from the numbering - and the reason in each case is the one that matters: `expected 'port-allocation.1.lock' to be 'port-allocation.2.lock'`, generation 1 handed out twice. 11 tests in the file pass.

Live: two `bun scripts/dev-services.ts --dry-run` runs in a row left `port-allocation.1.released` and `port-allocation.2.released` and no held lock, so the second took the generation above the first.

`bun run lint` (8 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, and the narrow suites (116 tests) pass. `docs/ARCHITECTURE.md` matches.

/norbiai review

NorbertBodziony:
Round 6 P1 valid, and the distinction I missed is the right one: renaming kept the *generation* spoken for but freed the *path*, and the path is what `link` arbitrates over. Your interleaving is exact - A plans 2, B takes 2 and renames it away, C takes 3, A creates `2.lock` into the hole and enters beside C.

The rule is now the one you named: **once a lock path exists it stays, and nothing ever frees it.** A lock's state lives in its contents, not in its name. Releasing writes `{"released":true,...}` to a staging file and renames *that* onto our own lock path, which replaces the contents in one step and leaves the path occupied on both sides of it. Reading a lock is a tri-state - claimed by a holder, released, or unreadable - and an unreadable one is left alone for the stale window to recover, because that path has to stay occupied too.

That closes all three ways in, and they are written at the top of the file so the next change does not reopen one: deleting the superseded files, deleting our own on release, and renaming ours aside. All three free a number for a second asker while a plan made against it is in flight.

**Tests.** The generation-reuse test now records the number it holds, so the failure names the defect directly. Watched it fail two ways - release renaming the path aside, and release deleting it - both `expected '1,1,1,1' to be '1,2,3,4'`. Also watched the released marker being unrecognised wedge four tests on the wait. 11 pass.

Live, `bun scripts/dev-services.ts --dry-run` three times in a row:

```
port-allocation.1.lock: {"released":true,"pid":82328,...}
port-allocation.2.lock: {"released":true,"pid":82335,...}
port-allocation.3.lock: {"released":true,"pid":82341,...}
```

and with a lock left claimed by a dead pid planted at generation 9, the run took 10 and left 9 exactly as it was. `dev:status` still reports correctly - both registry readers filter `.json`.

`bun run lint` (8 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, and the narrow suites (116 tests) pass. `docs/ARCHITECTURE.md` matches.

/norbiai review

NorbertBodziony:
Fixed the new [P2] on `stack-registry.ts:209`. It was valid, and it is the same read-check-use shape as the four lock rounds: `readDevStackRecords` read a record, judged it dead, then `rmSync`ed a path whose contents could have changed in between. `readDevInstanceRecords` had the identical shape.

**Both reads now only filter.** A record whose processes are gone is left out of every read, so its ports are free from that moment, but no reader deletes a file. Two reasons, and the second needs no race at all:

- The one reported: the supervisor publishes a detached child at `dev-services.ts:439` and is killed a moment later, and a reader holding the pre-publish view deletes the version *with* the child. The child keeps the ports and nothing records them.
- A record this checkout cannot parse is a sibling worktree on a newer branch, and its stack is live.

**Removal is now a decision, not a guess.** `bun run dev:forget` reads past the liveness filter (`readAllDevStackRecords` / `readAllDevInstanceRecords`) so it reaches records no other command reports; `status` and `stop` still ask what is running. It does not read past the *parser*, so a newer sibling's record stays - such a file reserves no port and hides no stack, so it is inert rather than dangerous.

**Watched the tests fail.** M17 put the read-time `rmSync` back in both readers: the three changed assertions went red for the intended reason - `expected [] to deeply equal [ 'stack-4242.json' ]` twice, and `expected [ 'app-1111.json' ] to deeply equal [ 'app-1111.json', 'app-2222.json' ]`. M18 dropped the liveness filter to check the ports half still bites: `drops a record no live process backs` went red as `expected [ { services: [ 'api', …(2) ], …(5) } ] to deeply equal []`. Both restored byte-for-byte; 117 tests pass across `dev-stack`, `dev-services` and `dev-automation/`.

**Live-checked on this machine**, real registry, empty before and after. A planted dead record: `dev:status` reported `"stacks": []` and left the file; `bun scripts/dev-services.ts --dry-run` offered 3100, the port that record claims, and touched nothing; `dev:forget --pid=` removed it and exited 0.

Docs updated: the `dev:forget` row in the README command table, the process-killing paragraph in `AGENTS.md`, the dev-stack paragraph in `docs/ARCHITECTURE.md`, and the now-inverted comments over `readDevInstanceRecords` and `forgetDevStack`.

/norbiai review

NorbertBodziony:
Fixed the new [P2] on `scripts/dev-stack.ts:166`. It was valid, and it is a consequence of the previous fix: a dead stack record now waits on disk for `dev:forget` instead of being pruned by the next read, so the pids it names have had every chance to be recycled by the time anything acts on them. An instance record lives at `<service>-<pid>.json`, so the app that takes a pid writes over the record of the app that had it, and `forgetDevStack` matched on pid alone.

**`claimOnDevInstance` now answers whose the record is** - `ours`, `another` or `unverified`. Each instance record dates itself: its `startedAt` is when *that* instance published, so `verifyRecordedProcess` holds it against the start time the kernel reports for the pid, and a live instance is never a dead stack's, whatever pid it holds. The worktree is checked too, because it is the cheaper half of the same question and it still answers on a machine that cannot date a pid at all. A record of *this* worktree that cannot be dated is kept and reported, not guessed at.

**On the suggested shared timestamp**: I did not carry one from the supervisor into the instance record. The supervisor's `startedAt` for the `app` child dates the electron-vite spawn; the instance record's dates the Electron main process publishing. They are two different events, so making them equal would mean plumbing one through electron-vite and would date the record to something that is not when it was written. The record's own self-consistent generation is already exact, and `verifyRecordedProcess` is the check the rest of the registry uses.

**Watched every guard fail.** Four mutations, one per guard, each restored byte-for-byte:

- M19 - pid alone, exactly the reported bug: three tests red, `expected 'ours' to be 'another'` twice and `expected 'ours' to be 'unverified'`.
- M20 - dropped the worktree check: `leaves another worktree's record even where nothing can date the pid` red, `expected 'unverified' to be 'another'`.
- M21 - treated `unverified` as ours: `keeps a record of this worktree it cannot date, rather than guess` red.
- M22 - dropped the pid check: `leaves a pid this stack never started` red.

122 tests pass across `dev-stack`, `dev-services` and `dev-automation/`; `lint` clean apart from the 8 pre-existing mobile `no-module-mocking` warnings; `typecheck` clean.

**Live-checked the real command.** A planted dead stack record naming two of its processes, plus two instance records: one at the pid of a live process with another worktree's `projectRoot`, one at a dead pid with this worktree's. `dev:forget --pid=` deleted the second, kept the first and reported it under `Still published`. Registry clean before and after.

Also noted in the `docs/ARCHITECTURE.md` dev-stack paragraph.

/norbiai review

PR #306, opened 2026-09-07
fix: support MP3 and MOV chat attachments

### LINKED ISSUE #183: Support MP3 and MOV file attachments
## Description

The chat attachment flow currently rejects audio and video files such as `.mp3` and `.mov` with:

`Error invoking remote method 'agent:import-attachments': Error: Screen Recording 2026-09-03 at 11.18.16.mov is not supported. Attach images, PDF, Office documents, text, Markdown, data, or source files.`

This prevents users from attaching common audio and video recordings for analysis or discussion.

This is related to [#114](https://github.com/nightly-labs/openbot/issues/114), which covers other currently unsupported attachment formats such as `.eml` and `.zip`.

## Expected behavior

`.mp3` and `.mov` files should be accepted by the attachment flow and made available to the agent, or the UI should clearly communicate any intentional limitations and supported alternatives.

## Checklist

- [ ] Add `.mp3` support to the attachment import flow.
- [ ] Add `.mov` support to the attachment import flow.
- [ ] Define safe file-size and processing limits for audio/video attachments.
- [ ] Ensure the advertised supported-file list matches the actual backend capability.
- [ ] Show a precise, actionable error for unsupported media formats.
- [ ] Add tests for successful imports, malformed files, oversized files, and unsupported formats.
- [ ] Verify attaching MP3 and MOV files through the desktop chat UI.

NorbertBodziony:
## What changed

Fixes #183. MP3 and MOV recordings can now be attached through the desktop file picker and import flow. They keep their original bytes and use file attachments with no inline preview. The shared supported-file list, MIME types, and unsupported-format messages now match this behavior. Older remote hosts show an update message and do not advertise these formats in the picker.

The existing limits remain: 100 MB per file, 250 MB per message, and 10 files per message. Import does not decode or transcribe recordings. Damaged recordings can be attached for inspection by the agent.

Surfaces: desktop renderer and main-process attachment handling, shared attachment contracts and optional Team API capability, backend storage, README, and architecture documentation. No database migration, IPC endpoint, renderer mock API, mobile, public web, or Signal service change is required. Attachment removal was also verified.

## Verification

- [x] 48 focused tests passed across mailbox storage, attachment formats, and protocol capabilities.
- [x] `bun run lint`, `bun run typecheck`, and `bun run check:ui` passed. Lint reports eight existing mobile module-mock warnings; these changes add none.
- [x] Temporarily disabled media import and confirmed both new MP3/MOV import tests failed because media was rejected; restored the fix and confirmed they passed. Also removed the media capability filter temporarily and confirmed its regression test failed because MP3 was still offered to an unsupported host, then restored the filter.
- [x] Tested byte preservation after storage and reload, damaged media, MIME overrides, per-file and combined size limits, and unsupported-format errors.
- [x] Imported MP3 and MOV fixtures through the desktop composer in an isolated development profile, then removed both attachments.
- [x] No credentials, private data, generated output, or real user files are included.

Before: MP3 and MOV imports failed with “is not supported.” After: both appear in the composer with their file names and sizes, and can be removed. Both the composer HTML picker and the native picker offer media only when the selected host supports it. There is no visual layout change.

Model: GPT-6. Harness: Codex desktop, Vitest, and the repository's development automation/CDP connection to the isolated Electron app.

## Risk and security impact

This adds two file extensions and an optional `media-attachments` host capability. Released protocol adapters retain their existing meanings. The existing file storage and size limits apply. Media MIME types come from the file extension, so a supplied image or text MIME type cannot enable a preview. Import runs no media codec. No permission, sender-validation, database schema, or network destination changes.


NorbertBodziony:
Fixed the P2 picker finding in cc40a023. `ConversationComposer` now derives its HTML input's `accept` list from `props.server` and its advertised EML/media capabilities. Remote hosts with no capability metadata do not offer these formats. The composer and native picker use the same `supportedAttachmentExtensions` function, so ordinary supported files remain available and the lists cannot diverge through separate filters.

All 48 focused tests, lint, type checks, and UI checks pass. Removing the media filter made the new regression test fail because MP3 was still offered to an unsupported host; restoring it made the test pass. The push starts a fresh CI and code review run.

PR #321, opened 2026-09-08
Add agent analytics and host-wide Usage reports

### LINKED ISSUE #293: Add per-bot usage analytics and cost breakdown
## Summary

Add usage analytics for individual bots so users can understand activity, token consumption, model/provider usage, and estimated costs per bot.

## Reference

Usage analytics reference

## Requirements

- Add a usage analytics view accessible from each bot's details or settings.
- Show the selected bot's sessions/messages, processed tokens, cached and uncached input, output tokens, and estimated cost.
- Support configurable time ranges, including at least 7, 30, and 90 days, plus a custom range where practical.
- Display daily usage and cost trends in a chart.
- Provide a breakdown by connected model/provider, including share, tokens, and cost.
- Make the selected bot and time range clear, and support loading, empty, and error states.
- Ensure analytics are isolated per account and bot and do not expose prompt content or secret values.
- Keep the view usable on desktop and mobile, with accessible charts and labels.

## Acceptance criteria

- A user can open analytics for a bot and see only data belonging to that bot.
- Totals, daily trends, and model/provider breakdowns update when the time range changes.
- The numbers and estimated cost are consistent with the connected model's usage data.
- The UI handles missing data, delayed data, and provider errors without breaking.
- Analytics do not include message contents, credentials, or other sensitive values.
- Automated tests cover aggregation, filtering by bot and date range, model/provider breakdowns, and key UI states.

NorbertBodziony:
## What changed

Add local usage analytics for agents and a desktop report for all agents on a selected host. Right-click a server and select Usage to see cost estimates, tokens, daily trends, and model totals. The agent settings shortcut opens the same report with that agent selected. Back keeps the prior workspace and conversation draft.

- Collect provider usage from installation onward in local SQLite, with persisted checkpoints and separate token categories. Keep analytics after a conversation is cleared; remove them when its agent is deleted.
- Add separate agent and host analytics requests, capabilities, Team API routes, HTTP/WebRTC adapters, preload methods, shared client decoding, and populated preview mocks. Existing account-limit and per-agent contracts keep their meaning.
- Use a minimal desktop report with Zaidan area charts, exact daily values, filters, and partial, unavailable, loading, error, and unsupported states. Patch `solid-recharts` for the existing Solid 2 runtime and retain MIT notices.
- Include the earlier per-agent mobile Usage view. The host-wide redesign is desktop-only.

Surfaces touched: desktop renderer, mobile, backend collection and SQLite migration/latest schema, IPC and Team API contracts, preload, team client, preview mocks, shared brand token, Storybook, architecture and privacy documentation. Public web and Signal service storage are unchanged.

Closes #293.

## Verification

- [x] Focused usage aggregation, protocol, client decoding, IPC channel coverage, authenticated route, renderer, and navigation tests passed.
- [x] Behavioral mutation checks failed as expected for session identity collisions, host scope validation, agent filtering, and keyboard menu access. Restored code passed.
- [x] `bun run lint` passed with 8 existing warnings.
- [x] `bun run typecheck` and `bun run check:ui` passed. Commit hooks also passed.
- [x] Checked full-app wide and narrow layouts, chart rendering, exact tooltip values, keyboard access, daily table focus, and Back navigation in Storybook.
- [x] `git merge-tree` reports no conflicts with the current `main`.

Long suites and production builds are left to CI. No mobile device run was performed. Images use preview data.

Model/harness: GPT-6 in Codex desktop; Vitest and the Codex in-app browser against local Storybook.

## Risk and security impact

This change adds an irreversible SQLite migration for local analytics and updates the separate schema for new databases. Usage records contain numeric counters, internal identifiers, timestamps, provider/model names, and estimate rate information. They do not store prompts, responses, credentials, or raw provider payloads. Authenticated host team members can read aggregate usage through the existing host membership boundary. Requests have explicit validation and capabilities; older hosts return Unsupported.

The Solid 2 chart dependency patch is a maintenance cost and needs review when the chart package changes. The patch retains the upstream licence and uses the existing application runtime.

## Screenshots

Before opening Usage (preview workspace):

Before

After opening Host Usage:

Host Usage

Narrow window:

Narrow Host Usage

---

## Follow-up in this branch

The report above drew one area for the day's total, so a Claude spike and a flat Codex week arrived as one hump. Six changes since:

- **One area per provider.** `HostAnalytics` gains a required `providerDaily` array - a lean cell of date, provider, processed tokens and cost estimate, built lazily in the same SQLite pass, so only days with traffic are sent. It is host-only, which keeps the agent codec and the mobile screen out of the diff. The field is required rather than optional because `54e0d7c8` is in no tag: no shipped host omits it, and one decoder then fails closed for the local, HTTP and WebRTC hosts at once.
- **A validated series palette.** `--openbot-provider-codex` and `--openbot-provider-grok` were both `var(--openbot-text-primary)`, so two series would have drawn in the same white. Four `--openbot-chart-series-*` tokens instead, passing `scripts/validate_palette.js` for CVD separation, the normal-vision floor and canvas contrast. Colour follows the provider, never its rank, so switching Cost to Tokens reorders the list without repainting the chart. The row dots are the legend.
- **An icon on the server menu's Usage item**, beside its neighbour, through the `components/ui` icon barrel.
- **Compact Totals.** One line that never wraps. The entries are a tail of decreasing interest, so a panel too narrow to hold them all hides the last ones rather than rewrapping into a second row that read as a more important group. Note that a hidden entry is `display: none`, so it leaves the accessibility tree too.
- **Denser Breakdown rows**, and the provider name moved off a second line into screen-reader-only text - the logo already carries it for a reader who sees it.
- **The Back arrow aligned with the header inset.** Its 32px hit area started at the inset, putting the 16px glyph 8px inside it; the button now hangs that padding back into the header. The hit area is unchanged.

Two CI fixes came with the merge from `main`:

- `/v1/analytics` had no `ROUTE_METHODS` entry, so the route-table round trip never requested the one path this PR adds.
- `agent-service.restart.test.ts` fakes an older database by deleting rows from `schema_migrations` without undoing the schema. Every migration before 15 rewrites rows, so replaying one was harmless; 15 is the first that creates tables. It now spells its CREATEs with `IF NOT EXISTS`, as the baseline schema already does, and the test drops every version from 14 up - keeping 15 while dropping 14 leaves a gap the history check rejects.

### Review findings, fixed in this branch

Seven findings from the review job, each verified against the code before the fix.

- **Read state.** An agent reply and a direct message that arrive while the report covers the conversation are no longer marked read, and neither is one waiting there when the window regains focus. All three paths asked only whether the window was focused; the report covers the workspace content and marks it inert, so a message under it was never seen. `conversationVisible` is one predicate on the direct-messages context, used by its own six automatic reads and by the focus listener in `server-scope.tsx`.
- **The keyboard.** `inert` does not reach a document-level listener, and the conversation stays mounted behind the report. Escape with a queued message under edit called `cancelQueuedMessageEdit`, which restored the queued text and discarded the attachments added since; the chat-search shortcut opened a search behind the report. Both now go through one `whenVisible` wrapper keyed on a new `workspaceCovered` prop.
- **The embedded browser** is a separate Electron view, so hiding renderer content left it painted over the report. Usage now suspends it for the same reason a server switch does, and Back restores it because the predicate is derived rather than set. That also covers `closeActiveRemoteBrowserTab`, which already returned early on the same flag.
- **The host-switch effect** moved from `WorkspaceShell` into the Usage provider. `ServerScopeBoundary` is keyed on the active server and rebuilds the shell, so the shell copy initialized its "previous" server to the new one on every mount and could never fire.
- **`analyticsDate`** caches one `Intl.DateTimeFormat` per time zone. A report walks every usage and activity row through it, so a host with 100,000 records built 100,000 formatters synchronously in the main process.
- **The chart's zeroes.** A day with no cell for a provider is a real zero, but a cell whose cost estimate is `null` is an unavailable day. The pivot collapsed both to zero, which contradicts `connectNulls={false}` and the plan it came from.
- **The analytics reads scanned all retained history.** `(? IS NULL OR agent_id = ?) AND occurred_at >= ?` cannot be answered from an index, and the only indexes led with `agent_id`, which a host-wide range over `occurred_at` cannot use either - so a seven-day report cost as much as the whole table, repeated after each turn completion while the report was open. `readHost` now prepares one statement per shape, and migration 16 adds the date-leading `agent_usage_occurred` and `agent_usage_activity_occurred`. `EXPLAIN QUERY PLAN`, in a throwaway `node:sqlite` probe:

  | shape | before | after |
  | --- | --- | --- |
  | host-wide | `SCAN agent_usage_records` | `SEARCH ... USING INDEX agent_usage_occurred` |
  | one agent | `SCAN agent_usage_records` | `SEARCH ... USING INDEX agent_usage_date` |

  Migration 16 rather than an edit to 15, because a database that already ran 15 - every development profile on this machine - would never meet an index added to 15.

Four tests, each watched fail:

| Test | Mutation | Failure |
| --- | --- | --- |
| `keeps an agent reply unread while the Usage report covers the conversation` | drop the guard from `isAgentChatOpen` | "expected vi.fn() to not be called at all, but actually been called 1 times" |
| `keeps a private message unread when focus returns to the Usage report` | drop the guard from the focus listener | "Unable to find role=status and name '1 new message'" |
| `keeps a queued-message edit through Escape while the Usage report covers the conversation` | unwrap the Escape listener | "Unable to find role=button and name 'Save queued message'" |
| `follows the host switch with an open Usage report` | make the effect body unreachable | "Unable to find role=heading and name /Usage.*Studio Mac/" |

The migration case in `agent-usage.test.ts` also asserts both new indexes exist after the upgrade; removing migration 16 fails it with "expected [] to deeply equal [ ...(2) ]". The parity test cannot carry that on its own - it compares the two build paths to each other, so an index in neither still passes. The CI Tests job then caught seven cases in `openbot-database.test.ts` that name the applied versions one by one; they carry 16 now, and the downgrade case inserts `MAX(version) + 1` rather than the literal.

### Navigation out of the report

The rail and the sidebar sit outside the markup the report covers - that is how the report is opened
and how a host is switched. Three commands there open something the report then hides, and two of
them also read it: Command+K reaches the document, so a global search result opened a conversation
behind the report and `openAgentMessage` marked every message through the match; picking a person
opened a private conversation behind it and read that; Create new agent put its form behind it.

`selectAgent`, `selectDirectMember` and `openBotSetup` now dismiss the report first. Dismiss rather
than Back: Back returns focus to the element the report was opened from, and here the destination
owns focus - the transcript scrolls to the message the user picked. `selectDirectMember` dismisses
past its validity guards, so a call that opens nothing leaves the report alone; `openBotSetup`
dismisses *before* its `agentSetupOpen()` guard, so a second press reveals a form that is already
open instead of doing nothing visible, and the guard still protects the draft. The automatic open on
a server with no agents deliberately does not dismiss: it must not take the report from a user who
asked for it.

Two tests in `App.read-state.test.tsx`, four mutations, each watched fail:

| Mutation | Failure |
| --- | --- |
| no dismiss in `selectAgent` | "Unable to find role=heading and name 'Sales Outbound'" |
| no dismiss in `selectDirectMember` | "Unable to find role=main and name 'Direct conversation with Alice'" |
| no dismiss in `openBotSetup` | "Unable to find role=main and name 'Create a new agent'" |
| dismiss back below the `agentSetupOpen()` guard | the same, on the second press |

### A refresh that keeps the report

A completed turn and a host coming back online both refresh the open report, and both blanked
`result` first. That unmounted `AgentUsageReport`, so the report came back with its breakdown reset
to Agent and scrolled to the total, losing the table the user had chosen. A turn completes while the
report is open often enough that this was the common case, not the rare one.

`AgentUsagePanel.load` now takes a `background` flag. Only a request the user made - the first load, a
filter change, Refresh, Retry - may blank the body; an event-driven refresh replaces the numbers in
place, and one that fails leaves the last good report where it is rather than an error the user
cannot act on. One extended test in `AgentUsagePanel.test.tsx` selects Day, emits `turn-completed`,
waits for the new session count to arrive, and asserts the table node is the same one.

| Mutation | Failure |
| --- | --- |
| `load(state.range)` on `turn-completed` again | "Unable to find role=table and name 'Daily usage and cost'", while the new totals still arrive |
| the failure guard back to `!background` | `Unable to find role="alert"` |

A follow-up review finding on the same guard: a refresh can start while the first load is still in
flight, and it takes the generation with it, so the foreground response is discarded. Dropping the
background error too then left the panel on Loading with Refresh disabled and no Retry. The guard now
keeps the last good report only while a report exists; with nothing on screen the failure surfaces.
The second test holds the first request pending, emits the turn, and rejects the refresh.

One review finding was argued down rather than fixed: "Keep Grok prompt usage separate from session
counters" read `response.usage` as per prompt. ACP documents it as session-cumulative - `inputTokens`
is "Total input tokens across all turns", `totalTokens` is "Sum of all token types across session" -
so one counter per session with delta subtraction is right, and the proposed per-turn counter would
re-count every earlier turn. `usage-collection.ts` now names that contract in a comment. The
reviewer withdrew the finding.

### Verification

Every added assertion was watched fail for the reason it names. The mutations: dropping `providerDaily` from the wire codec; swapping its date decoder for a plain string; making the field optional; keying a usage cell on the provider alone; creating a cell before the day guard; drawing the chart from `daily` again; deleting the tooltip's Total row; deleting only version 14; a strict migration 15; removing the rollback case's squatted name; and turning the migration runner's `ROLLBACK` into a `COMMIT`. Each failed as intended, and the code was restored.

`bun run lint` (8 pre-existing warnings), `bun run typecheck` and `bun run check:ui` pass. The long suites are CI's.

One thing left alone: `ServerRail.stories.tsx` `Populated` has a play assertion on `transitionDuration` that expects `0.16s` and reads `0s, 0s`. It predates this branch, and CI's Storybook job only builds, so it is not blocking - but it is animation timing asserted in a test, which the guidance rules out, and it should be deleted rather than fixed.

Model/harness: Claude Opus 5 in Claude Code; Vitest and Playwright against local Storybook.

### Screenshots

Before, as published above at `54e0d7c8`: one white area, no legend dots, two-line model rows, Totals wrapped under a caption.

After - three provider series, the row dots as legend, Totals on one line, denser rows, the arrow on the inset:

Usage with one area per provider

The hover card names every provider for that day and totals them:

Per-day tooltip

The server menu item:

Usage in the server context menu


NorbertBodziony:
The Grok finding is wrong: ACP `Usage` is session-cumulative by contract, not per prompt.

`@agentclientprotocol/sdk` documents the very fields this branch reads (`schema/types.gen.d.ts`, type `Usage`):

- `totalTokens` — "Sum of all token types across session."
- `inputTokens` — "Total input tokens across all turns."
- `outputTokens` — "Total output tokens across all turns."
- `cachedReadTokens` / `cachedWriteTokens` — "Total cache read tokens." / "Total cache write tokens."

So `#consumePrompt` forwarding `response.usage` after each prompt forwards a running session total, which is exactly what one counter per session plus delta subtraction expects. Two turns do not "each report 350": the second reports the cumulative figure, and the delta is the new part. Codex is handled the same way (`${session.id}:codex`) for the same reason.

The proposed fix would break it. With the turn ID in the counter ID and `baseline: false`, every turn would start a fresh counter and store its whole session-cumulative figure as new usage — turn 2 would re-count turn 1, turn 3 would re-count turns 1 and 2, and the totals would grow quadratically.

The baseline branch is also deliberate, not a discard bug. For a session created before the analytics schema was applied, the first report carries all the tokens that session spent before collection existed. Recording it as usage would attribute pre-installation spend to today. The sample still writes the checkpoint, so every later prompt in that session is counted in full from that point.

`usage-collection.ts` now carries a comment naming this contract, so the next reader does not have to re-derive it (`184e943d`).

/norbiai review

PR #323, opened 2026-09-08
fix(chat): stop an agent id problem from hiding chat history (#298)

### LINKED ISSUE #298: Chat history disappears after OpenBot update
## Summary

After updating OpenBot, a user can no longer see a chat that was previously visible. The chat history appears to have disappeared or is no longer loaded.

## Steps to reproduce

1. Open OpenBot and verify that an existing chat is visible.
2. Update OpenBot to a newer version.
3. Launch the updated version.
4. Open the chat list or the relevant integration.

## Actual behavior

The previously visible chat and/or its history is missing after the update.

## Expected behavior

Existing chats and their history should remain available after an application update. Updating OpenBot should preserve locally stored chat data and continue displaying it to the user.

## Impact

Users may lose access to previous conversations and the context stored in them.

## Additional context

The issue was reported for the `OpenBot_Posty` setup. Before the update, the chat was visible; after the update, it is no longer visible. The exact OpenBot version, operating system, chat identifier, and whether the data still exists on disk are not yet known.

## Acceptance criteria

- Existing chat history remains visible after updating OpenBot.
- Chat data is migrated when the storage format changes.
- The application does not silently hide or overwrite existing conversations during an update.
- If migration fails, the user receives a clear error and the original data is preserved.
- Automated coverage verifies chat-history persistence across update and migration scenarios.

NorbertBodziony:
Closes #298.

A user reports a chat that was visible before an OpenBot update and gone after it. The report carries no version, no operating system, no chat id and no statement about whether the data is still on disk, so **this is not a reproduction**. What the investigation found is why an id or roster problem becomes permanent, silent history loss instead of a visible error, plus one defect that hides a chat with no data loss at all. Five paths, each silent. The first is the root cause the rest only mitigate, and one of the five is deliberately left in place — see the migration section.

## The roster had no replay path

`orchestration_events` is the source of truth and every roster write appends the whole agent list to one aggregate, but nothing read it back: `projection_agents` was written only by `replaceAgents`, and `hasAggregateEvents` had one caller. An empty projection beside a non-empty log therefore passed the startup guard in silence — `.every(isStoredAgent)` on an empty list is true — so the user saw **no agents at all**, and the very next persist wrote the empty list as the new truth while each thread and message row stayed on disk. Demonstrated by running it before the fix:

```
WARN [agent-store] Threads remain that no agent claims. 1
agents after restart: []
event log still has roster history: true
messages still on disk: 1
```

`AgentRoster.latestRosterAgents` reads the agent list out of the newest event on the `agents` aggregate. Newest only, never a fold: `hardDeleteAgent` appends the *remaining* agents, so a fold would resurrect an agent the user deleted on purpose, and a newest event with an empty list correctly means the user has no agents. The payload shape — `aggregateType: "agents"`, `aggregateId: "agents"`, `payload: { agents }` — is unchanged since the first SQLite commit and identical at `v0.4.0`, so no version adapter is needed. `AgentStore` validates each entry with the guards it already has, drops the ones that fail with a warning rather than throwing (an event can be older than any shape this build knows, and nothing here may stop the app), persists the result so the repair lasts, and runs before `#reconcileUnclaimedThreads` so a restored agent can then claim the thread that names it. It works per agent rather than only on an empty roster: a partial loss is the more likely half — `replaceAgents` truncates the roster and re-inserts the in-memory list, so a persist made with one agent missing drops that one row and keeps the rest — and the more dangerous half, because a later `getOrCreate` rebuilds the missing agent with `threadId: null` and both read paths then report an empty history. The restored profile carries the `threadId` the event holds, which is what reaches a thread minted under a random id by an earlier build.

## Nothing could reach a thread whose roster row was gone

`projection_threads.agent_id` carries no foreign key to `projection_agents`, the chat list is the roster with no `WHERE` and no join, nothing in the app enumerated threads, and there is no soft-delete column. So a thread that fell out of the roster had no way to appear and no way to be reported — it read as an empty chat while every message stayed on disk. `replaceAgents` truncates the roster and re-inserts the in-memory list while `ensureThreadProjection` never deletes, so any persist made with an agent missing produced exactly that state.

`AgentRoster.unclaimedThreads` is the one query that can see such a row. `AgentStore.initialize` now gives the thread back to the agent that names it, matched under either the `bot-` or `agent-` spelling. It runs after the two existing recoveries, so a thread belonging to an agent they just removed is gone rather than re-adopted; an agent that already holds a thread keeps it, because handing it a second would hide the first; and a thread it cannot place is counted in a warning and left alone, because nothing here may stop the app.

## A new thread id was random

`ensureThreadIdNow` minted `openbot-thread-${randomUUID()}` while `stableThreadId(agentId)` already existed, was used by the legacy `bots.json` import, and was asserted by the v13 test. Both conversation read paths call `getOrCreate`, so reading a chat whose roster row is gone rebuilt the agent with no thread and landed here — and the random id filed the rebuilt agent against an empty thread, leaving the user's own thread addressable by nothing. It now uses the stable id, which is the row the history is already under.

## Migration v13 can merge two conversations into one, and is left frozen

`readAgentIdRenames` builds its `taken` set from `projection_agents` alone, so an orphan thread already keyed at a rename target is invisible to it. `OR REPLACE` then deletes the orphan's thread row, both threads' messages are rewritten onto the one surviving thread id in the same pass, `PRAGMA foreign_key_check` stays empty, and the two conversations merge in silence. Two messages sharing an id lose one outright.

**This corrects the failure mode in the plan**, which said the collision caused a rollback loop. The test showed it does not — there is no rollback, only the merge.

An earlier revision of this PR widened `taken` to cover `projection_threads`. **That is reverted**: v13 is byte-identical to the released implementation again, and only a comment differs. A migration that transforms one database differently from another under the same version number is the larger hazard, `AGENTS.md` makes "never edit a migration that may have shipped" non-negotiable, and the guard helped only a database still below v13.

The hole is closed from the other side instead, at runtime and at every startup: the roster replay restores an agent the projection lost, and `unclaimedThreads` gives an orphan thread back to the agent that names it — so the orphan the collision needs stops existing. Residual risk, stated plainly: a user updating straight from a pre-v0.5.0 release with an orphan thread already present runs the migrations before any startup reconciliation, so that one upgrade can still merge two chats. The comment in `readAgentIdRenames` now records this.

## A remote layout could hide a whole section

The sidebar draws only the sections `order` names, and `filteredAgentsBySection` fell back to Unassigned only when the assigned section was missing from `sections`, not from `order`. The local store cannot produce such a layout, but the wire decoder did not carry the invariant — `isSidebarLayoutSnapshot` checked only `order.every(isString)` — so a remote host could serve a layout that put every agent in a section off the screen with the agent, its chat and every message intact. This is the closest match to "the chat was visible, now it is not".

`isCompleteSectionOrder` states the invariant once, in the contract, and `SidebarLayoutStore` now uses it in place of its own copy of the same check. The sidebar also falls back to Unassigned for a section `order` leaves out, so an agent stays on screen if a layout ever reaches the component another way.

### UI change

No visual change on any valid layout: the fallback only fires for a layout the contract now rejects at the boundary. Before, an agent assigned to a section absent from `order` rendered nowhere — the new test fails with `Unable to find an accessible element with the role "button" and name /Chief/`. After, it appears under **Unassigned**, where the user can open it and file it again.

## Tests, each watched fail

`AGENTS.md` rule 2. Every new assertion was watched go red for the reason meant, by breaking exactly what it covers:

| Assertion | Broke | Failed as |
|---|---|---|
| every agent the roster projection lost comes back, not only a wholly empty roster | narrowed the replay to `persisted.length === 0` | `expected [ 'sales-outbound' ] to deeply equal [ 'sales-outbound', 'chief' ]` |
| the roster is rebuilt from the event log after its projection is emptied | disabled the replay call | `expected undefined to be 'openbot-thread-chief'` — the agent is absent from `list()` entirely |
| `ensureThreadId` returns `openbot-thread-chief` (extended the existing test, which only checked a round trip) | restored `randomUUID()` | `expected 'openbot-thread-a5ccd610-…' to be 'openbot-thread-chief'` |
| a thread its roster row stopped naming is given back, and the page returns its message | removed the `initialize` call | `expected null to be 'openbot-thread-85e9dfac-…'` |
| the same, with only the stable-id fix in place | removed the `initialize` call alone | still red, so both fixes are independently load-bearing |
| the contract rejects a layout whose `order` omits a section, and one that repeats a section | restored `order.every(isString)` | `expected true to be false`, both cases |
| an agent under an unordered section renders in Unassigned | dropped the `order` half of the fallback | `Unable to find an accessible element with the role "button" and name /Chief/` |

Three of the six extend an existing test rather than adding one, and no new test file was added. The existing "rejects malformed sidebar layout events" case turned out to be partly a costume — its `order: ["people", "people", "unassigned"]` was only rejected because `revision` was `-1` — so the duplicate case now has an assertion that actually covers it.

## Verification

`bun run test:desktop` on `openbot-database.test.ts` (31), `agent-store.test.ts` (32), `Sidebar.test.tsx` (20), `sidebar-layout-store.test.ts` (10), `openbot-database-schema-parity.test.ts` (3), the full `packages/contracts` suite (136), and every other file that touches a sidebar layout — `App.test.tsx`, `App.servers.test.tsx`, `mock-sidebar-layout.test.ts`, `agent/profile-save.test.ts`, four `team-api-server` files, `host-service.test.ts`, `ipc/input-parsers.test.ts`, and all five `agent-service.*` files. All green. Plus `bun run lint` (only the eight pre-existing mobile module-mock warnings), `bun run typecheck` (11 projects) and `bun run check:ui`. CI owns the wider suites.

The end-to-end check in the plan is not possible here: the only live dev instance belongs to another worktree, so a snapshot would exercise that worktree's code rather than this change.

## A user with a missing chat now has somewhere to look

`docs/TROUBLESHOOTING.md` is where the startup error dialog sends people, and it had no entry for this symptom — while **Reset OpenBot**, which moves the only copy of their messages out of reach, sat one heading below. The new section says not to do that, says a restart now recovers most cases on its own, and gives the query that reports a chat no agent claims with the number of messages waiting in it. The query is verified against a database built by the real schema, not just eyeballed.

## Surfaces touched

Desktop renderer (`src/renderer`), backend and the user's SQLite (`src/backend`), the IPC contract (`packages/contracts`), and `docs/TROUBLESHOOTING.md`. `mock-openbot.ts` needed no edit — its fixture is `order: [people, unassigned]` with no sections, and every path in `mock-sidebar-layout.ts` keeps `order` complete, so the second implementation already satisfies the tightened guard. No mobile, no `apps/auth-api`, no `remote/api`, no `PRIVACY.md` change; nothing new leaves the machine. No new migration, so the two database build paths stay in parity.

## What this does not do

No pre-upgrade copy of `openbot.db` — `AGENTS.md` forbids it, so the migration is made safe instead. The issue's criterion "if migration fails, the user receives a clear error and the original data is preserved" was **already met** before this change: `src/main/index.ts` shows `dialog.showErrorBox` and says the data was not reset. Migration v13's core algorithm, the transactional runner and that failure surface are all untouched.

To close this against the reporter's own machine, they need to supply the app version, the operating system, and the output of this query against `~/Library/Application Support/OpenBot/openbot.db` (or the Windows equivalent):

```sql
SELECT t.thread_id, t.agent_id, (SELECT count(*) FROM projection_thread_messages m
  WHERE m.thread_id = t.thread_id) AS messages
FROM projection_threads t
LEFT JOIN projection_agents a ON a.agent_id = t.agent_id
WHERE a.agent_id IS NULL;
```

A non-empty result confirms the first defect and means the history is recoverable by the reconciliation in this PR.

---

Model: Claude Opus 5. Harness: Claude Code (T3 Code).


NorbertBodziony:
[P2] Fixed in `8f01d63` (latest commit): `#restoreRosterFromEvents` no longer waits for an empty roster. It compares the newest roster event against the agents it loaded and puts back each one that is absent, carrying the `threadId` the event holds — which is what reaches a thread an earlier build minted under a random id. Still the newest event only, so an agent the user deleted stays deleted. The test now deletes one roster row of two, and was watched fail against the old condition with `expected [ 'sales-outbound' ] to deeply equal [ 'sales-outbound', 'chief' ]`.

[P1] The explicit developer decision the finding asks for exists: the approved plan for this issue directs this exact change — "add the `agent_id` values of `projection_threads` to the `taken` set in `readAgentIdRenames`" and "do **not** edit or renumber v13 itself in any other way, and do not add a migration". It is quoted in the PR body under the ⚠️ flag so a human looks at it deliberately.

Two points of substance against the finding:

- "A database already marked v13 skips the guard" is the reason a later migration cannot replace this one. v13 merges the two conversations *during its own run*: `UPDATE OR REPLACE` deletes the orphan's `projection_threads` row and rewrites both threads' messages onto the surviving thread id in the same pass, so `PRAGMA foreign_key_check` stays empty and nothing dangles. Any migration numbered above v13 runs after that merge and after the row that separated the two conversations is gone, so it has nothing left to distinguish them by. The guard is preventive only, and to be preventive it has to sit inside v13. The population it protects is every database still below v13 — users on a release before v0.5.0 who have not updated yet.
- "Older databases receive a different transformation under the same version" is true only for a database that holds a thread whose `agent_id` matches no roster row *and* whose id is the target of a rename. With no such row the `taken` set is identical and the transformation is byte-for-byte the same, which is why every existing v13 assertion and `openbot-database-schema-parity.test.ts` are untouched. The alternative is not "no change under v13" — it is that those databases silently merge two chats into one and lose a message wherever two share an id.

/norbiai review

NorbertBodziony:
The repository owner has approved this exception for this change specifically, which is the explicit developer decision the finding asks for. Recorded in the PR body under the ⚠️ flag. The finding is accurate about the rule and correct to hold at P1 rather than withdraw — leaving it unresolved so a human signs it off is the intended outcome, not something to review away.

NorbertBodziony:
Reverted the v13 change instead of arguing it further: `readAgentIdRenames` is byte-identical to the released implementation again, and only its comment differs — it now records what the migration cannot see and where the hole is closed instead. The orphan-collision test and its `seedUnclaimedThread` helper are gone with it, because pinning the merge as correct would block a real fix later; the runtime half stays covered in `agent-store.test.ts`.

Residual risk kept on the record in the PR body: a user updating straight from a pre-v0.5.0 release with an orphan thread already present runs the migrations before any startup reconciliation, so that one upgrade can still merge two chats. Every later launch cannot, because the roster replay and `unclaimedThreads` stop the orphan from existing.

/norbiai review

PR #340, opened 2026-09-08
Show every integrated model and let a user CLI update itself

NorbertBodziony:
## What changed

Three things a user sees, all about the provider CLIs OpenBot does not own.

**Every model the CLI lists is shown, and named by the model.** Claude Code names a model by the
part it plays in that CLI ("Default (recommended)", "Opus"), which says which pick it is there, not
which model an agent runs here. `claudeModelName` derives the product name from the id instead, and
falls back to the CLI's own label for an id that does not read as one.

| Claude tab, before | Claude tab, after |
| --- | --- |
| Default (recommended) | Claude Opus 5 |
| Fable | Claude Fable 5.1 (1M context) |
| Opus (default) | Claude Sonnet 5 *(default)* |
| Haiku | Claude Haiku 4.5 |

**Each provider names a default model.** `DEFAULT_PROVIDER_MODELS` in the contracts package replaces
the same ternary written out in three places plus a renderer map: `gpt-5.6-luna`, `claude-sonnet-5`,
`grok-4.6`. Claude moves from Opus to Sonnet, and Grok gains a default it never had - it used to take
whatever model the CLI listed first. Every caller still falls back to the provider's first model, so
an id a given CLI does not list is not an error.

**A CLI the user installed can update itself.** The provider row already compared a `system` install
against the pinned lock; the Update button on such a row now runs that CLI's own updater
(`codex update`) rather than doing nothing. OpenBot downloads nothing on this path, so the pinned
artifact checksums are untouched, and the managed copy still refuses the command - the runtime
manager replaces that install whole.

That updater decides for itself what "latest" means, and its channel can name an older version than
the lock: `grok update` reports success and leaves the CLI on 1.0.13 while the lock pins 1.0.22.
Without a record, the offer returned at every start and every click walked the user through an update
that cannot happen. `ProviderRuntimeManager.noteSystemCliUpdate` now records an update that finishes
on the version it started on, in `cli-update-refusals.json` beside the managed runtimes, and drops
the offer. The record holds the installed and the pinned version together, so a new pinned version is
a new offer, and so is a CLI the user moves by other means.

Surfaces: desktop renderer, IPC contracts (`app:update-provider-cli`) and `mock-openbot.ts`, the
preload boundary, main-process provider handlers, `docs/ARCHITECTURE.md`. No migration, no mobile,
no `apps/auth-api`, no palette change. Reverse states covered: the offer appears again once the
updater installs something, and once a different version is pinned.

## Verification

- [x] Narrow checks for what changed: `biome check --write <paths>`, `bun run lint`,
      `bun run typecheck`, `bun run check:ui`, and the affected test files -
      `provider-runtime-manager.test.ts` (14), `provider-update-toast.test.tsx` (10),
      `provider-runtime.test.ts` (18), `ProviderModelPicker.test.tsx` (3),
      `profile-generation.test.ts` (7), `claude-client.test.ts` (15). CI owns the full suite.
- [x] Manual smoke test in the dev app: the Grok row offered v1.0.13 -> v1.0.22, Update ran the CLI's
      own updater, the notification settled on "Grok is up to date v1.0.13",
      `cli-update-refusals.json` recorded `{"grok":{"version":"1.0.13","pinnedVersion":"1.0.22"}}`,
      and after a main-process restart no offer returned. The picker showed
      "Claude Sonnet 5, default" and "Grok 4.6, default".

Each added assertion was broken and confirmed to fail for the intended reason, then restored:
removing the refusal check in `getStatus()`, removing `#readRefusedUpdates()` from `initialize()`,
and removing the record's delete branch each failed exactly one of the new manager tests; naming
Opus as the Claude default failed the picker test.

## Risk and security impact

One new IPC request, `app:update-provider-cli`, taking a provider id parsed by the existing
`parseProviderId` and returning a decoded `AgentStatus`. It spawns the provider CLI that is already
resolved and already spawned for every session, with `argv: ["update"]`, no shell, an empty added
environment, `stdio: ["ignore", "ignore", "pipe"]` and a ten-minute timeout; the managed copy is
refused. New persisted state is one small JSON file of version strings in the provider-runtimes
directory, written atomically through a temp file and a rename, and read behind runtime guards - a
missing or malformed file only costs the user one more refused offer. No new network access, no new
filesystem reach, no change to sandboxing, context isolation, navigation policy or sender validation.

## Screenshots

Text over pixels here: the change is the accessible names in the model picker, which the tables above
give before and after. The `dev:automation` accessibility snapshot from the running app reads
`option "Claude Sonnet 5, default"` and `option "Grok 4.6, default"`.

## Review follow-ups

Added after the first review round, each with its own commit.

**The update no longer cuts a turn short.** The updater replaces the binary under a running client,
and the client is restarted after it, so a turn in flight would lose its process. A provider that has
an agent in a turn - or a delivery on its way to one, which holds no turn id yet - refuses the
command and asks the user to wait. `ProviderHooks` gains `isProviderBusy` and `onProviderResumed`;
`DrainScheduler` counts deliveries between their first await and their turn, and skips an agent whose
provider reports `isReplacingCli`. Deliveries wait in the mailbox instead of the event loop, and
`onProviderResumed` schedules the held ones when the replacement ends, after a failure as well as
after a success.

**A stale refusal is cleared.** A CLI that moves by other means is a different install, so the
renderer store keys the refusal on the installed version as well as the offered one and offers the
pinned version again.

**The composer stays on screen.** `.usage-workspace-content` (from #321) uses `display: contents`,
so `.app-frame > * { min-height: 0 }` no longer reached the conversation panel: the panel took its
automatic minimum size, 1970px in an 820px frame, and pushed the composer under the window edge.
`usage.css` restores `min-height: 0` for the wrapper's children. Measured live in the dev app, the
composer moved from top 1926 (off screen) to top 776 / bottom 804 in an 820px viewport, and
"Message Chief" is on screen again. This is a main-branch regression the rebase brought in, not a
change this PR made.

Tests for these: "refuses to replace a CLI that is running a turn", "refuses to replace a CLI while a
delivery is on its way to a turn", and "delivers a message queued while a failed update held the CLI"
in `provider-runtime.test.ts`, and the reoffer assertion in `provider-update-toast.test.tsx`. Each was
break-checked and restored: blocking `turn/start` and dropping `isProviderBusy` failed the first two;
removing `onProviderResumed` timed out the third. Removing the `isReplacingCli` guard in
`DrainScheduler` alone does not fail the third test, because the global `isReady()` guard also holds
while the provider is "connecting" - the guard is kept as the explicit per-provider rule.

### Second review round (P2)

**A compaction holds the update too.** A compaction is a provider turn the agent never owns -
`ContextCompaction.claimTurn` takes its `turn/started` away, so it sets no active turn id.
`isProviderBusy` now asks `ContextCompaction.mayDrain` as well.

**The offer is for this computer only.** Every runtime the store reaches is local, while the agent
status beside it describes whichever server is open, so a remote workspace could be announced an
update to a runtime the user is not looking at. The store takes `isLocalServer`, bound in
`providers.tsx` to `activeServer()?.kind === "local"` - the guard the provider row and the picker
already use. A remote workspace announces nothing, and `startProviderUpdate` rejects.

Tests: "refuses to replace a CLI that is compacting a thread" (`provider-runtime.test.ts`) and
"offers no CLI update while a workspace on another computer is open"
(`provider-update-toast.test.tsx`). Break-checked and restored: removing the compaction clause fails
the first; removing the effect guard, and separately the action guard, each fails the second.

### Third review round (P1, P2)

**A replacement is not a restart.** `#reloadProviderCli` ended in `onProvidersReady`, which is
restart recovery: it settles every unresolved delivery, and the other providers keep running through
the replacement, so a live turn was recorded as `interrupted` - a state `MailboxStore.markTerminal`
then refuses to correct. `#activateProviderClient` takes `notifyReady`, the replacement passes
`false`, and `onProviderResumed` still schedules the deliveries it held back.

**A closed notification stays closed across a server switch.** The switch rebuilds the runtime store,
so the version each notification was closed on is kept by the notification module, which outlives it.
Teardown hides its notifications instead of ending the offer; an update that was acted on, cancelled
or settled still clears the record.

Tests: "leaves another provider's running turn alone while a CLI is replaced"
(`provider-runtime.test.ts`) and "keeps an offer the user closed closed when the workspace is opened
again" (`provider-update-toast.test.tsx`). Break-checked and restored - restoring `notifyReady: true`
fails the first with `expected 'interrupted' to be 'running'`; removing the announce-loop skip, and
separately making teardown dismiss instead of hide, each fails the second.

### Fourth review round (P1, P2)

**Both replacement paths skip restart recovery.** `notifyReady: false` covered only the provider that
still has a client. A provider whose client is gone is connected again through `#connect`, which also
ended in `onProvidersReady`, so updating it while another provider ran a turn still recorded that live
delivery as `interrupted`. `#connect` takes the same option. The test is now a loop over both
branches; reverting only the `#connect` guard fails the "with no client is connected again" variant
with `expected 'interrupted' to be 'running'`.

**One queue for the refusal record.** Two providers can finish an update at once. Each write
serialized the record as it stood when it ran and renamed its own temporary file, so an older snapshot
could land last and drop a newer refusal - which the user would meet as an offer they had already
turned down. `#writeRefusedUpdates` now writes one at a time. No test: making the broken code fail
needs two renames to complete in a chosen order, which is a timing-based test.

### Fifth review round (P1, P2)

**The updater's own reason is redacted.** That line goes into the provider row, and the Team API
broadcasts the row to the team's connected clients. A CLI that fails on authentication prints the
header or key it sent, so `readProcessReason` redacts before it bounds the line.

**The CLI owner survives a signed-out provider.** `cliSource` came only from the CLI a client runs
on, and a signed-out provider has no client although its binary is resolved and its version is on the
row - so the row named no owner, and an unowned CLI reads as the managed copy, which would send the
user's own install to a download. Resolution records the owner and drops it when nothing resolves.

Tests: "redacts a secret in the reason the updater printed" and "keeps the owner of a CLI whose
provider is signed out". Break-checked and restored.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -459,10 +459,12 @@ failure keeps the reason the CLI gave, in one error that goes to the provider ro
 OpenBot downloads nothing on this path, so the pinned artifact checksums are untouched. The managed
 copy refuses this command, because the runtime manager replaces that installation whole.
 
-The update replaces the binary under a running client. A provider that has an agent in a turn
-therefore refuses the command and tells the user to wait, and no turn may start on that provider
-until the new client is ready: the drain scheduler skips an agent whose provider reports
-`isReplacingCli`, and the delivery waits in the mailbox until the new client schedules it again.
+The update replaces the binary under a running client. A provider that has an agent in a turn -
+or a delivery on its way to one, which holds no turn id yet - therefore refuses the command and
+tells the user to wait. No turn may start on that provider until the new client is ready: the drain
+scheduler skips an agent whose provider reports `isReplacingCli`, before it can reschedule the
+delivery, and `onProviderResumed` schedules the held deliveries when the replacement ends, after a
+failure as well as after a success.
 
 That updater decides for itself what the newest version is, and its release channel can name an
 older one than the lock: `grok update` can report success and leave the CLI where it was. The IPC
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -462,7 +462,15 @@ copy refuses this command, because the runtime manager replaces that installatio
 Every runtime the store reaches is on this computer: `window.openbot.providerRuntimes` addresses no
 other one, while the agent status beside it describes whichever server is open. The store therefore
 takes `isLocalServer`, and a workspace on another computer announces no offer and starts no update -
-the same rule the provider row and the picker already follow.
+the same rule the provider row and the picker already follow. A server switch rebuilds that store,
+so the version a user closed the notification on is kept by the notification module, which outlives
+the switch: the offer is raised again on the way back only if the user never closed it.
+
+Replacing the CLI is not a start. `#activateProviderClient` skips `onProvidersReady` for it, because
+that hook is restart recovery: it settles every unresolved delivery, and the other providers keep
+running through the replacement, so a live turn would be recorded as `interrupted` - which
+`MailboxStore.markTerminal` then refuses to correct. `onProviderResumed` schedules the deliveries
+the replacement held back.
 
 The update replaces the binary under a running client. A provider that has an agent in a turn -
 a delivery on its way to one, which holds no turn id yet, or a context compaction, whose
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -466,11 +466,14 @@ the same rule the provider row and the picker already follow. A server switch re
 so the version a user closed the notification on is kept by the notification module, which outlives
 the switch: the offer is raised again on the way back only if the user never closed it.
 
-Replacing the CLI is not a start. `#activateProviderClient` skips `onProvidersReady` for it, because
+Replacing the CLI is not a start, on either path: `#activateProviderClient` swaps the client of a
+provider that has one, `#connect` connects one whose client is gone, and both skip
+`onProvidersReady` for the replacement, because
 that hook is restart recovery: it settles every unresolved delivery, and the other providers keep
 running through the replacement, so a live turn would be recorded as `interrupted` - which
 `MailboxStore.markTerminal` then refuses to correct. `onProviderResumed` schedules the deliveries
-the replacement held back.
+the replacement held back. The refusal record is written through one queue, because two providers
+can finish an update at once and the older snapshot must not be renamed over the newer one.
 
 The update replaces the binary under a running client. A provider that has an agent in a turn -
 a delivery on its way to one, which holds no turn id yet, or a context compaction, whose
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -455,8 +455,14 @@ one Update button, and one entry point in the runtime store, `startProviderUpdat
 behind it differs: a `system` install goes to `updateProviderCli`, which runs that CLI's own updater
 (`codex update`) and then restarts the provider on the binary now on disk. That updater reports no
 progress, so the notification holds its indeterminate step until the provider comes back, and a
-failure keeps the reason the CLI gave. OpenBot downloads nothing on this path, so the pinned artifact checksums are untouched. The
-managed copy refuses this command, because the runtime manager replaces that installation whole.
+failure keeps the reason the CLI gave, in one error that goes to the provider row and to the caller.
+OpenBot downloads nothing on this path, so the pinned artifact checksums are untouched. The managed
+copy refuses this command, because the runtime manager replaces that installation whole.
+
+The update replaces the binary under a running client. A provider that has an agent in a turn
+therefore refuses the command and tells the user to wait, and no turn may start on that provider
+until the new client is ready: the drain scheduler skips an agent whose provider reports
+`isReplacingCli`, and the delivery waits in the mailbox until the new client schedules it again.
 
 That updater decides for itself what the newest version is, and its release channel can name an
 older one than the lock: `grok update` can report success and leave the CLI where it was. The IPC
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -438,10 +438,36 @@ installation is display metadata until the pinned runtime passes the existing do
 checks. Runtime snapshots carry the previous version and an optional `availableVersion` through the
 preload decoder. Cancellation and failure preserve the previous installation and its update offer.
 
-Settings starts the shared renderer runtime store. An explicit update opens one notification;
-revisioned snapshots move it through progress, failure, retry, and completion. Closing the
-notification does not cancel the download, and later reports do not reopen it. Fresh provider
-downloads retain their existing flow. These actions apply only to the local desktop host.
+Settings starts the shared renderer runtime store. The store announces each provider that gains an
+offer as one notification, from an effect over both the runtime snapshot and the agent status,
+because the two arrive separately and either one can complete an offer. An explicit update opens
+the same notification; revisioned snapshots move it through progress, failure, retry, and
+completion. Only the crossing into "update available" is announced, so a dismissed notification
+stays dismissed until the offer changes. Closing the notification does not cancel the download,
+and later reports do not reopen it. Fresh provider downloads retain their existing flow. These
+actions apply only to the local desktop host.
+
+A CLI the user installed themselves is not managed, but it is still compared against the lock.
+Each provider status row reports `cliSource`, and main passes the version of a `system` row to
+`ProviderRuntimeManager.setSystemVersion`, which compares it against the pinned version exactly as
+it compares a managed installation. The row and the notification therefore use the one update offer, the
+one Update button, and one entry point in the runtime store, `startProviderUpdate`. Only the work
+behind it differs: a `system` install goes to `updateProviderCli`, which runs that CLI's own updater
+(`codex update`) and then restarts the provider on the binary now on disk. That updater reports no
+progress, so the notification holds its indeterminate step until the provider comes back, and a
+failure keeps the reason the CLI gave. OpenBot downloads nothing on this path, so the pinned artifact checksums are untouched. The
+managed copy refuses this command, because the runtime manager replaces that installation whole.
+
+That updater decides for itself what the newest version is, and its release channel can name an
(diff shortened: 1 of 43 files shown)


NorbertBodziony:
GitHub did not create a check run for 24595cd3 (the push event was dropped), so CI is re-triggered by hand. Both P1 findings are addressed in that commit: the drain guard now runs before the try block and `onProviderResumed` schedules the held deliveries after success and after failure, and `startDelivery` claims its provider before its first await so an update cannot replace the client it is about to prompt.

/norbiai review

NorbertBodziony:
Both P2 findings are addressed in `64422cf5`.

**Compaction (finding 1) - valid, fixed.** `isProviderBusy` now also asks `ContextCompaction.mayDrain`, so a provider that is compacting a thread refuses the update. New test: "refuses to replace a CLI that is compacting a thread" in `provider-runtime.test.ts`. Break-check: with the compaction clause removed from `isProviderBusy`, that test fails; restored.

**Local-server guard (finding 2) - valid, fixed.** `createProviderRuntimeStore` takes `isLocalServer`, which `providers.tsx` binds to `activeServer()?.kind === "local"`. A remote workspace announces no offer, and `startProviderUpdate` rejects. New test: "offers no CLI update while a workspace on another computer is open" in `provider-update-toast.test.tsx`. Break-check: removing the effect guard, and separately removing the action guard, each fails that test; restored.

**The second half of finding 2 - announcement state across server switches - is not a defect.** The store's `onSettled` cleanup already calls `dismissProviderUpdateToast` for every provider, so a server switch takes the notification off the screen. Nothing the user dismissed is re-raised: the switch removed it, and returning to this computer restores an offer that is still true. Retaining the record across the switch would do the opposite - a live offer with no notification and nothing to raise it again. With the guard above, the record is also left alone while a remote workspace is open, so it keeps describing this computer.

/norbiai review

NorbertBodziony:
Both findings are fixed in `8ff4788b`.

**P1, restart recovery during a replacement - valid, fixed.** `#reloadProviderCli` activated the new client through the normal path, which ends in `onProvidersReady`, and that hook is restart recovery: `reconcileUnresolvedDeliveries` settles every unresolved delivery, including the live ones of the providers that keep running. `#activateProviderClient` now takes `notifyReady`, and the replacement passes `false`; `onProviderResumed` still schedules the deliveries it held back.

New test: "leaves another provider's running turn alone while a CLI is replaced" in `provider-runtime.test.ts` - Codex holds a running turn, Claude's CLI is replaced, and the Codex delivery stays `running` on its turn id. Break-check: with `notifyReady: true` restored, the test fails with `expected 'interrupted' to be 'running'`, which is the failure described in the finding. Restored.

**P2, a closed offer returning after a switch - the argument was wrong, fixed.** The version each notification was closed on is now kept in `provider-update-toast.tsx`, which outlives the server scope, and the announce loop skips an offer that carries that version. The store hides its notifications on teardown (`hideProviderUpdateToast`) rather than ending the offer, so a switch takes the notification off the screen without forgetting what the user did with it; `dismissProviderUpdateToast` still clears the record, because an update that was acted on, cancelled or settled ends the offer.

New test: "keeps an offer the user closed closed when the workspace is opened again". Break-check: removing the skip in the announce loop, and separately making teardown dismiss instead of hide, each fails that test. Restored.

Local runs: `bun run lint` (11 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, `bun run check:ui`, `src/backend/agent` (20 files, 205 tests) and `src/renderer/src/features/provider-updates` all pass.

/norbiai review

NorbertBodziony:
## Fourth review round — both findings valid, both fixed (`ab4542ec`)

**[REMAINS][P1] Skip restart recovery when replacing a live provider client.** Correct: `notifyReady: false` covered only the branch that swaps an existing client. A provider whose client is gone goes through `#connect`, which still ended in `onProvidersReady()`. `#connect` now takes the same `notifyReady` option, and `#reloadProviderCli` passes `false` on that path too.

Break-check: the recovery test is now a loop over both branches (`for (const claudeSignedIn of [true, false])`). Reverting only the `#connect` guard fails exactly the "with no client is connected again" variant with `AssertionError: expected 'interrupted' to be 'running'`; the code was restored and 24 tests pass.

**[NEW][P2] Serialize writes to the update-refusal record.** Correct: each write serialized the map as it stood when it ran and renamed its own temporary file, so an older snapshot could land last and drop a newer provider's refusal — the user would then be offered an update they had already turned down. `#writeRefusedUpdates` now chains through a `#refusedUpdatesWrite` promise, one write at a time.

No test was added for this one. Making the broken code fail needs two `rename` calls to complete in a chosen order, which is a timing-based test; the repository's test rules require waiting for state, an event, or a promise, not for elapsed time.

Checks run: `provider-runtime-manager.test.ts` + `provider-runtime.test.ts` (38 passed), `bun run lint` (11 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, `bun run check:ui`.

/norbiai review

NorbertBodziony:
## Fifth review round — both findings valid, both fixed (`68be457f`)

**[NEW][P1] Redact updater errors before broadcasting them.** Correct: `readProcessReason` returned the updater's raw stderr line, `updateProviderCli` puts it into the provider row, and the Team API broadcasts that row. `readProcessReason` now applies `redactText` before it bounds the line, so the one error the row and the caller share is already redacted.

**[REMAINS][P2] Preserve CLI ownership when the provider is signed out.** Correct, and my earlier rounds never answered it. The signed-out branch of `#connect` reports the resolved version and returns before `#cli` holds anything, so the row named no owner and an unowned CLI reads as the managed copy. Resolution now goes through `#resolveProviderCli`, which records the owner and drops it when nothing resolves; `status()` falls back to that record.

Tests, both in `provider-runtime.test.ts`: "redacts a secret in the reason the updater printed" and "keeps the owner of a CLI whose provider is signed out". Break-checked and restored — removing the redaction fails the first with `expected [Function] to throw error matching /\[redacted\]/`, and removing the ownership fallback fails the second on the missing `cliSource: "system"`.

Checks run: `provider-runtime.test.ts` + `provider-runtime-manager.test.ts` (40 passed), `bun run lint` (11 warnings, all pre-existing in `apps/mobile`), `bun run typecheck`, `bun run check:ui`.

/norbiai review

PR #366, opened 2026-09-09
Stop offering a pinned CLI version to the user's own install

NorbertBodziony:
## Problem

A Grok row offered `v1.0.5 → v1.0.22` on every app start, and its Update button appeared to do nothing.

`~/.grok/bin/grok` is the user's own install; the resolver prefers it over the managed copy. `ProviderRuntimeManager` compared it against the version pinned in `native-runtime.lock.json` and offered that version. The CLI's own updater decides what it installs — `grok update --check --json` reports 1.0.24 — so 1.0.22 named a version that install would never reach. The offer returned on every start, because the record of a closed notification lives only in renderer module memory.

Two silent paths made the button look dead:

- `startProviderUpdate` was called through `void`, so a rejection before either updater owned the notification was discarded.
- `updateProviderCli` queued behind `#runProviderConnectionCommand`, so an update started while a sign-in was open waited for the user, with nothing on screen but the step it was on.

## Change

- **Main** — `getStatus` offers the pinned version only to the copy the manager owns. `setSystemVersion`, `noteSystemCliUpdate`, and the refusal record are removed; `initialize` deletes the `cli-update-refusals.json` a previous version wrote, so an upgraded profile keeps no file that no reader opens.
- **Renderer** — `availableVersion` is `null` for every user install. A failure that reaches neither updater is put on the notification with a Retry. The notification names the version the provider comes back on, whatever it is, because OpenBot offered none to compare it against.
- **Picker** — a `system` row keeps the update without an offer: an always-available `Update CLI` action that claims no version.
- **Backend** — `updateProviderCli` refuses at once while another command runs on that provider, and names the reason.
- `MINIMUM_*_VERSION` in `src/backend/cli.ts` stays the one version claim OpenBot makes about a foreign install. `native-runtime.lock.json` stays as well: it is the download manifest (URLs, checksums, cloudflared, the remote-desktop recipe), and only its use as an offer for a foreign install is gone.

## UI

Settings → General, ChatGPT row, CLI installed by the user:

- **Before:** `ChatGPT · v0.146.0` with an update badge and an emphasized **Update** button labelled `Update ChatGPT to 0.153.4`, plus a `ChatGPT update available / v0.146.0 → v0.153.4` notification on every start.
- **After:** `ChatGPT · v0.146.0`, no badge and no notification, and an outline **Update CLI** button (`Update the ChatGPT CLI`). Pressing it opens the notification on its running step and settles on the version the CLI reports.

The `Setup/ProviderPicker` story `UserOwnedCliUpdate` shows the row in its new state. A managed row is unchanged.

## Checks

- `src/main` (65 files, 499 tests), `src/backend/agent/provider-runtime.test.ts` (27 tests), and the renderer settings and provider-update suites (59 tests) pass.
- `bun run lint` (13 pre-existing mobile warnings, none in the changed files), `bun run typecheck`, `bun run check:ui` pass.

### Each new assertion, broken and watched to fail

| Break | Failure |
| --- | --- |
| Skip `#removeRefusedUpdatesRecord()` | "promise resolved instead of rejecting" — the file is still there |
| Restore the pinned offer for a `system` CLI | "expected '0.153.4' to be null" |
| Drop `failProviderUpdate` | "Unable to find … ChatGPT update failed" |
| Drop the backend guard | the second update hangs; the test times out at 20 s, which is the wait the guard removes |
| Hide the `Update CLI` button | "Unable to find role=button and name 'Update the ChatGPT CLI'" |

Model: Claude Opus 5, harness: Claude Code (T3 Code).


NorbertBodziony:
Superseded by #367, which fixes the same report with the opposite design: OpenBot installs and
activates its pinned CLI and resolution then prefers that copy, so the offer this PR withdrew is
now actionable instead. This PR removed the offer for a CLI the user installed and routed it to
that CLI's own updater, which #367 deleted, so its 9 conflicting files are design conflicts rather
than textual ones.

The one part that still applies against main is carried over in #370: a refusal that reaches
neither the download nor its report - an update started while a workspace on another computer is
open - was discarded, leaving the offer on screen and nothing said.

The second silent path this PR fixed, an update queued behind a connection command that waits for
the user, is closed by #367 in a better place: its sign-in and busy-turn guards answer before the
wait, and no connection command holds the queue for user time.

PR #393, opened 2026-09-10
Add mobile queued message editing and management

nrb-dev:
Queued messages can be managed on mobile and edited on desktop and mobile without starting while the edit is open. The host stores the edit before removing the delivery from dispatch, and both clients recover its text, attachments, and reply target after restart.

- Add mobile queue presentation with cancel, steer, and edit actions.
- Save draft changes on the host. Send updates and requeues the original delivery, preserving user, agent, and routine senders. Cancel removes the saved edit without sending it.
- Serialize client writes and check revisions on the host. Retry a lost response without creating another message; reject a different request against a finished edit.
- Preserve other recipients of a broadcast message, attachment references, and the separate composer draft that existed before editing.
- Add the optional `queue-edit` capability, current v3 HTTP/WebRTC adapters, validated IPC, and the preview implementation. Older hosts require an update before editing can guarantee a pause. Released protocol schemas remain unchanged; the earlier `queue-take` route remains available.

Validation: 127 tests passed across eight individual files: mailbox store, desktop queue, mobile queue, Team API agent routes, preload queue, v3 protocol, IPC channel coverage, and queue-edit session. Full `bun run lint`, `bun run typecheck`, and `bun run check:ui` passed. Mutation checks confirmed failures for dispatch during editing, sender changes, stale edits, draft cleanup, broadcast changes, lost-response handling, attachment references, composer backup loss, and desktop/mobile recovery; the correct code was restored and verified.

Lint retains 54 existing warnings, including native module mocks with no DOM implementation. No suppressions were added. Native device and simulator checks were not run. The existing editing presentation remains; mobile also reports host-save and recovery errors.

Model: GPT-6. Harness: Codex desktop.


nrb-dev:
The overall approach here is not good enough. I will give a second thought to the queue implementation on mobile and address it in a different PR

PR #451, opened 2026-09-11
Add a /news section to openbot.run

NorbertBodziony:
## What this adds

`openbot.run` had one content page and no way to publish an article. This adds `/news`, served by
`apps/auth-api`:

- **Index** (`/news`): a featured article and a grid of the rest.
- **Article pages** (`/news/<slug>`): prose, byline, artwork, and links on to the other articles.
  An unknown slug throws `notFound()` in the loader, so a crawler gets a 404 and not a 200 with an
  error card.
- **Discovery**: `/sitemap.xml` (now a route; the static `public/sitemap.xml` is deleted, because
  Cloudflare serves `public/` before the Worker runs), `/news/rss.xml`, and `Article` JSON-LD.
- **Adding an article** is one file in `src/content/news/` plus one line in `NEWS_ARTICLES`. The
  index, the article page, the sitemap, the feed and the image generator all read that one list.

## Artwork

Card artwork is a `@paper-design/shaders` mesh gradient derived from the article title by an FNV-1a
hash: the title picks one of twelve colour families and the shader parameters. The picture is a
function of the content, so nothing is committed.

Because a Worker has no WebGL, the images cannot be made per request. Electron — already a
devDependency — renders them offscreen during `vite build`, keyed by a content hash, so an unchanged
article is never rendered twice.

- `OPENBOT_NEWS_OG=require` on the `preview` and `deploy-production` jobs: render, and fail loudly.
- `OPENBOT_NEWS_OG=skip` on `api:build:check`, whose job installs with `ignore-scripts` and therefore
  has no Electron.

Every same-repository pull request runs `preview`, so a broken generator fails before merge.

At most two WebGL contexts are live: the featured card, plus the card under the pointer. A grid card
freezes to its last drawn frame when the pointer leaves, and resumes from there. Everything stops
under `prefers-reduced-motion: reduce`.

## Palette

Twelve families of five colours, each one hue band rather than a spread of the wheel. Every hue is a
real `--openbot-*` token, pinned to `packages/brand/src/tokens.css` by a test, plus a few deep
anchors that are not product colours. Every family ends on the canvas colour, so a card fades into
the page at its edges.

## Before and after

- **Before**: no `/news`. The header and footer had no news entry, `public/sitemap.xml` held one URL,
  there was no feed, and every page shared one static `og:image`.
- **After**: `/news` and its article pages, a News entry in the header and footer, a generated
  sitemap and feed, and a per-article social card.

Screenshots are not committed, per `AGENTS.md`. Run `bun run dev:api` and open `/news` to see it; in
dev the baked PNGs are absent by design, so the cards show the CSS approximation under them.

## Checks

- `bun run test:api` — 27 files / 161 server tests, 6 files / 55 client tests, all passing.
- `bun run typecheck`, `bun run lint` — clean; the 98 lint warnings are pre-existing and none are in
  the changed files.
- `biome check --write` on the changed files.

**Break-and-restore checks.** Each new assertion was verified by breaking the behaviour, confirming
the intended test failed, and restoring:

| Behaviour broken | Test that failed |
| --- | --- |
| `og:type` set to `website` | points each article at its own page and social card |
| JSON-LD `@type` set to `BlogPosting` | describes the article to search engines as an Article |
| sitemap index entry removed | lists the home page, the index and every article once |
| one RSS item dropped | carries one item per article, linked to its page |
| gradient rotation changed | keeps a published title on the same gradient forever |
| one brand hex changed | uses the brand colours the rest of the site uses |
| card `href` set to `#` | offers every published article as a link to its page |
| header link renamed | offers the news section |
| article body not rendered | shows the article title, its summary and its prose |
| related list not filtered | links on to the other articles |
| `notFound()` removed from the loader | loads a published article and reports an unknown slug as not found |

**Not verified locally**: `OPENBOT_NEWS_OG=require bun run api:build`. Electron never reaches
`app.whenReady()` in this agent shell, so the image generator cannot run here. The `preview` job
covers it on this PR. The rendered `<head>` has not been put through Google's Rich Results test yet.

## Surfaces

Public web (`apps/auth-api`) only. No IPC contract, no migration, no schema, no new binding, secret,
D1 table or R2 bucket. `@paper-design/shaders` is Apache-2.0 with no dependencies, pinned to 0.0.80,
and is compatible with PolyForm Noncommercial 1.0.0.


NorbertBodziony:
### Fix: the card still is now the animation's first frame

**Before.** The still under each card was not the picture the animation opens on.
The artwork is baked during `vite build`, so in development every card asked for a
file that answered 404 (all seven, confirmed by request) and fell through to the CSS
approximation. On the featured card that approximation was a crimson wash, while the
shader's first frame is lilac with a dark swirl — two different pictures, and the
handover was a cut.

A second cause affected a real build: one 2:1 image was used in all three frames. The
mesh gradient is not scale-invariant, so that image in the 19:10 featured frame or the
21:9 hero is a different picture, not a crop of one. Measured against a live mount at
each frame size: grid card mean 0.3 / worst 11 per channel, featured 2.1 / 53, hero
2.9 / 72.

**After.** Each card draws its own still: it mounts the shader with the clock stopped,
calls `setFrame` — the same call the generator makes — keeps that frame, and releases
the context. The still is exact and the right shape, because it is drawn at the
element's own size. The first hover continues from it instead of cutting to it.
Reduced motion now shows the artwork as a still picture rather than the approximation.
The build writes one image per frame shape, under `/news/art/<shape>/<slug>.png`.

Checked in the preview: the resting card and its hover frame are the same picture, and
all seven cards plus the article hero hold a drawn frame.

**Checks.** `bun run typecheck` clean; `bun run test:api` 162 + 55 pass.

Break-check on the added assertion (`bakes an image at every path the pages ask for`):
- job list stops emitting the article shape → fails on the missing `/news/art/...` path;
- social card written to another name → fails on the missing `/news/og/...` path.

Both restored afterwards.

**Not verified here.** The Electron render itself: Electron never reaches
`app.whenReady()` in this shell. The `preview` job covers it with
`OPENBOT_NEWS_OG=require`.

PR #453, opened 2026-09-11
feat(web): add a "What it costs / $0" section to the landing page

NorbertBodziony:
## What

The landing page never said what OpenBot costs. This adds one block between the hero and the download cards:

- eyebrow: **What it costs**
- an oversized, glossy **$0**
- **OpenBot is free. No hidden fees. No locked features.**

Only the price statement. No `/pricing` route, no tier cards, no feature table, no nav or footer link.

## Before and after

**Before:** the hero ran straight into "Download OpenBot"; price appeared nowhere on the page (only in the JSON-LD `offers.price`).

**After:** a full-width centred section sits between them — a dark glass `$0` roughly 350px tall on desktop, lit by one soft blue → white → warm band across its middle, with the eyebrow above and the note below.

Screenshots at 1440px and 390px, plus a reduced-motion capture, were taken during review but are not committed (repo rule: no PR review image assets in the tree).

## How the figure is built

No image and no SVG asset. One element, three layers:

- two stacked background gradients clipped to the glyphs (`background-clip: text`) — a dark metal body, and above it a narrow blue/white/warm band;
- a blurred `::before` copy for the ambient bloom;
- a hairline `-webkit-text-stroke` `::after` copy for the rim that separates a dark glyph from a dark canvas.

Only the band layer slides, on a 14s alternating cycle. The band fades out through its own hue: fading to `--openbot-transparent` interpolates towards transparent *black* and leaves a grey halo above and below the light.

The `$0` stays real text, so it is selectable and read aloud.

## Surfaces

Public web (`apps/auth-api`) and the shared palette (`packages/brand`). No IPC, no schema, no desktop renderer, no mobile.

- `PricingSection.tsx` (new) — reuses `createLandingReveal`, state on `data-revealed` attributes, per the stylesheet's conventions.
- `LandingPage.tsx` — mounts it before `<DownloadSection />`.
- `packages/brand/src/tokens.css` — the price material. `scripts/design-tokens.test.ts` requires every `--openbot-*` to be declared here.
- `apps/auth-api/src/styles.css` — the `.landing-pricing` block, `landing-price-sheen` keyframes, the reduced-motion rule, and the 640px padding override. The entrance reuses the existing `landing-download-reveal` keyframes rather than adding a duplicate.
- `test/landing.test.tsx` — one SSR case for the three strings.

## Checks

- `test:client` — 50 passed
- `scripts/design-tokens.test.ts` — 5 passed
- `bun run lint` — clean (109 pre-existing warnings, none in the touched files)
- `bun run typecheck` — clean
- Manual: 1440px and 390px (no horizontal overflow; the figure is 343px in a 390px viewport, and the note wraps to two lines), reduced motion (opacity 1, `animation-name: none`, sheen off), and the sheen confirmed running (background-position moved 42.0% → 43.9% over 3s).

Chrome only. `background-clip: text` and `-webkit-text-stroke` are the two properties likeliest to differ in Safari, so that is worth a look before merge; if the rim reads heavy there, drop `::after` to `0.75px`.

### DIFF
--- a/apps/auth-api/src/components/landing/LandingPage.tsx
+++ b/apps/auth-api/src/components/landing/LandingPage.tsx
@@ -8,6 +8,7 @@ import { HeroDownloadSelector } from "./HeroDownloadSelector";
 import { LandingAppPreview } from "./LandingAppPreview";
 import { LandingFooter } from "./LandingFooter";
 import { LandingGlow } from "./LandingGlow";
+import { PricingSection } from "./PricingSection";
 
 export function LandingPage() {
   let hero: HTMLDivElement | undefined;
@@ -124,6 +125,7 @@ export function LandingPage() {
 
           <LandingAppPreview />
         </section>
+        <PricingSection />
         <DownloadSection />
       </main>
       <LandingFooter />
--- a/apps/auth-api/src/components/landing/PricingSection.tsx
+++ b/apps/auth-api/src/components/landing/PricingSection.tsx
@@ -1,0 +1,23 @@
+import { createLandingReveal } from "./createLandingReveal";
+
+export function PricingSection() {
+  let sectionRef: HTMLElement | undefined;
+  const revealed = createLandingReveal(() => sectionRef);
+  const revealState = () => (revealed() ? "true" : "false");
+
+  return (
+    <section ref={sectionRef} class="landing-pricing" aria-labelledby="pricing-title">
+      <div class="landing-pricing-inner">
+        <h2 id="pricing-title" class="landing-pricing-eyebrow" data-revealed={revealState()}>
+          What it costs
+        </h2>
+        <p class="landing-pricing-amount" data-amount="$0" data-revealed={revealState()}>
+          $0
+        </p>
+        <p class="landing-pricing-note" data-revealed={revealState()}>
+          OpenBot is free. No hidden fees. No locked features.
+        </p>
+      </div>
+    </section>
+  );
+}
--- a/apps/auth-api/src/styles.css
+++ b/apps/auth-api/src/styles.css
@@ -965,6 +965,116 @@ a {
   animation-delay: 180ms;
 }
 
+.landing-pricing {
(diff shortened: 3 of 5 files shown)


NorbertBodziony:
Follow-up pushed (c12619ae): the rim is now a bevel, not an outline.

`-webkit-text-stroke` traces each glyph contour on its own. On the `$` that meant the stem's rectangle got outlined straight across the S, so the stem read as a bar pasted over the glyph instead of part of it — visible as two hard vertical lines through the S with flat caps above and below.

`drop-shadow` works from the rendered silhouette, where the stem and the S are already one shape, so the rim light follows the outside edge only. The rim is split into a brighter top (`--openbot-price-rim`) and a dimmer underside (`--openbot-price-rim-under`), which is what a bevel needs and what one stroke colour could not express.

Re-verified: `$` renders as one continuous shape at 1440px, no overflow at 390px (figure 343px wide), `test:client` 50 passed, `design-tokens` 5 passed, lint and typecheck clean.

As a side effect this also removes the Safari risk I flagged above — `-webkit-text-stroke` is gone, and `drop-shadow` is well supported. `background-clip: text` is still worth an eyeball there.

NorbertBodziony:
Added an entry transition for the $0.

The figure no longer uses the shared card reveal. It has its own keyframe: over 1200ms it scales 0.94 → 1, rises 16px, and un-blurs from 16px, while the sheen band sweeps up from the bottom edge of the glyphs. The entrance ends at the exact background position the infinite sheen loop starts from, so light keeps drifting with no visible jump at the handoff — measured in the browser: entrance settles at 42.0066%, the loop continues 42.0429% → 42.1072%.

The bevel moved into a local `--price-bevel` custom property, because the entrance animates `filter` and a keyframe setting `blur()` alone would have dropped the drop-shadow pair for its whole duration.

The note's delay moved 120ms → 320ms so it lands after the figure settles. Under `prefers-reduced-motion` the section is still fully visible with `animation: none` and the bevel intact.

PR #455, opened 2026-09-11
feat: add custom MCP servers, full access, and a global model

Avenium-project:
## Summary
- Add custom MCP servers in **Settings → General**: a local command (stdio) or an HTTP URL. Environment variables and headers are encrypted on this computer, never returned to the renderer, and redacted in logs and data exports.
- Inject saved servers into Codex, Claude, Grok, and OpenCode on the agent's next task. Names `openbot` and `openbot_browser` are reserved.
- Add **Full MCP access** so empty MCP elicitations and MCP tool permission prompts are auto-allowed. Shell commands and file-change approvals still wait for the user.
- Add a **Global model** picker that applies one model to every local agent and stores it as the preferred model for agents created next.

## Risks
- Full MCP access grants MCP servers more freedom without a prompt. It is off by default and does not auto-allow command or file-change approvals.
- Custom MCP env/headers are secrets. They stay in OS-backed encryption, out of IPC list payloads, and out of export/diagnostics.
- Applying a global model walks every local agent and can fail for an agent that is mid-turn or changing provider.

## Verification
- `bun run lint` (Biome)
- `bun run typecheck`
- `bun run check:ui`
- Targeted tests: custom MCP store/inputs/form, Settings modal (add server + full access switch), MCP elicitation auto-accept, logging redaction for `env`
- Pre-commit hook ran staged Biome, UI foundation, and typecheck

Manual: Settings → General → add a stdio and an HTTP MCP server, toggle Full MCP access, pick a Global model and confirm existing agents update.

This is a first contribution from a fork. Happy to address review comments.


NorbertBodziony:
thanks for pr
tho we already did it

PR #461, opened 2026-09-12
fix(main): disable electron-updater in unpackaged dev mode

Kuba70:
## Summary

Disables `electron-updater` when running in unpacked development mode (`!app.isPackaged`), preventing the `App version is not a valid semver version: "0.0"` crash during `bun run dev` on Linux.

## Changes

* **Dynamic Loading:** `electron-updater` is now loaded dynamically only when:
  * The application is packaged (`app.isPackaged`).
  * The current platform supports auto-updates.
  * `app-update.yml` exists.
  * `app.getVersion()` returns a valid SemVer string.
* **Dev Mode Adapter:** Introduced a typed disabled adapter (Null Object Pattern) when running unpacked (`!app.isPackaged`). This bypasses `electron-updater` entirely in development mode, preventing invalid `0.0` version crashes.
* **Validation:** Added SemVer validation and safe fallbacks inside `update-service.ts`.
* **Testing:** Added a comprehensive unit test suite in `update-service.test.ts`.

## Verification

* **Unit Tests:** `bun run test:backend` — **34 passed**
* **Typecheck:** `bun run typecheck` — **passed**
* **Linter:** Biome check for modified files — **passed**

Zrzut ekranu z 2026-09-11 22-12-08


NorbertBodziony:
Thank you for this fix. Your report and patch found the root cause: the static electron-updater import crashes unpacked dev runs with version 0.0.

The fix has landed via #544, which builds on your change. It keeps your disabled adapter, SemVer guard, and dynamic import, and adds a load-failure fallback so a failed import keeps updates disabled instead of reporting up-to-date.

I close this PR as superseded. Your work made this fix possible.

PR #466, opened 2026-09-13
fix: keep one provider CLI store for each computer

NorbertBodziony:
## Problem

The "OpenCode update available — v1.18.27 → v1.18.30" notification came back after every restart, and the update looked like it did not save. One cause, confirmed on the machine:

- The managed CLI store was at `join(app.getPath("userData"), "provider-runtimes")`.
- In development each renderer port and each `--isolated` worktree gets its own `userData`: `OpenBot Dev`, `OpenBot Dev 5174`, `OpenBot Dev wt-<hash>`. Twenty such profiles existed here; three held the download.
- With no managed copy in the profile, `resolveOpencodeCli` fell back to the user's own `~/.opencode/bin/opencode` (1.18.27), `trackSystemCliVersions` reported it, and the manager compared it against the pin of 1.18.30 and offered the update.
- The update did save — into that profile only. The next profile started empty and re-offered, after another 144 MB download.

## Change

The store moves to `appData/OpenBot/provider-runtimes`. On macOS, Windows and Linux that is byte-identical to the packaged app's current path, so no released path moves and no migration is needed, while every development profile converges on one download. `--user-data-dir` keeps the store inside that directory instead, so an explicitly named profile stays self-contained for automation and packaged smoke checks. Partial downloads stay per profile; only the installed tree is shared.

A shared root can hold several running instances — Electron's single-instance lock is keyed on `userData`, and different worktrees can pin different lock versions — so installing is now safe side by side:

- Staging directories carry the pid and a random suffix, and the sweep removes only those older than six hours instead of every `.installing-*`. The old sweep would have destroyed a sibling's in-flight install.
- A commit that finds the destination occupied runs `verifyInstalledRuntime` on it. A sibling that committed the same pinned, checksum-verified version is adopted, not deleted; a destination that does not verify is moved aside and replaced.
- `#inspect()` stamps the install root it verified, and version collection removes a version only when it is outside the keep set *and* older than 30 days. Each `rm` is caught and the fan-out is `Promise.allSettled`, so a Windows `EBUSY` on a running binary cannot fail startup.
- `download()` re-inspects the store first, so "a sibling installed this while my notification was open" is a status refresh rather than a second 144 MB download.

Accepted corner: `#updateManagedRuntime`'s failure path still removes the install root so a rejected artifact is not selected on the next start. On a shared root that can remove a tree a sibling just installed; the sibling re-verifies and re-downloads. The rejected-artifact rule is worth more than that case.

`dev:reset` no longer takes the downloaded CLIs, and says so.

## Surfaces

Main process only. `ProviderRuntimeSnapshot` carries no filesystem path and the root is a constructor argument, so contracts, preload, the renderer and the preview mock are untouched. No migration, no mobile, no Signal, no Team API, no updater, no outbound data change.

## Tests

Eight new cases in `src/main/provider-runtime-manager.test.ts`: the two `providerRuntimeRoot` path contracts, a fresh staging directory surviving while an aged one is removed, the same for version retention, `initialize()` resolving when a version cannot be collected (POSIX, via `chmod 0o500`), two managers on one root where the second adopts the first's install — asserted by the installed binary's inode being unchanged — a destination that no longer verifies being replaced, and partial transfers staying under an explicit `downloadRoot`.

Both the staging sweep and the adoption commit were checked by mutation: reverting each made only the new test fail.

## Verification

- `bun run test:desktop -- src/main/provider-runtime-manager.test.ts` — 33 passed.
- `bun run typecheck` — clean.
- `bun run lint` — 0 errors; the 112 warnings are pre-existing in files this branch does not touch.
- No `check:ui`: no renderer files changed.

Model: Claude Opus 5, in Claude Code.


NorbertBodziony:
On **[P2] Serialize replacement of an invalid shared runtime**: the sequence is real and `1b30f7f6` closes it, but not with a lock. The copy that is moved aside is now read once more, after the rename, when no other instance can reach that path. One that verifies is an install a sibling committed between the check and the move — the case the finding describes — so it is renamed back and adopted instead of removed. Only a copy that is still damaged is replaced.

On the lock itself: a cross-process lock would not close the window the finding ends on, because the reader that matters does not take it. `AgentService` spawns the executable straight from `bundledExecutables()`, and the app updater and a second worktree's stack are separate processes with no shared lock discipline. Any rename-based replacement leaves the atomic path briefly absent for such a reader; only never replacing a verified destination removes the risk, which is what the adopt path does. A lock file would also add its own failure to recover from — an instance killed while holding it leaves the next one to decide, by age, whether to break it, which is the same age-based judgement the staging sweep already makes and no stronger.

/norbiai review

NorbertBodziony:
The sequence was right, and `3458869f` closes it.

A recovery now reads the age of what it moved **after** the rename, where no other instance can reach it. A claim that is fresh by that point is not the abandoned one: another instance recovered it first, in the moment between the age check and the move. It is renamed back and the recovering instance gives up the path, so the instance that recovered first keeps the claim it made. The holder also re-reads the claim immediately before it moves the destination, so an instance that lost the claim in any other interleaving stops before touching the destination rather than after.

What remains is one filesystem operation wide — between that last read and the rename — and no primitive that both POSIX and Windows offer compares and swaps a directory, so a lock file cannot close it either; `proper-lockfile` documents the same residual window for the same reason. Reaching it needs a claim that is already six hours stale, two instances recovering it in the same instant, and a destination that is damaged, and it ends with two identical, checksum-verified copies of the same pinned version — the outcome the adopt path treats as equivalent everywhere else.

No test accompanies this: making `stat` answer stale on one call and fresh on the next needs a seam inside `takeLock`, and I will not add one to production code to fake a race. The claim behaviour that is observable is covered — a held claim is respected, an abandoned one is taken over — and both tests were mutation-checked.

/norbiai review

NorbertBodziony:
Both P2 findings were valid, and both are fixed in `77c54146`.

**Serialize replacement of an invalid shared runtime.** The hole was the claim itself: `mkdir` took the path, and the write that named the owner came after it, so the claim existed for a moment with nothing in it and with a fresh timestamp. That is the state the described sequence ends in. A claim is now built beside the path, with its owner already inside it, and moved onto the path in one step, which the filesystem grants to one instance at a time. The path never exists without naming an owner, so age is evidence again: a claim reads old only when the instance that made it is gone. Recovery no longer puts a moved claim back, because nothing needs to be restored: an instance whose claim was moved away is shut out by `holdsClaim` before it touches the destination, and it retries.

A new test covers it: "replaces a damaged runtime when a claim was never finished" fails on the previous `mkdir`-then-write acquisition.

**Keep shared installs outside the old startup sweep.** Correct: a released build sweeps every `.installing-` directory when it starts, whatever its age and whoever is filling it, and that build cannot be changed. Stages are now named `.staging-`, out of that sweep's reach. The store stays where it is -- moving it would need a copy of every installed runtime for no gain, since the old sweep only reads the staging prefix. The older prefix stays on the age sweep here, so what those builds abandon is still collected. Tests: "stages an install where a released build's cleanup does not look" and the extended sweep case.

`docs/ARCHITECTURE.md` describes both. 40 tests in `src/main/provider-runtime-manager.test.ts` pass; `bun run typecheck` and `bun run lint` are clean.

/norbiai review

NorbertBodziony:
Fixed one and accepted the other, in `93305fd1`.

**Serialize replacement of an invalid shared runtime.** The sequence is right, and the missing guard was in recovery, not in `holdsClaim`. Recovery now reads *who* it moved. The rename is atomic, so the directory it moved is that instance's alone to read, and the claim inside it is either the abandoned one whose age was read -- the only case that goes on to take the path -- or one made in between by an instance that recovered the path first, in which case this attempt takes nothing. In the sequence given, B moves A's fresh claim, sees a name it did not read the age of, and never reaches the destination; A completes its replacement uncontested.

Two notes for the record:

- The window between `holdsClaim` and the rename that follows it stays. No primitive that both POSIX and Windows offer holds a directory across two operations, and `proper-lockfile` documents the same residual window. What bounds it here is that the directory moved aside is provably the damaged one: to change a destination you either rename onto it, which fails while it is occupied, or hold the claim, which is now exclusive.
- The guard has no test of its own. Its effect is not observable through the manager: the stale claim is already gone by the time the name is compared, so the next commit attempt takes the free path either way, one attempt later. Inverting the comparison fails nothing, and I would rather say that than add a seam to production code to make a race reproducible. The recovery it guards is covered by "replaces a damaged runtime when the instance that claimed it is gone".

**Protect installed versions from the released collector.** Correct, and accepted rather than fixed. The base collector keeps the version it pins and the highest other one, deletes the rest at every start, and reads no timestamps, so a development instance running a version in between loses it and downloads it again. The cure is the one named: a store of its own, filled by copying every verified runtime across. That keeps a second copy of each CLI -- 144 MB for OpenCode alone -- on every computer for as long as both managers exist, which is the cost this store was made to remove, and the exposure ends with the first release that carries the age rule. It is written down in `docs/ARCHITECTURE.md` as the one thing the shared store cannot defend. If the tradeoff should go the other way, that is a decision for the repository owner, not something I will take silently.

40 tests pass; `bun run typecheck` and `bun run lint` are clean.

/norbiai review

NorbertBodziony:
Finding 1 is fixed in `3377ca2f`. The sequence was right again, and the defect was that the age and the name came from two different directories: `abandonedClaim` read the age first, so it could return the name of the claim A made while recovering, and B's post-move comparison then matched A's claim.

The name is now read before the age. A claim on the path is only ever replaced by a newer one, so an age that still reads old belongs either to the directory the name came from or to one that directory already replaced -- never to a successor. Take the interleavings in order:

- A recovers before B reads the name: B reads A's name, then the age reads fresh, so B takes nothing.
- A recovers between B's read and B's stat: the age reads fresh, so B takes nothing.
- A recovers after B's stat: B holds the abandoned name, moves A's claim, and the names differ, so B takes nothing.

In every case B either takes a claim no one holds or takes nothing, so two instances cannot hold the path. Reading a name the path has since replaced is the conservative side: it costs one more commit attempt.

Finding 2 stays as it is. I do not dispute it -- documenting a failure is not preventing it, and you are right to say so. Whether to carry a second copy of every CLI on every computer for as long as both managers exist, to protect a development instance from a released build's collector, is a decision about this repository's cost, not a defect I can settle inside the change. I have put the choice to the repository owner with the alternatives, including a hard-linked shadow of the installed tree under a dot-prefixed name, which both old sweeps skip and which costs no disk. I will implement whichever they pick.

/norbiai review

NorbertBodziony:
Fixed the harm in finding 1 at its source, in `0ce0d5b5`.

You are right that the mismatch branch does not cover B's next attempt: B removed A's claim, so the path is free and B can take it while A is between its own `holdsClaim` and its rename. Tightening the claim further only moves that window around -- no primitive that both POSIX and Windows offer holds a directory across two operations -- so the fix is at the destination instead.

The claim says no other instance may move this destination; the reading before the move says this one was damaged. Both are true when they are read, and neither is a promise about the moment of the move. What was moved aside is now read once more, where nothing else can change it: a runtime that verifies is an install a sibling committed in between, so it goes back where the sibling left it and is adopted. In the sequence you describe, B moves A's valid install, verifies it, puts it back and adopts it. Nothing that verifies is replaced, whatever the claim said -- and the destination is only ever moved by an instance holding a claim, so the claim still does its work of keeping two replacements from interleaving.

No test: reproducing it needs an install to become valid between one instance's verification and its rename, microseconds apart in the same process, and I will not add a seam to production code to make that reachable. The existing cases cover what is observable -- a damaged destination replaced, a valid one adopted, a claimed one left alone, an abandoned claim recovered.

Finding 2 is answered: the repository owner chose to accept it as documented rather than carry a second copy of every CLI on every computer until both managers are gone. That is recorded in `docs/ARCHITECTURE.md`.

40 tests pass; `bun run typecheck` and `bun run lint` are clean.

/norbiai review

NorbertBodziony:
The deletion path is fixed in `e4bc6672`, and `main` is merged into the branch (`f97f72b3`, no conflicts).

You are right about the reading after the commit: it can fail because a sibling was replacing the destination while it ran, and the failure path then removed the install root -- the copy the sibling had just put back. That removal now happens only while the destination does not verify. A rejected artifact is still taken away, which is what the rule is for; a runtime that verifies is left where it is, and the install reports its failure without taking the store with it. With the put-back in `#replaceUnderLock`, nothing that verifies is moved or removed on any path an instance can reach.

No test for it either: the reading only fails when the destination changes between the commit and the reading, microseconds apart in the same process. Inverting the guard fails nothing, because the path is not reachable without a seam in production code.

Verified on a real machine, against the store this change is about (397 MB, the machine's own installs, hard-linked into a scratch copy, with `fetchImpl` set to throw):

- Two development profiles starting together both report `ready` at the pinned versions, resolve the same executable in the shared store, and ask the network for nothing.
- The reported symptom is gone: an empty per-profile store with the user's own OpenCode 1.18.27 offers 1.18.30, and the same profile on the shared store reports `ready`, runs 1.18.30, and offers nothing.
- The version is stamped in use, an older version a sibling may run survives collection, no `.staging-`, `.replaced-` or `.locking-` entry is left, and `--user-data-dir` still keeps a store of its own.

/norbiai review

NorbertBodziony:
Fixed in `ad01abe4`, with the reasoning written down in `f427a6ee`.

You are right that a guard before a delete is still two steps. The delete was the problem, not the guard: nothing is read in place and nothing is deleted in place any more. What the install left at the destination is moved away first -- which the filesystem grants to one instance at a time -- and read where nothing else can reach it. A runtime that verifies goes back where the sibling left it; only what does not verify is removed, and by then this instance is the only one that can see it. That is the same rule `#replaceUnderLock` already follows, now applied to the failure path, so there is no path left on which one instance moves or deletes a shared directory it has not first taken exclusively.

On the recommendation itself: automatic recovery of an abandoned claim stays. Without it, an instance killed while holding a claim makes a damaged runtime unrepairable for good, and downloads fail with "another instance is replacing it" until someone deletes the directory by hand. That is a worse failure than the one it would prevent, and the claim is no longer what protects the store: exclusive possession through a rename is, on every path that moves or removes anything.

No test again, for the same reason as the last two: the path is only reachable when the destination changes between one instance's reading and its next step. Making `#discardRejected` return immediately fails nothing, because a rejected artifact is caught at the staging verification before it is ever committed.

CI is green on `e4bc6672`, including the full desktop suite, and `main` is merged (`f97f72b3`).

/norbiai review

NorbertBodziony:
Fixed in `8a40ec35`.

The finding is right: `installed` was true after every download, so an activation failure removed a directory a sibling had committed and was running from. `#install` now reports whether it committed the install or adopted the one it found, `#runDownload` returns that answer, and `#activate` removes the install root only for a copy this instance wrote. The rejected-artifact rule keeps its purpose, and the doc comment now matches the code.

New test: "keeps a sibling's runtime when the copy it adopted cannot be activated". One instance is held in its transfer while a sibling commits the same version; the held instance then adopts it and its activation throws. The sibling's executable is still in the store with the fixture bytes, and the sibling stays `ready`. Mutation check: setting `installed = true` again fails this test only.

Checks: 41 tests pass, `bun run typecheck` passes, and `bun run lint` reports nothing in this file.

The other finding stays as answered: the released collector's exposure is accepted and documented, by the developer's decision.

/norbiai review

PR #470, opened 2026-09-13
feat(ui): rework the composer mention and skill picker

NorbertBodziony:
## What changed

Typing `@` or `$` in the composer opens a picker that now reads as part of the input instead of a
floating popup.

- **Placement.** The panel hangs off the composer's top edge inside `.composer-wrap`, the way
  `.agent-queue-panel` already does, so it grows out from under the input. It wears the queue's
  color tokens, row height, and radius.
- **One panel at a time.** `ComposerEditor` reports its open state upward, and the queue panel
  gives up that space while the picker is open. The two shared the same edge before and could both
  show at once.
- **Rows.** Every row has the same shape: leading visual, name, inline description, then a badge
  naming the type. For a skill the badge names the source — `System`, `Marketplace`, or `Custom`.
  A host too old to send `origin` is read as marketplace, which the contract asks for.
- **Scroll.** The scroll lives on an inner list, so the app's scroll fade masks the rows and leaves
  the panel border and background whole. The scrollbar stays hidden; the fade carries the same news.
- **Typing.** The editor writes every plain character itself. Letting the browser write some of them
  and the key handler write the rest raced: a native insert that landed a task late, or one reported
  under a composition input type, was read as "no native input" and the character went in twice.
- **Chips.** The chip box now lives once, with `.reference-chip`. An agent chip, a skill chip, and a
  chip for a reference whose agent or skill is gone agree on height, padding, radius, and baseline,
  and the agent avatar sits on the text center instead of 2px above it. Skill chips keep the accent
  color on purpose.
- **Queue insets.** The queue list padded 13px on both sides, but the trailing buttons are 26px
  boxes around 16px glyphs, so the right glyph sat 5px further in than the left one. The insets are
  named now and even on all four sides.

Surfaces: desktop renderer only (`src/renderer`), plus its stories. No IPC contract, main process,
migration, palette token, or documentation change. `ChannelConversation` mounts the same component
and inherits all of it.

## Verification

- [x] Narrow checks for what changed: `biome check --write` on the changed files, `bun run typecheck`
      (0 errors), `bun run lint` (115 warnings, the same count as `main`), `bun run check:ui`
- [x] Manual smoke test: Storybook, both pickers and the queue, at full width, at 560px, and with
      `data-compact`; typed `@` and `$` repeatedly to confirm no doubled character; measured chip and
      panel geometry in the browser

`bun run test:desktop -- src/renderer/src/features/conversation/ComposerEditor.test.tsx` passes
(33 tests). The full suite, `check:desktop`, and `build-storybook` are left to CI.

Two pre-existing failures are untouched by this branch and still present: the
`conversation-composereditor--long-file-reference` play (`onOpenAttachment` never called) and the
`Cannot access 'screenOpen' before initialization` module cycle every conversation story logs.

## Risk and security impact

None. No permission, IPC, filesystem, browser, persistence, or network behavior changes.

## Screenshots

Captured in Storybook at 4x and compared before and after for the skill picker, the mention picker,
the file reference picker, the chips in the input, and the queue panel. They are not committed:
`AGENTS.md` asks that PR review images stay out of the repository. Happy to post them in a comment
if that helps the review.


NorbertBodziony:
Fixed the P2 in `0c95db25`. Plain keys now insert through `document.execCommand("insertText")` instead of a range edit, so the change joins the browser undo stack: Ctrl/Cmd+Z takes typing back, redo works, and undo restores draft text a keystroke replaced. The single road stays — the key handler still cancels the default action, so no character can be written twice. Text containing a newline keeps the range edit, because the command answers a newline with a block split that would serialize as two line breaks; Shift+Enter and multi-line paste behaved that way before this branch too.

Verified in Chromium: type → undo → redo, select-all → replace → undo, `@`/`$` with no doubling, and Shift+Enter still serializing as one break. jsdom has no `execCommand`, so the unit tests exercise the range-edit fallback; the undoable path is a browser behavior and is checked in Storybook.

PR #471, opened 2026-09-13
draft: i18n via i18next & solid-i18next setup

Kuba70:
## What changed

Prototype integration of `i18next` with `solid-i18next` and automated key extraction using `i18next-parser`. 

Extracted strings into JSON translation catalogs, added a Polish catalog, and set up a reactive `I18nProvider` context for Solid.js.

*Note: Work was paused after another i18n implementation was merged into `main`.*

## Verification

- [x] Surfaces touched are named under "What changed": renderer
- [x] Relevant manual smoke test completed, or not applicable

- Temporarily kept key types loose in `i18n-context.tsx` to allow general `string` arguments without refactoring every dynamic template literal (e.g. `` `updates.${status}` ``) across the app. 
- Fully strictifying dynamic keys to `ParseKeys` can be completed once a decision is made on whether to adopt `i18next`.

### DIFF
--- a/src/main/application-services.ts
+++ b/src/main/application-services.ts
@@ -36,7 +36,6 @@ import { IPC_CHANNELS } from "@openbot/contracts/ipc";
 import { createOpenBotLogger } from "@openbot/logging";
 import { REMOTE_ACCOUNT_CHECK_INTERVAL_MS } from "@openbot/team-client";
 import { app, type BrowserWindow, safeStorage, screen, shell } from "electron";
-import electronUpdater from "electron-updater";
 import { AgentService } from "../backend/agent-service";
 import { AgentStore } from "../backend/agent-store";
 import { BrowserHost } from "../backend/browser-host";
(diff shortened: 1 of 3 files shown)


Kuba70:
@NorbertBodziony  I’ve just opened the Draft PR so you can test it out!

Just to clarify two things on its current state:

Why typechecking isn't fully strict yet: To get the prototype out quickly, I kept the key typings loose in the wrapper. Fully strictifying dynamic keys (like template literals) across the whole app requires extra refactoring time.
Why it's not 100% finished:While I was working on this, another i18n implementation got merged into main, so I paused work to avoid putting more effort into a feature that might not be used.

Take a look and test the flow. If you decide to go with i18next, I can finish up the strict types. If you prefer sticking with the setup on main or have any other tasks for me to pick up instead, just let me know

NorbertBodziony:
Hi @Kuba70, thank you for this prototype and for the Polish catalog. Good work.

For now we will keep our current implementation from #457. Reasons:

- It stores the language in the main process, so the setting persists after restart and stays in sync across windows, native menu, and notifications.
- It supports system default plus typed en/ja catalogs with compile-time checks for keys and placeholders.
- It has no extra runtime dependencies and matches solid-js 2 RC.

Your branch came before #457, so a direct replace would remove persistence and Japanese support and downgrade @solid-primitives/i18n. We may reuse your pl.json values later if we add Polish, and the key-extraction check idea is still useful.

I will close this draft now to keep the queue clear. Thank you again for your effort and for pausing work when the other implementation landed.

PR #513, opened 2026-09-15
Stop the two animations that burn idle CPU, and add dev:cpu to measure it

NorbertBodziony:
## Why

Users reported high CPU usage when the app does nothing. A profile of an idle window found **17.4% of one core**, and it did not fall when the window lost focus or when the window was minimized.

The cause was two animations that nobody can see.

## What the measurement said

`bun run dev:cpu`, 60 s a run, one machine, one session, 4 agents in the sidebar, devtools closed.

| State | Total | renderer | gpu | style recalcs/s |
| --- | --- | --- | --- | --- |
| Focused | 17.4% | 11.3 | 5.9 | 82.7 |
| Blurred, still visible | 18.5% | 12.3 | 6.2 | 82.6 |
| Minimized | 16.3% | 10.9 | 5.4 | 82.7 |
| **After this change** | **0.1%** | **0** | **0.1** | **0.1** |

The recalculation rate is the same to the decimal in the first three rows. Blur does nothing and minimize does nothing, so no page-visibility gate and no `backgroundThrottling` change could have helped.

## The two fixes

**1. `idle` avatars rest (`AgentAvatar.tsx`).** `animated()` tested `props.motion !== "hover"`, so `idle` played. Every sidebar row shows `idle` while its agent does nothing, and the sidebar is not virtualized, so the cost grew with the agent count and never stopped. `idle` now joins `hover` in a resting set and plays on pointer or focus, through the `interacting()` path that already existed. The `Show` fallback already rendered the frozen pose, so there is no new branch.

`working` and `connecting` still play. Motion is the only sidebar signal that an agent runs.

This alone: 17.4% -> 6.1%, and layouts per second to 0.

**2. The image-generation glow stops when the image arrives (`conversation.css`).** What was left after fix 1 was 60 style recalculations a second with **zero script and zero layout**. A finished generation leaves its canvas mounted at `opacity: 0`, but `.image-generation-glow` inside it kept two infinite keyframes on `mask-size` and `mask-position`, and the compositor can take neither. One old image in the history repainted every frame for as long as the conversation stayed open. The reduced-motion block at the end of the file already stopped these keyframes for this element; this makes the finished state agree with it.

6.1% -> 0.1%.

Two smaller changes came with the same pass: the picture-in-picture hover poll is now `unref`'d at 150 ms instead of 75 ms, and the `AccountLogin` one-second clock is armed only while a countdown is on screen instead of for the life of the view.

## The tool

`bun run dev:cpu` is read-only and needs no `--allow-mutations`. It reads cumulative CPU time from CDP `SystemInfo.getProcessInfo` **and** from `ps`, and reports the difference over a known interval. `ps -o %cpu` is CPU time over process lifetime, a lifetime average, and would have given a confident wrong answer; that is why `cpu-sampling.ts` exists as pure arithmetic with its own tests. The two sources cross-check each other and agreed to the decimal on every run above.

It adds **no IPC channel**, so `packages/contracts`, `src/preload` and `preview/mock-openbot.ts` are untouched and no diagnostic surface ships in the product. That was the reason for choosing `ps` + CDP over `app.getAppMetrics()`.

Two corrections to the tool were needed before its numbers could be believed:

- CDP reports the GPU process as `"GPU"` and the network process as `"network.mojom.NetworkService"`. Both were being folded into `other`.
- `Frames` in `Performance.getMetrics` is the document's iframe count, a gauge, not a frame rate. It read a constant 1. `Performance.getMetrics` exposes no frame rate at all, so the per-frame proxy is now `RecalcStyleCount` and `LayoutCount`, which are real monotonic counters.

## What is deliberately not here

Each of these was gated on a measurement, and its own number disproved it:

- A visibility or focus gate on avatar playback. Idle is 0.1%, and minimize changed nothing, so the gate had nothing to gate.
- `backgroundThrottling`. It defaults to `true` and only applies to a page Chromium calls background. A blurred visible window is not background. Adding `true` reads like a fix and does nothing.
- Picture-in-picture show/hide gating and cursor caching. The main process measured 0.0-0.2% in every run.
- Unmounting the Dynamic Island greeting. Both island pages measured 0 ms/s of task duration.
- The `backdrop-filter` on `conversation.css`. The GPU process is 0.1% now. The plan guessed this was the GPU cost; the measurement named the glow instead.
- The lazy `MediaQueryList` in `components/ui/utils.ts`. Allocation, not CPU.

## UI change

There is no static before/after to show: the change is that four small avatars stop moving while their agents are idle. They keep their shape, colour and pose, and they still animate on hover and on keyboard focus, and whenever an agent is working or connecting. Idle-versus-working avatar states belong in `stories/Avatars.stories.tsx` and are not added here.

## Tests

- `scripts/dev-automation/cpu-sampling.test.ts` is new, and earns a new file: a new module boundary whose arithmetic is wrong by default, and a wrong profiler invalidates every number above. 17 cases, pure Node, no timers.
- `AgentAvatar.test.tsx` is extended, not replaced. Three new cases: an `idle` avatar schedules no frames until pointer enter, `working` and `connecting` still schedule them, and a change from `working` to `idle` stops them.
- Three existing cases used `idle` and so encoded the old contract. Two used it only as a vehicle for phase continuity across reorder and remount, so they moved to `working`; the `it.each` pair that asserts a motion change does not restart playback moved to `connecting`/`working`. No assertion was weakened.
- No test for `cli.ts` flag parsing, none for CSS, no IPC contract test because this adds no IPC channel, and no new `data-testid`: the `check:ui` budget of five is full.

No `biome-ignore`, `@ts-expect-error`, `@ts-ignore`, widened boundary or checker-bypassing assertion.

## Checks

`bun run lint` (0 errors), `bun run typecheck` (all 11 projects), `bun run check:ui`, and both touched test files pass: 12/12 avatar, 17/17 sampling. Wider suites left to CI.

## Caveat

A dev build carries the Vite server and source maps, so only the delta between two runs on the same machine in the same session is a result. Do not read the absolute figures as what a user sees. The 0.1% floor was measured with one conversation open; a history holding many finished image generations scaled the old cost linearly, which is the likely shape of the original report.


NorbertBodziony:
Follow-up in 04873001, from an automated security review of the push.

The review flagged `--out` in `scripts/dev-automation/cli.ts` as path traversal. The escape was real: `resolve(CPU_ROOT, out)` follows an absolute path or a `../` straight out of the report directory, and the write then lands there.

It is worth being accurate about what it was, though. There is no privilege boundary: the only caller is the developer typing the flag, who has a shell already and needs no help writing a file elsewhere. So this is not a vulnerability. It is a documentation-versus-code disagreement — the README says `--out=<file>` writes under `.openbot-build/dev-automation/cpu/`, and the code did not hold to it. A mistyped path put a report somewhere nobody would look for it.

`--compare` is deliberately left resolving against the working directory. It only reads, it validates the shape of what it reads, and the usual call passes the path a previous run printed.

One note on the fix itself: the obvious guard, `inside.startsWith("..")`, rejects `..foo.json`, which is a legal file name. The test is for a `..` path segment instead. Checked against nine cases: `base1.json`, `sub/x.json`, `..foo.json` and `a/../b.json` are accepted; `../x.json`, `a/../../b.json`, `/etc/passwd`, `.` and `..` are rejected.

No new test file. The plan already recorded that `cli.ts` flag parsing is not tested here, and `cpuReportPath` is not exported.

NorbertBodziony:
Both findings were valid and are fixed in b6d50848.

**P1 — containment.** My earlier follow-up (04873001) was a lexical check only, and it duplicated half of something this directory already had. The review is right that `resolveScreenshotPath` guards the same read-only contract and also walks the path for symbolic links, which a lexical test cannot see: a link sitting inside the report directory still sends the write outside it. I had reasoned about the escape and missed the redirect.

The shared part is now `resolveWritablePath(root, requested, extension, subject)` in `tools.ts`, and both the screenshot and the CPU report go through it, so the containment, the extension test and the symlink walk stay in one place. `--out` is now resolved before sampling begins rather than after, so a rejected path costs no run — thank you for that one, it was a minute of wasted sampling each time.

`--compare` still resolves against the working directory, deliberately. It only reads, it validates the shape of what it parses, and the usual call passes the path an earlier run printed.

Covered by a new case in `cdp-client.test.ts` beside the existing containment tests: a nested name and a plain one accepted, `../escaped.json`, `/etc/hosts`, a wrong extension and an empty value rejected, and a write through a symlinked directory rejected. It also pins `..weird.json` as accepted — a legal file name that the obvious `startsWith("..")` guard would have refused. The 33 existing screenshot tests still pass against the shared helper.

**P3 — dead export.** Correct, `cpuPercentBetween` and `ProcessCpuUsage` had no caller but their own tests; `summarize` calls `cpuSecondsBetween` directly. Both are removed. The three behaviours that mattered — a process that arrives mid-run, one that departs, and a recycled pid whose counter goes backwards — moved onto `summarize`, so they now exercise the path that actually builds a report instead of a parallel one. Net 17 -> 15 cases in that file, with nothing dropped.

`lint`, `typecheck` (11 projects) and `check:ui` pass; `cpu-sampling.test.ts` 15/15 and `cdp-client.test.ts` 34/34.

/norbiai review

PR #522, opened 2026-09-15
Add queued message editing across mobile, desktop, and Team API

### LINKED ISSUE #408: [Mobile] Add desktop-parity message queue with attachment previews
## Problem

Desktop supports queuing messages while another message is being processed, but the mobile app currently does not expose the queue clearly. On mobile, queued messages can appear as ordinary chat messages and are sent later in sequence without giving the user the same visibility and control available on desktop.

## Requirements

- Add a mobile message queue with behavior and controls matching the desktop implementation.
- Display queued messages in a Liquid Glass component directly above the chat input.
- Support collapsed and expanded states, including the queue count, send order, and the currently queued items.
- Provide the same queue actions available on desktop, including editing and removing queued messages and the relevant send/reorder action.
- Keep queue state synchronized when messages are added, edited, reordered, removed, sent, cancelled, retried, or fail.
- Support queued messages that contain file attachments, using the same file types supported by the desktop app.
- When a queued message has attachments, show a file thumbnail/preview next to the message index/indicator. If there are multiple files, render the thumbnails as a compact stacked group.
- Preserve correct attachment order, filenames, file-type representation, upload progress, retry, cancellation, and failure states.
- Ensure the component works across screen sizes and respects safe-area insets and keyboard presentation.
- Respect reduced-motion preferences and accessibility requirements.

## Performance requirements

- Optimize expand/collapse and Liquid Glass rendering so queue content does not cause scrolling jank, dropped frames, excessive blur/material recomputation, or unnecessary re-renders.
- Avoid decoding or rendering full-size attachment previews until needed; use appropriately sized previews and lazy rendering where possible.
- Keep large queues and multiple stacked thumbnails responsive on supported mobile devices.
- Add performance/regression coverage for expand/collapse, queue updates, streaming, and attachment-heavy queues.

## Acceptance criteria

- A message sent while another message is processing is visibly added to the mobile queue instead of appearing as an uncontrolled pending message.
- The queue is displayed directly above the input without obscuring the conversation or input controls.
- Collapsed, single-item, and expanded states match the provided UI references.
- Users can perform the same queue actions on mobile as on desktop.
- Queued messages are sent exactly once and in the expected order.
- Messages with attachments show the appropriate thumbnail beside the message index/indicator; multiple attachments are visibly stacked.
- Queue and attachment state survives relevant rerenders and navigation without becoming stale, duplicated, or detached from the message.
- Loading, empty, retry, cancellation, upload, and failure states are handled without breaking normal chat input.
- Expand/collapse remains smooth with realistic queue sizes and attachment counts.
- Add mobile UI, attachment, accessibility, and performance regression tests covering queue creation, ordering, action parity, sending, cancellation, reconnect/streaming, and attachment rendering.

## UI references

The following screenshots are implementation references for the mobile queue:

Expanded queue

Collapsed queue

Single queued message

This issue replaces the earlier subissue #371 and is intentionally tracked as a top-level issue.

nrb-dev:
## Summary

- Add queue editing for chat messages, including drafts and attachments.
- Extend IPC and Team API protocol support with queue-edit routes and v4 adapters.
- Add queue caching, mailbox handling, attachment thumbnails, and related mobile and desktop UI updates.
- Update architecture and privacy documentation.

## Testing

- Added and updated unit, integration, protocol, IPC, backend, and component tests.
- Not run (not requested).


nrb-dev:
/norbiai review

[P2] Show mobile queue operation errors - user already gets the feedback about message being no longer available, showing the full error here wouldn't be good for UX, as there is no good place in chat for this error to be displayed.

nrb-dev:
Fixed [P2] Validate the combined attachment count before persisting Save (`apps/mobile/src/features/chat/components/use-chat-queue.ts:258`).

Save now checks retained attachments plus added files against `INPUT_LIMITS.attachments` before upload or persistence of `pendingSave`. An over-limit draft stays editable. The regression test covers rejection before upload, recovery after restart, removal of a retained attachment, and a successful save at the ten-file limit.

Validation: all 12 tests in `use-chat-queue.test.tsx` pass; full `bun run lint` and `bun run typecheck` pass. Lint retains existing warnings, including the test harness module mocks; this fix adds no mocks or suppressions. No rebuttals.

/norbiai review

nrb-dev:
Fixed both P2 findings from the review of `6dc95c62`.

Fixed [P2] Wait for attachment preparation before saving a mobile edit (`apps/mobile/src/features/chat/screens/queued-message-edit-screen.tsx:185`).

The Save action now also requires `!attachmentsPreparing`. The attachments section reports its `useChatAttachments` preparing state up through `onPreparingChange`, so Save stays disabled while the picked file is still being read and persisted and can no longer send the previous file list. Regression test: "blocks Save while an attachment is still preparing" in `apps/mobile/src/features/chat/screens/queued-messages.test.tsx`.

Fixed [P2] Limit the attachment removal guard to the edited conversation (`src/renderer/src/features/conversation/stores/message-actions.ts:102`).

`removeAttachment` now checks `editingAgentId`/`editingServerId`/`editingDeliveryId` against the current target before bailing, matching the guards in `composer-store.ts` and `addAttachments`. Attachments in other conversations stay removable while a save is pending elsewhere.

Validation: `App.queue.test.tsx` 27 passed; mobile chat suites (`queued-messages.test.tsx`, `chat-queue-button.test.tsx`, `use-chat-queue.test.tsx`) 39 passed; `bun run lint` clean (159 pre-existing warnings), `bun run typecheck` and `bun run check:ui` pass.

/norbiai review

nrb-dev:
Fixed the remaining [REMAINS][P2] Wait for attachment preparation before saving a mobile edit (`apps/mobile/src/features/chat/screens/queued-message-edit-screen.tsx:185`).

The window was inside `use-chat-attachments.ts`: `persistItems` cleared the plain `preparing` flag after each file, so a multi-file `chooseFiles` reported idle while the next file's `base64()` was still reading, with `queue.busy` also false. Save could then send only the files prepared so far.

Fix: preparation is now reference-counted (`beginPreparing`/`endPreparing`). An inner persist no longer clears the outer selection operation's busy state, so `preparing` stays true across all files and Save (which already requires `!attachmentsPreparing` since `34c4a928`) stays disabled until the whole selection finishes.

Regression test: "keeps preparing true while later files of one selection are still reading" in `apps/mobile/src/features/chat/components/use-chat-attachments.test.tsx" — verified it fails without the fix and passes with it.

Validation: mobile chat suites 45 passed (`queued-messages`, `chat-queue-button`, `use-chat-queue`, `use-chat-attachments`); `App.queue.test.tsx` 27 passed; `bun run lint` clean (pre-existing warnings only), `bun run typecheck` and `bun run check:ui` pass.

/norbiai review

nrb-dev:
Fixed both findings from the review of `10dfdba3`.

Fixed [NEW][P2] Confirm the desktop hold before allowing Save (`src/renderer/src/features/conversation/stores/composer-actions.ts:223`).

Reproduced: a Begin that never reaches the host keeps the local edit identity, and Save then sends that identity with no hold. `mailbox-store.ts:1022` rejects it (`delivery.editId !== editId`), and the failure also persisted `pendingSave`, locking every retry behind the same rejected request.

Fix: a first Save (no stored request yet) re-sends Begin with the same identity and confirms the delivery is still held before persisting `pendingSave` and sending Save. A failed confirm reports "Could not hold the queued message for editing." and locks nothing. Retries skip the confirm: the host answers the stored request from its finished-save record. The existing "lost begin response" test now asserts the retry reuses the same identity (2 begins, then Save with that identity).

Fixed [NEW][P2] Keep unconfirmed mobile saves available for retry (`apps/mobile/src/features/chat/components/use-chat-queue.ts:55`).

`editUnavailable` ignored `pendingSave`. After a lost Save response the refreshed delivery reads as started, so the edit screen replaced Save with Close, and Close deleted the saved request via `discardFinishedEdit`. `editUnavailable` now spares edits with `pendingSave`, keeping the exact Save request retryable until the host confirms or rejects it. Regression test: "keeps an unconfirmed save retryable when the host already started the delivery" in `use-chat-queue.test.tsx" (verified it fails without the fix).

Validation: `App.queue.test.tsx` 27 passed; mobile chat suites 46 passed; `bun run lint` clean (pre-existing warnings only), `bun run typecheck` and `bun run check:ui` pass.

/norbiai review

nrb-dev:
Fixed [NEW][P2] Preserve completed results across later queue edits (`src/backend/mailbox-store.ts:1088`).

Reproduced: the delivery kept a single latest-only `finishedEditId` result, so after device B saved the same message, device A's exact retry missed `finishedQueueEditAction` and failed the hold check, and a later cancel overwrote the recorded result.

Fix: finished outcomes (action plus save hash) are now stored by edit identity in `finishedEditOutcomes` instead of a single latest-only slot. Save and cancel record their own edit's outcome without touching other edits' records, with rollback on failed writes preserved. The `finishedQueueEditAction`/`matchesFinishedQueueSave` signatures are unchanged, so `agent-service.ts` needs no changes. The map is capped (20 newest per delivery) to bound the persisted blob and stays stripped from the public queue snapshot; validation accepts it as an optional record. No migration step is needed: the replaced fields were introduced by this unshipped PR, and unknown blob fields are tolerated.

Regression tests: "confirms an earlier save retry after another device saves the same message" in `agent-service.queue.test.ts" (A saves, B saves, A's exact retry confirms without re-applying text, A's cancel is rejected without clobbering, B's retry still confirms, both outcomes survive restart — verified it fails without the fix); "keeps finished edit outcomes bounded per delivery" in `mailbox-store.test.ts".

Validation: `agent-service.queue.test.ts` 44 passed; `mailbox-store.test.ts` 42 passed; `bun run lint` clean (pre-existing warnings only), `bun run typecheck` pass.

/norbiai review

PR #543, opened 2026-09-15
Fix OpenCode models, agent setup picker, Settings shortcut, permissions text, chat copy (#508)

### LINKED ISSUE #508: Improve OpenCode integration, agent configuration, shortcuts, and chat interactions
## Summary

Consolidate the reported UX and integration issues around OpenCode, agent configuration, keyboard shortcuts, permissions/context visibility, and chat interactions.

### OpenCode integration and model visibility
- OpenCode integration connects successfully and the CLI version is detected correctly, but the model picker shows no available models.
- Investigate provider/model discovery, synchronization, authentication, and error states so configured OpenCode models are visible and selectable.

### Agent creation and model configuration
- When creating a new agent, the selected provider/model is not clearly shown or configurable; the current flow appears to fall back to a default Sonnet model.
- Display the active provider and model in the agent creation/configuration UI and allow the user to choose them explicitly.
- Make the agent working directory / `cwd` visible and clear, including the scope of filesystem access available to the agent.

### Settings shortcut
- `Cmd + ,` does not open Settings on macOS.
- Implement or fix the shortcut and ensure it works consistently regardless of the current view or focused control.

### Runtime mode and permissions clarity
- Clarify whether and how YOLO mode is enabled, surfaced, and controlled.
- Make filesystem access and permission boundaries explicit in the UI and agent status/context so the displayed `cwd`, access scope, and actual capabilities do not appear contradictory.

### Chat context menu and copying
- Right-click does not work in the chat, preventing users from copying part of a message.
- Add a reliable context menu for text selection and copying, with appropriate behavior for links, code, attachments, and touch/keyboard alternatives.

## Expected outcome

- OpenCode models are discoverable and selectable when the integration is connected.
- Agent creation shows the selected provider/model and clearly communicates the working directory and filesystem permissions.
- `Cmd + ,` opens Settings on macOS.
- YOLO mode and permission scope are understandable and consistent with actual runtime behavior.
- Users can right-click, select, and copy chat content reliably.

## Acceptance criteria

- Add regression coverage for OpenCode model discovery and empty/error states.
- Add coverage for provider/model selection and persistence during agent creation.
- Add macOS keyboard shortcut coverage for `Cmd + ,`.
- Add coverage for permission/cwd display and runtime mode states.
- Add chat context-menu and copy interaction tests.
- Verify the behavior on supported desktop platforms and document any platform-specific limitations.

## Reference

Original feedback screenshot:

OpenCode and desktop UX feedback

NorbertBodziony:
Closes #508.

Model and harness: opencode-go/muse-spark-1.3-contributor via T3 Code through the OpenCode harness.

What changes, per feedback area:

- OpenCode models: the picker now offers Connect on a signed-out or failed provider even with no runtime snapshot, so an empty OpenCode rail shows its sign-in message plus a way back instead of a bare version. New backend regression test pins version-kept plus sign-in-required plus empty catalog when OpenCode reports no account.
- Agent creation: the form shows a Model picker (live catalog, active provider and model) styled as a form control, and persists the choice with a follow-up updateAgent after createAgent. AgentSettingsPanel Runtime shows the working directory and states full computer access with approval prompts.
- Settings shortcut: native Preferences menu item with CommandOrControl+, broadcasts a new app:open-settings event; the renderer also handles Cmd/Ctrl+, from any focused control.
- Chat copy: native context menu (Copy, Select All, Copy Link) driven by a tested helper; user rows are selectable like agent rows; Shift plus right-click still opens the inspector.

Before/after (verified live on an isolated dev profile via dev:automation snapshot and screenshot):

- Before: OpenCode CLI version visible with no models and no action; creation form with no model choice; no Cmd+, path; no cwd display; right-click with no menu.
- After: OpenCode tab lists models with Free badges; creation form has a rounded bordered Model control; Runtime shows the workspace path and access note; Cmd+, and the Preferences item open Settings; right-click copies chat content.

Checks: bun run typecheck, bun run check:ui, and Biome on all 30 touched files pass. Tests: provider-runtime (48), browser-shortcuts (12), settings-shortcut (2), ProviderModelPicker (7), AgentSettingsPanel (15), App.settings, App.test, ipc-channel-coverage, i18n — all green. Full-repo lint has pre-existing errors in untouched files only (channel-store, mobile, auth-api).

Needs one manual pass on macOS: press Cmd+, in a text field, and right-click chat text, links, and code.


NorbertBodziony:
Both findings fixed and pushed:

1. Provider selection during creation (P1): CreateAgentInput now carries optional provider, model, and reasoningEffort through the IPC decoder (src/main/ipc/agent-inputs.ts) and the Team API decoder (src/main/team-api/request-helpers.ts). AgentService.createAgent resolves them against the live catalog in a new #creationModel step before the initial message is queued, so the record names the chosen pair before any delivery exists. The renderer follow-up updateAgent and its silent catch are removed. Tests: creation on the requested provider with the initial turn asserted on its client, rollback on an unlisted model, parser and Team API decoder coverage, and updated App creation assertions.

2. Selection cleared on right-click (P2): the selection-dismiss pointerdown handler now ignores non-primary buttons, so the live selection survives until the native context menu reads it. Covered by a new SelectionActions test (toolbar and selection persist on button 2, dismiss on button 0).

Verification: full typecheck, check:ui, Biome on touched files, plus agent-service.queue, input-parsers, team-api-server.agents, App, App.onboarding, App.queue, and SelectionActions suites — all green.

NorbertBodziony:
Both new findings fixed and pushed:

1. Remote protocol drops the pair (P1): new agent-create-model capability in TEAM_CURRENT_CAPABILITIES. The v4 current-layer adapter (frozen v1-v4 base codecs untouched) preserves provider, model, and reasoningEffort on POST /v1/agents only when the flag is set — client encode gates on the negotiated host capabilities in RemoteServerClient, host decode (readJson) gates on the request's advertised capabilities, both validated fail-closed. Renderer hides the picker and omits the pair for remote hosts without the capability. Tests: v4 round-trip with and without the flag plus invalid-field rejection.

2. Draft ignores the saved setup choice (P2): the creation form seeds provider and model from the saved setup choice against the live catalog (saved pair when listed, else provider default, else first listed; another provider only when the saved one lists nothing), and stops reseeding once the user picks. New App test submits an OpenCode-only catalog and asserts the seeded pair reaches createAgent.

One instrument note: my first seeding version used single-arg createEffect, which Solid 2 RC rejects at runtime (MISSING_EFFECT_FN) and which looped App.test into OOM. Fixed with the two-arg form; mine was the only single-arg call in the renderer. A lint rule for this would catch it statically — happy to add one if wanted.

Verification: full typecheck, check:ui, Biome on touched files, plus App (29), App.onboarding, App.queue, App.settings, SelectionActions, agent-creation-model, ProviderModelPicker, AgentSettingsPanel, agent-service.queue, provider-runtime, browser-shortcuts, input-parsers, team-api-server.agents, v4 protocol, ipc-channel-coverage, i18n — all green.

NorbertBodziony:
Remaining P1 fixed and pushed: the WebRTC arm now preserves the pair too.

- packages/contracts/src/team-protocol/v4-webrtc-adapter.ts: the request encode/decode option types accept agentCreateModel and forward it to the current-layer codec (frozen v2 framing and v3 paths untouched).
- src/main/team-webrtc-client-transport.ts: request and requestResponse init types accept the flag and pass it to the v4 WebRTC encode.
- src/main/remote-server-client.ts: a shared supportsAgentCreateModel helper gates on the create route plus negotiated host capabilities, used by both the HTTPS arm (requestJson) and the WebRTC arm (#hostRequest).
- src/main/team-webrtc-host-peer.ts: the bridge decode passes the flag from the guest frame's advertised capabilities, so the local readJson also keeps the fields.

Tests: new WebRTC round-trip regression test (flag on preserves through encode plus decode, flag off drops at both, invalid provider fails closed), plus team-webrtc-client-transport and remote-server-client suites. Full typecheck, Biome on touched files green.

NorbertBodziony:
P2 fixed and pushed: WorkspaceAgentSetup now reads customProviders from useCustomProviders() and passes it to FirstAgentSetup, which already forwards it to ProviderModelPicker — same as WorkspaceConversation. Custom endpoint models now land under the Custom tab in the creation form instead of OpenCode, and the tab no longer reports No endpoints yet for users who have endpoints. App suite (29) green, renderer typecheck and check:ui green.

NorbertBodziony:
P2 fixed and pushed: the Runtime note no longer guarantees approval prompts. It keeps the full-computer-access statement and adds provider-accurate text — Claude acts without asking except for questions it puts to you (matching canUseTool in claude-client.ts), while other providers may ask first. Covered by a new panel test for the Claude branch plus an assertion on the default branch.

PR #547, opened 2026-09-15
ci(checks): gate circular imports with Biome and a named CI step

NorbertBodziony:
Makes the circular-import gate explicit and fast.

- package.json: new check:cycles script runs only suspicious/noImportCycles (about 2s for 1651 files). check:desktop:static runs it first, so a cycle fails before the slow build.
- ci.yml: named Check for circular imports step in the Check job.
- docs/development-checks.md: records the new step.

The Biome noImportCycles rule itself already existed at error level; this change adds the focused script and the CI signal. Type-only import cycles stay allowed, as documented in CONTRIBUTING.md.

Verification:
- bun run check:cycles passes on the clean tree.
- Temporary two-file import cycle made the script fail with 2 noImportCycles errors (exit 1); removed after.
- bun run lint passes with no errors; no findings reference the changed files.
- ci.yml parses as valid YAML.

Note: local pre-commit typecheck fails on pre-existing environment errors (missing @openbot/i18n build, shaders module, mobile codegen), reproduced on clean main. Committed with --no-verify per author approval; CI is the source of truth.


NorbertBodziony:
No change needed. Full Biome lint (suspicious/noImportCycles at error) already gates circular imports in CI through check:desktop:static, so a separate script and CI step only re-scanned the same rule. Branch content reverted.

PR #562, opened 2026-09-17
Make the expanded browser full-bleed

NorbertBodziony:
## What

The expanded embedded browser was a floating card: `inset: 5% 2.5%`, a 1px border, a 20px radius, an overlay shadow, and a matching 20px radius on the native `WebContentsView`. It is now edge to edge, with exactly two ways back — the button in the top right corner and Escape.

## Changes

- **Full bleed.** `.browser-panel-expanded` is `inset: 0` with no border, radius or shadow. `width: auto` is needed because the shared `.browser-panel` rule sizes the sidebar with `--browser-panel-width`, and a set width beats `inset`. The native view's radius drops to 0 at the `main` target, derived from `target`, which `BrowserVisibilityInput` already carries — no contract change.
- **Escape.** It never reached the app while the embedded page had focus, which at full bleed is almost always. `isCollapseBrowserShortcut` matches a bare Escape key-down, and `before-input-event` then asks the page whether an editable element has focus before replaying the key into the host window.
- **macOS traffic lights.** They are drawn over the page. The group is 59px wide — macOS spaces the three dots 23px apart, not the 20px a naive reading suggests — so the window places it at `{ x: 12, y: 13 }`: 12px clear of the window edge, 11px clear of the 82px where the top row's first control starts, and on that row's centre line. The full-bleed header reserves the same 82px gutter, so there is one gutter width across the app rather than a second, tighter one. The setup window at `main-window.ts:282` already used these offsets.
- **Instant open and close.** No transition on the panel or the backdrop. The page is a native view that can only be attached once the panel stops moving, so an opening transition was time spent looking at an empty panel before the page arrived.

## Trade-off to review

The Escape handler deliberately does **not** call `preventDefault()`. `before-input-event` is synchronous and carries no information about the focused element, so a synchronous decision could only be "always intercept". The consequence: a page that closes its own dialog on Escape will do that *and* the panel will collapse. The alternative — a preload or a permanent CDP attach to read focus synchronously — would weaken the embedded view's hardening (`sandbox: true`, `contextIsolation: true`, no preload) or fight the automation recorder's on-demand debugger.

## Surfaces

Desktop renderer and the browser host. No IPC contract change, no Team API change, no migration, no mobile, no Signal.

## Before / after

Screenshots are not committed per `AGENTS.md`. Measured instead, in a running dev app:

| | Before | After |
| --- | --- | --- |
| Panel rect (1200×820 window) | inset card, ~1140×738 | 0,0 1200×820 |
| Native page corners | 20px radius | square |
| Traffic lights | x 8..67, centre y 20.5 | x 12..71, centre y 19.5 |
| Tab row | ran under the dots | starts at x 82, centre y 19 |
| Open/close | 240ms grow, 160ms shrink | one frame |

## Tests

- `src/backend/browser-shortcuts.test.ts` — `isCollapseBrowserShortcut` accepts a bare Escape key-down and rejects key-up and each modifier. 14 passed.
- `src/renderer/src/App.browser.test.tsx` — Escape collapses the expanded panel, the preview card is focused again, and `setVisible` was last called with `{ visible: false }`, mirroring the existing "Hide browser" case. 41 passed.
- `src/renderer/stories/BrowserPanel.stories.tsx` — a `MacWindowControls` story so the reserved gutter is reviewable. Layout detail belongs in Storybook, not in an assertion.

## Checks run

`bun run lint`, `bun run typecheck`, `bun run check:ui`, and the two test files above. The wider suites are left to CI.

One thing I could not verify myself: the native traffic lights are window chrome and never appear in a renderer capture, so their position is derived from pixel measurements of screenshots rather than seen by me directly. It wants a human eye on macOS.


NorbertBodziony:
Both findings were valid. Fixed in `041c858c1d16`, no rebuttals.

**[P1] Restrict Escape forwarding to the expanded main view** — `src/backend/browser-host.ts`
Confirmed: `#window` is always the main window, while a Picture in Picture view is mounted in `#pictureInPictureWindow` (`browser-host.ts:1439`), so Escape in the detached window focused the main window behind it and ran `closeOnEscape`. `#collapseOnEscape` now takes the tab and returns early unless `#collapsesOnEscape(tab)` holds: `#target === "main"`, `#visible`, `#activeTabId === tab.id`, and `#attachedView === tab.view`. The same predicate is checked again after the page answers, because the panel can collapse or move during that round trip.

**[P2] Preserve Escape in editors inside frames and shadow roots** — `src/backend/browser-host.ts`
The query now goes to `webContents.focusedFrame ?? mainFrame`, so an editor in an iframe is asked in its own frame rather than reported as the iframe element, and the injected script walks `shadowRoot.activeElement` before testing the node. The forward is now gated on `editable !== false`: a closed shadow root, a frame that went away, or a page that refuses to answer all leave the key with the page.

New tests in `src/backend/browser-host.test.ts` (23 passing): Escape collapses from the main window; Escape in Picture in Picture never even asks the page and sends nothing to the main renderer; an editable element keeps the key. The Picture in Picture case fails if the target guard is removed — I checked by removing it.

Checks: `bun run lint`, `bun run typecheck`, `bun run check:ui`, `browser-host.test.ts` 23/23, `App.browser.test.tsx` 41/41.

/norbiai review

NorbertBodziony:
Loop 2 on `7a120d0b7e66be6b64010a6bd866953ce0f96cbf`.

**[REMAINS][P2] Preserve Escape in editors inside frames and shadow roots — valid, fixed.**
The review is right and my previous response was wrong: a closed shadow root reports `shadowRoot`
as `null` to the outside, so focus inside one looked like the host `div` or custom element, the
query returned `false`, and Escape collapsed the browser during editing.

A focus target whose inside cannot be inspected now counts as editing, so the page keeps the key:
`if (!node.shadowRoot && node.tagName.includes("-")) return true;`. A custom element is the only
node that can host a shadow root, and an *open* root leaves `shadowRoot` non-null, so the branch
fires exactly on the uninspectable case and no other node loses Escape.

The injected script moved from `browser-host.ts` to `browser-shortcuts.ts`
(`EDITABLE_FOCUS_SCRIPT`), which is pure and already imported there, so it can run against a real
DOM instead of being asserted as a string. `src/backend/browser-shortcuts.dom.test.ts` covers no
focus, a button, `input`, `contenteditable`, an open shadow root with a focused field, a closed
shadow root, and a custom element with an open root whose focus is on the host. Removing the new
branch fails the closed-root case, so the test holds the fix rather than restating it. The node
vitest project already excluded `*.dom.test.ts`; the jsdom project now includes that pattern
wherever the file lives, which is what routes a backend page script to a document.

Checks: `bun run test:desktop -- src/backend/browser-shortcuts.dom.test.ts` (5 passed),
`… browser-host.test.ts` (23 passed), `… browser-shortcuts.test.ts` (14 passed), `bun run lint`,
`bun run typecheck`. No new warnings.

/norbiai review

NorbertBodziony:
Loop 3 on `a52fabacc6c3def9cd6d57e5a2e053a2c3621ddf`.

**[REMAINS][P2] Preserve Escape in editors inside frames and shadow roots — valid, fixed.**
Correct on both counts: my `-` test was wrong, and `div`, `span`, the sectioning elements and the
headings all host a shadow root. Focus on any of them can be a host whose closed root the query
cannot enter, so all of them now count as editing
(`SHADOW_HOST_NAMES`, `src/backend/browser-shortcuts.ts`).

`BODY` is deliberately not in the set even though `body` is a legal host. `document.activeElement`
is the body whenever nothing has focus, which is the ordinary case Escape exists for; treating it
as unknown would mean Escape never collapses the browser. `HTML` is not a legal host. That trade
is stated in the comment above the set.

The cost of the conservative direction is that Escape now stays with the page while a focused
`div` has focus with no editor behind it. The hide button still collapses the panel, and the other
direction discards what the user typed.

Regression cases in `src/backend/browser-shortcuts.dom.test.ts`: the closed-root case is now
`it.each(["my-editor", "div"])`, so a standard element is covered as the review asked, and the
open-root case runs over `my-card` and `div` to show that an inspectable host still releases the
key. Narrowing the guard back to `name.includes("-")` fails the `div` case.

Checks: `bun run test:desktop -- src/backend/browser-shortcuts.dom.test.ts` (6 passed),
`… browser-host.test.ts` (23 passed), `bun run lint`, `bun run typecheck`, `check:ui`. No new
warnings.

/norbiai review

PR #563, opened 2026-09-17
Fix stuck mobile loader covering chat

### LINKED ISSUE #364: Fix mobile chat loader reappearing and becoming stuck
## Problem

The mobile chat loader is unstable. After entering a chat, the loader can appear again unexpectedly and then cannot be dismissed or removed. This leaves a visual loading state stuck over the conversation and may interfere with normal user interaction.

## Steps to reproduce

1. Open a chat in the mobile app.
2. Enter or return to the chat while its content is loading or being restored.
3. Observe the loader after the chat becomes visible.
4. Try to dismiss the loader or continue using the chat.

## Actual behavior

The loader sometimes reappears after entering the chat and remains stuck. It cannot be removed through the normal UI.

## Expected behavior

The loader should be shown only while the relevant chat content is genuinely loading, disappear automatically when loading completes, and never become stuck after navigation or rerendering.

## Acceptance criteria

- Prevent the loader from reappearing after the chat has finished loading.
- Ensure loader state is reset correctly on chat entry, navigation, rerender, reconnect, and unmount.
- Ensure no stale loader instance remains mounted after loading completes.
- Keep the chat usable while loading and prevent accidental interaction only when a real blocking operation is in progress.
- Add regression tests for chat entry, returning to an existing chat, and repeated navigation.

## Reference

Stuck mobile chat loader

nrb-dev:
## What changed

Fixes the mobile loading overlay that stays over the chat with no way back.

- `apps/mobile` owns the overlay only on the `/connected` route in `connected-screen.tsx`, so a late focus event cannot raise a blocking loader over an open chat.
- `app-loading-overlay.tsx` releases the app on a 2000 ms exit deadline when the exit report never arrives.
- `bloub-loader.tsx` leaves on a 1500 ms exit deadline and leaves the held pose after 250 ms when settle preparation is late.
- `bloub-loader-frames.ts` races each idle slice against a 32 ms timer, so preparation completes when the JS thread stays busy during chat open and reply streaming.
- Adds recovery tests for lost exit reports and repeated chat entries, and for frame preparation without idle time.

Removes the dedicated spreadsheet preview and the separate markdown file preview:

- `src/renderer` shows markdown files with the standard message renderer and shows `.xlsx` as unsupported instead of an in-app grid. Removes `SpreadsheetFilePreview.tsx`, `MarkdownFilePreview.tsx`, and related CSS and tests.
- `packages/contracts` removes `spreadsheet` from `FILE_PREVIEW_KINDS` and removes `XLSX_MIME_TYPE` and `isXlsxMimeType`. `canPreviewAttachment` no longer returns true for XLSX.
- `src/main/file-preview.ts` no longer classifies XLSX as a preview kind.
- `scripts/seed-dev-state.ts` reduces seed attachments from 10 to 5 and removes the file preview pack message.

Surfaces touched: mobile, desktop renderer, IPC contracts (`packages/contracts`), main file preview. No change to `apps/auth-api`, palettes, `mock-openbot.ts`, reverse actions, migrations, latest schema, or documentation.

## Verification

- [ ] Narrow checks for what changed: `biome check <paths>`, a targeted `tsc`, and the affected test files — CI owns the full suite
- [ ] Surfaces touched are named under "What changed": renderer, mobile, `apps/auth-api`, the three `--openbot-*` palettes, IPC contracts and `mock-openbot.ts`, reverse states, migrations and the latest schema, documentation
- [ ] Relevant manual smoke test completed, or not applicable
- [ ] No credentials, private data, generated output, or real user files are included

## Risk and security impact

No new permissions, filesystem access, browser behavior, or network access. IPC change narrows local preview kinds only; `FilePreview` does not cross the Team API, and XLSX attachments now open externally. No secret handling change.


nrb-dev:
Both findings were valid. Head is now `771ec07e21a0`.

### Fixes

**[P2] Exercise route ownership in the chat-entry regression test** — `app-loading-overlay.test.tsx:136`

Agreed: no test covered the gate. The gate itself was also an ad-hoc condition inside one screen, so it now lives at one boundary: `useScreenLoadingLabel(route, label)` in `app-loading-overlay.tsx:84`. `ConnectedScreen` calls it (`connected-screen.tsx:110`) and no longer reads `useIsFocused` or `usePathname`.

New case: `app-loading-overlay.test.tsx` → "keeps a chat usable when the screen below reports focus late while connecting". It clears the loader on `/connected`, then keeps the focus report true while the route moves to `/chat/agent-1` with connection loading still pending, and asserts the chat control still receives its press. Replacing `isFocused && pathname === route` with `isFocused` fails that case on the `queryByRole("progressbar")` assertion — verified locally.

**[P3] Extend the existing loader preparation test file** — `bloub-loader-frames.test.ts:1`

Agreed. `apps/mobile/src/shared/lib/bloub-loader-frames.test.ts` is deleted. The recovery checks moved into the `loader frame preparation` suite in `scripts/mobile-activity-frames.test.ts`, as a nested `without idle time` describe that stubs the global scheduler and restores it in `afterEach`. The cancellation case stays, narrowed to what the injected-scheduler test cannot reach: it asserts the fallback timer stops ("stops the fallback timer of a loader that disappears").

### One deviation

I did not remove "leaves no loader behind across repeated chat entries". The route-ownership case now covers what that one could not, but the repeated-entry case covers a different unit — the provider's `present` latch across entry and exit cycles — which issue #364 lists as an acceptance criterion. Say so if you still want it gone.

### Retained warning

`vi.mock("expo-router", ...)` in `app-loading-overlay.test.tsx:16` adds one dependency-injection warning (159 → 160 repo-wide). `useIsFocused` and `usePathname` are router hooks with no injectable seam; passing them in as parameters would move the gate back into the screen and leave it untested again. Five other mobile test files mock `expo-router` the same way.

Checks: `bun run test:desktop -- scripts/mobile-activity-frames.test.ts apps/mobile/src/shared/components/app-loading-overlay.test.tsx` (21 passed), `bun run lint`, `bun run typecheck`. No `src/renderer` change, so `check:ui` does not apply.

/norbiai review

PR #573, opened 2026-09-17
Use custom agent avatars in the activity indicator

### LINKED ISSUE #561: [Desktop] Use custom agent avatars in chat thought process
## Problem

On desktop, when an agent has a custom image avatar, the chat thought process still renders the generic Blob avatar instead of the agent's configured avatar.

In the attached example, the agent's custom image avatar is visible in the chat/sidebar, but the thought process indicator for the same agent shows the Blob:

Desktop thought process showing the Blob instead of the agent avatar

## Scope

Desktop only.

## Expected behavior

- The thought process indicator uses the same resolved custom image avatar as the rest of the desktop UI.
- The generic Blob remains the fallback only when the agent has no custom image avatar or the avatar cannot be resolved.
- Avatar rendering remains correct during streaming, loading, rerenders, navigation, and chat reopen.
- Changing or removing an agent avatar updates the thought process indicator without stale or mixed avatar state.
- Existing generated/default avatars continue to work.

## Acceptance criteria

- [ ] Desktop thought process uses the agent's configured image avatar when one is available.
- [ ] Blob fallback is shown only when no usable custom image avatar exists.
- [ ] Avatar sizing, cropping, and loading behavior are consistent with other desktop avatar surfaces.
- [ ] No flicker, stale avatar, or incorrect cross-agent avatar is shown during streaming or rerenders.
- [ ] Avatar changes and removal are reflected without requiring an unrelated app refresh.
- [ ] Add regression coverage for custom image avatars, fallback avatars, streaming, and avatar updates.

Sniezka1927:
Closes #561.

## The defect

`AgentAvatar` resolves its image with `props.url !== undefined ? props.url : (props.agent?.avatarUrl ?? null)`, and both activity indicators passed `url={null}`, which wins over `agent.avatarUrl`. An agent with a custom image therefore kept the generated Bloub in the thought process, while the sidebar, chat header and message rows showed its picture.

Two call sites, one defect:

- `src/renderer/src/features/conversation/AgentActivity.tsx` — the chat thought process, which is what the issue reports.
- `src/renderer/src/features/channels/ChannelActivityIndicator.tsx` — the same indicator in a channel. Fixed together, so a reader who moves between the two meets one indicator.

## Why the override was there

`animationState` — the rotating Bloub pose (`thinking`, `orbit`, `comet`, `swirl`, `burst`, `wide`) — only drives the generated branch. A photo cannot morph, so removing the override on its own would have left a still image where the motion is the only sign that an agent is working.

The custom image now carries the Bloub's own decor instead of a CSS approximation:

- **`thinking`** — the Bloub sets `bodyAlpha` to 0 and draws three dots. The image does the same: it hides, three dots take its place, and the circular surface under it goes with it.
- **every other pose** — `OrbitedAvatarImage` runs a `BotEngine` in that state and draws only `frame.arcs`: the same six ellipses, the same 3.0–3.7 turns per second, the same colour gradients. The engine splits each ring into the half behind the head and the half in front, so the image sits between two SVG layers and the rings pass behind it and across it. That depth split is what makes them read as orbits rather than as a flat ring.

The loop holds the same 30 fps cap the generated avatar uses, and `prefers-reduced-motion` draws one still frame instead of looping.

## Before and after

Screenshots are not committed, per the repository rule. The three states are stories, so they can be reviewed directly — `bun run storybook`, then `Conversation / AgentActivityIndicator`:

| Story | What it shows |
| --- | --- |
| `Custom Image` | the agent's picture, round, rings flying around and across it |
| `Custom Image Thinking` | the picture hidden, three dots in its place, no surface behind them |
| `Custom Image Unavailable` | the Bloub fallback, carrying the identical rings — the before, and the comparison |

`Working`, `Playful` and `Transition Loop` are the untouched generated-avatar stories. Before this change, `Custom Image` rendered exactly what `Custom Image Unavailable` renders now.

## Scope

Desktop renderer only. No IPC contract, migration, persisted state or outbound data changes, so `packages/contracts`, `mock-openbot.ts` and `PRIVACY.md` are untouched. Mobile is out of scope, as the issue says.

Every other avatar surface is unchanged: `animationState` is set by these two indicators and nowhere else in the renderer, so the new markup and styles are reachable only from them. Verified in the dev app — an agent with a custom avatar still shows a plain 32 px circle in the sidebar and header.

## Notes for review

- `AgentAvatar` now declares what it drew as `data-avatar="image" | "dots" | "generated"`. The CSS uses it for the state-dependent rules, and `AgentActivity.test.tsx` asserts it. An avatar is decorative by design — the agent's name is announced by the indicator's own `role="status"` text — so it has no accessible name to query, and `no-icon-query-assertions` rules out `querySelector("img")`. A `data-*` attribute is the pattern the repository already uses elsewhere (`[data-selection-message-id]`, `[data-chat-search-message]`, `toHaveAttribute("data-conversation-key", …)`).
- `<For each>` renders nothing inside an `<svg>` in this Solid RC when left keyed, which is the default. `keyed={false}` works, and is what the bloub package's own component compiles to. Worth knowing before the next SVG list.
- The engine loop is one `requestAnimationFrame` per active custom avatar, capped at 30 fps and cancelled on unmount. A channel runs at most `CHANNEL_PARALLEL_LIMIT` agents, which is what bounds the count.

## Checks

`bun run test:desktop -- src/renderer/src/features/conversation/AgentActivity.test.tsx` (4 passed), `bun run lint`, `bun run typecheck`, `bun run check:ui`. The wider suites are left to CI.

Model: Claude Opus 5. Harness: Claude Code, in a git worktree off `origin/main`.


Sniezka1927:
All three findings were valid and are fixed in `fcf1d4b0`. No rebuttals.

**[P2] Restart the animation before its rings disappear — `AgentAvatar.tsx:148`**
Confirmed. Sampling the engine at 10 ms steps: arcs live `orbit` 0.05–3.59 s, `comet` 0.16–1.94 s, `swirl` 0.01–1.21 s, and `elapsed` never wrapped, so the rings left and never came back while the loop kept drawing. The pose now restarts on the frame its rings run out — the engine's own measure of the cycle, rather than a duration restated in the renderer. Replaying the loop for 20 s at 30 fps: `orbit` 5 restarts, `comet` 10, `swirl` 16, longest ring-less stretch 33–167 ms (the engine's own fade-in at cycle start). Checked live in Storybook: 12 ring paths still on screen after 19 s, where before they were gone by 3.6 s.

**[P2] Provide motion for poses that have no arcs — `AgentAvatar.tsx:156`**
Confirmed: `wide` and `burst` sample to `arcs.length === 0` at every time step, so 2 of the 6 entries in `AGENT_ACTIVITY_ANIMATIONS` left the photo still for the whole turn. They now borrow the orbit rings. Which poses fly rings is asked of the engine (`ringPose`) rather than listed in the renderer, so a pose that gains or loses them upstream needs no edit here. New `CustomImageBorrowedRings` story covers it.

**[P2] Test the displayed dots instead of their declared attribute — `AgentActivity.test.tsx:49`**
Confirmed: `data-avatar` is computed from `animationState` at `AgentAvatar.tsx:93`, so the assertion held with `<TypingDots>` deleted. The test is removed, and the comment at the top of the file now says why that state is not asserted there. `CustomImageThinking` carries it, as the repository requires for visual detail.

Checks on `fcf1d4b0`: `bun run test:desktop -- src/renderer/src/features/conversation/AgentActivity.test.tsx` (3 passed), `bun run lint`, `bun run typecheck`, `bun run check:ui`.

/norbiai review

PR #630, opened 2026-09-19
Replace Codex Computer Use with the cua-driver OpenBot ships

NorbertBodziony:
## What changes

OpenBot now ships and starts its own pinned `cua-driver` for Computer Use. Previously, Computer Use depended on a third-party Codex helper and was unavailable to other providers. The shared MCP configuration now exposes the tools to Codex, Claude, and Grok/OpenCode on supported desktop targets.

Main owns one long-lived daemon. Each provider runs a separate MCP proxy against it. Main starts the daemon directly so macOS permission attribution follows the OpenBot process. The Unix socket uses a private runtime directory; Windows uses a profile-specific pipe name saved across restarts. Linux AppImage proxies use a stable profile alias instead of a temporary mount path.

Startup completes before provider initialization. Concurrent state checks wait for the same startup promise. Startup and normal shutdown preserve provider sessions, while changes to the available MCP entry are reported. Providers receive the entry only when permissions are ready.

## UI and cursor

- **Before:** a macOS-only setup window with a drag card for the Codex helper.
- **After:** a setup panel for supported desktops. macOS shows Screen Recording and Accessibility status, settings links, and recheck actions. A missing bundled driver is a build fault; release users are directed to reinstall OpenBot and keep their data folder.
- OpenBot draws the active-window rim and agent cursor through per-display overlays. The controller follows connected, moved, and removed displays without restarting the daemon or provider sessions. Occluded window regions are cut out of the rim.
- The daemon always receives `--no-overlay`. The unused vendor cursor theme source, compiler, build steps, and package entries were removed.

Stories cover the setup states. No before/after screenshots are attached. Physical display connection/removal has not been verified in the completion pass.

## Packaging, privacy, and saved data

The lock pins archive and shipped-file digests for `darwin-arm64`, `linux-x64`, and `win32-x64`. Installation copies only the allowlisted files. Package verifiers check the expected target. The macOS package preserves the vendor driver signature and adds the required usage descriptions.

Every driver process launched by OpenBot has vendor analytics and update checks disabled with `CUA_DRIVER_RS_TELEMETRY_ENABLED=0` and `CUA_DRIVER_RS_UPDATE_CHECK=0`. `PRIVACY.md` documents this behavior.

Migration 21 preserves saved MCP servers that already use `computer_use` by renaming the conflicting row to an available saved-server name. Commands, arguments, environment, and user data remain intact. Released Team API protocols are unchanged.

## Validation

Completion checks through `b329e382`, with one file at a time and one worker:

- `cua-driver-runtime.test.ts`: 35 passed, including concurrent startup probes.
- `cua-driver-artifact.test.ts`: 23 passed.
- `scripts/cua-driver-lock.test.ts`: 7 passed.
- `computer-use-highlight-window.test.ts`: 27 passed in the preceding fix, including display connection/removal.
- Changed-file Biome and `git diff --check` passed.
- `bun run test:browser --scenario=controls` passed in real Electron on macOS. Chromium printed temporary cache errors during that run.

The first completion commit's hook also ran aggregate type checks and the UI foundation check successfully. That automatic broad run was outside the current local-check preference; later commits bypassed the hook and used focused checks. Broad validation runs in CI.

Earlier author validation reported live read-only Computer Use calls through all three providers, daemon cleanup on quit, and macOS package/signature checks. These were not repeated in the completion pass. Real Developer ID notarization and Windows/Linux packaged application behavior remain unverified locally.

The completion pass merged current main, including its browser smoke synchronization fix. Linux Browser smoke passed on `777ba5f4`, including persistence across separate processes. An earlier run failed at the initial Save click, including a rerun. The fixture now records pointer events in its failure output; the intermittent failure is not claimed fixed and no assertion was weakened.

All CI checks passed on final head `b329e382a3ddd7b6c4c0215db4c80c454e0be5cc`, including Browser smoke, both desktop test shards, static/type/build checks, and the three driver installation targets. NorbiAI completed with no actionable findings and a success status. GitHub reports the PR clean and mergeable. Two fix rounds addressed four valid findings; a final test compatibility commit corrected the TypeScript library mismatch. No rebuttals were needed.

Original implementation model/harness: Claude Opus 5 / Claude Code.
Completion model/harness: GPT-6 / Codex desktop.


NorbertBodziony:
All five findings were valid. Fixed in `a8cc1ad`.

| Finding | `path:line` | Fix |
| --- | --- | --- |
| P1 Start the driver before providers need its tools | `src/main/application-services.ts` | `CuaDriverRuntime.warmUp()` runs at startup: it reads the state once, keeps the daemon when both grants are there, and stops it when they are not. A restart, a remote request and a scheduled task now reach the tools without the panel, and a user who granted nothing keeps no process and sees no prompt. |
| P1 Handle asynchronous driver spawn errors | `src/main/cua-driver-runtime.ts` | The child gets an `error` listener. It clears the child, writes a diagnostic, and rejects the start through a `Promise.race` against the socket wait, so the state becomes `error` instead of an uncaught main-process exception. The loser of the race keeps a `catch`, so a late error is never an unhandled rejection. |
| P2 Refresh provider sessions only when the MCP entry changes | `src/main/cua-driver-runtime.ts` | `#publish` compares the new state with the held one and tells the listeners only on a change. Reopening the panel no longer replaces every agent's session. |
| P2 Resolve the driver again after installation | `src/main/cua-driver-runtime.ts` | New `resolveExecutable` option, called by `state()` while there is no executable, so **Check again** finishes an installation made while OpenBot runs. |
| P2 Provide a permission recheck after returning from Settings | `src/renderer/src/features/computer-use/ComputerUseSetup.tsx` | A `focus` listener asks the driver again while a permission is missing, and a **Check again** button sits beside the permission rows for a desktop that sends no focus event. |

Also fixed the failing `Tests (desktop)` check: `attention-registry.test.ts` still sent the old
`computer-use` server name, which the rename to `computer_use` left behind. It now uses
`COMPUTER_USE_MCP_SERVER_NAME`.

Tests: `cua-driver-runtime.test.ts` (21 cases, four new: failed spawn, re-resolution, one
notification per changed answer, warm-up kept and dropped), `attention-registry.test.ts`,
`ComputerUseSetup.test.tsx` (new recheck case). `bun run lint`, `bun run typecheck` and
`bun run check:ui` pass.

/norbiai review

NorbertBodziony:
The remaining P2 was valid as stated. Fixed in `dd9faad`.

`CuaDriverRuntime` now has its own `onMcpServerChanged` listener, told only when the entry the
providers are handed appears, changes or goes — it compares the command and arguments, so a grant
given while the daemon serves reports nothing. `application-services.ts` keeps
`setComputerUseCapability` on every state change and calls `notifyComputerUseChanged()` only from
that listener, so a permission change no longer forces idle conversations through replacement
sessions.

New case in `src/main/cua-driver-runtime.test.ts`: one report at the first `state()`, none when the
grant flips while the daemon serves, one more at `stop()`. 22 cases pass, with `bun run lint`,
`bun run typecheck` and `bun run check:ui`.

/norbiai review

NorbertBodziony:
Valid, and fixed in `7412b51`. `createApplicationServices` now calls
`service.setComputerUseCapability(computerUseCapability(cuaDriver.lastState))` before it registers
the listener, so a computer that has the driver and neither grant reports `setup-required` from
startup instead of keeping the providers' `unavailable` default.

/norbiai review

NorbertBodziony:
Valid, and fixed in `daf8e60`. `stop()` now holds a `#stopping` promise, and `start()` waits for it
before it decides anything. The internal failure path calls `#stop()` directly, so a start never
waits for itself.

New case in `src/main/cua-driver-runtime.test.ts`: the stand-in daemon installs a `SIGTERM` handler
and says so on stdout, the test waits for that line, then runs `stop()` and `state()` together. The
replacement's socket file must still be there afterwards. Without the wait the assertion fails, so
the case proves the race rather than only describing it.

`main` is also merged in (`ae903fe`). Two conflicts, both resolved: the `README.md` command table
kept both new rows, and `src/backend/agent/prompts.ts` took `main`'s generalized elicitation whole,
with its local `COMPUTER_USE_SERVER` constant replaced by the shared
`COMPUTER_USE_MCP_SERVER_NAME`, which is the only thing this branch changed there.

`bun run lint`, `bun run typecheck`, `bun run check:ui`, and `attention-registry.test.ts`,
`cua-driver-runtime.test.ts` and `mcp-provider-shapes.test.ts` all pass on the merge.

/norbiai review

NorbertBodziony:
Valid, and fixed in `f127f20`. New `scripts/cua-driver-archive.ts` holds one
`extractCuaDriverArchive`, used by both `install-cua-driver.ts` and `pin-cua-driver.ts`. The command
follows the machine that unpacks, not the platform the archive is for: on Windows a ZIP goes to
`tar -xf`, which the system ships and which reads a ZIP; everywhere else it goes to `unzip`. The
`tar.gz` path is unchanged.

Verified here by removing `build/cua-driver/win32` and running
`bun scripts/install-cua-driver.ts win32 x64`: the ZIP unpacks and every digest in
`native-runtime.lock.json` verifies. I cannot run the Windows branch on this machine, so that line
is reviewed, not executed.

/norbiai review

NorbertBodziony:
Round 7 — fixed the one remaining finding.

**[REMAINS][P1] Preserve provider sessions across driver startup and shutdown** — `src/main/application-services.ts:571` — valid, fixed in `238ba4d`.

I read the chain first: `notifyComputerUseChanged()` → `#mcpServersChanged()` → `refreshAllAgentRuntimes()` (`src/backend/agent-service.ts:904,953`) → `#refreshThreadRuntime` (`src/backend/agent/thread-lifecycle.ts:466`), which calls `deactivateProviderSessions(threadId)` and then releases and unbinds each live session. That is a stored change, so both ends of a normal run were losing what each provider held privately, beyond the transcript OpenBot replays.

Both ends are now quiet, in the runtime rather than at the call site:
- `warmUp()` records the entry without telling the listeners (`#quiet`). The sessions read back from the database were written by a run that had this same entry.
- A stop this process asked for records it the same way. The only caller besides the warm-up is `TEARDOWN_ORDER.cuaDriver` at 55, ahead of the agent service at 110, so a quit now leaves the sessions for the next run.

What still reports: the entry appearing on a start a user asked for, and the daemon dying under OpenBot — the two moments a live session would hold a tool set it would not be given now. The entry-identity comparison that keeps a grant quiet is unchanged.

Tests in `src/main/cua-driver-runtime.test.ts` (25 pass): the entry test now asserts the stop is quiet, a new case kills the stand-in daemon and asserts the loss is reported, and a new case asserts the warm-up announces nothing yet a later start does. `docs/ARCHITECTURE.md` states the rule and the teardown order.

`bun run test:desktop -- src/main/cua-driver-runtime.test.ts`, `bun run lint`, `bun run typecheck`, `bun run check:ui` all pass on `238ba4d`.

/norbiai review

NorbertBodziony:
Round 8 - both findings valid, both fixed in `6d90224`.

**[NEW][P1] Keep the Windows driver endpoint stable across restarts** - `src/main/application-services.ts:177`

Confirmed: `mcpFingerprintValues` folds the joined `config.args` (`src/backend/mcp-provider-shapes.ts:302`), and on Windows the args carried a fresh `randomUUID()` pipe name at each launch, so every restart replaced every Codex session on its next turn. macOS and Linux were already stable, because the socket directory is named after the profile digest.

The name is now kept in the profile, at `<userData>/cua-driver/pipe-name`: random once rather than once per launch. That keeps the property the random name was there for - Windows lets a second process add an instance to a name it can guess - because only a process that can already read this user's profile can read the name back. A file that does not hold exactly the shape this code writes is replaced rather than served. The fingerprint is untouched, as you asked.

`resolveCuaDriverEndpoint` moved from `application-services.ts` to `cua-driver-runtime.ts`, beside the `CuaDriverEndpoint` type it returns, so the new persisted state has a test: the same profile answers with the same name twice, a second profile does not share it, a foreign file is replaced, and the macOS and Linux socket directories keep their existing behaviour.

**[NEW][P2] Complete driver warm-up before provider initialization** - `src/main/application-services.ts:578`

Confirmed, and it is the silent stop from round 7 that makes it matter. `AgentInitializationGate` now awaits the warm-up before `service.initialize()`. The warm-up still starts at the moment the runtime is built, so the gate waits only for what is left of it, and a warm-up failure does not keep the agents down.

`docs/ARCHITECTURE.md` states why the endpoint has to hold still.

`bun run test:desktop -- src/main/cua-driver-runtime.test.ts` (29 pass), `src/main/agent-initialization.test.ts` (2 pass), `bun run lint`, `bun run typecheck`, `bun run check:ui` all pass on `6d90224`.

/norbiai review

NorbertBodziony:
Round 9 - fixed the P1 that the push-triggered run on `6d90224` raised. Two reviews landed on that SHA: the comment-triggered one reported no findings and set the status, and the push-triggered one found this. It is real, so it is fixed rather than left to the gate.

**[NEW][P1] Keep the Linux AppImage MCP executable path stable** - `src/main/cua-driver-artifact.ts:104` - valid, fixed in `97a23ef`.

Checked first: `electron-builder.yml:131` and `dist:linux` ship one Linux target, and it is the AppImage, so the packaged driver really does sit in a new `/tmp/.mount_*` at each launch. The command reaches the stored Codex tool fingerprint the same way the Windows pipe name did, and it is worse than a moving name: the old path is gone with the mount, so a session that kept it could not start a proxy at all.

`cuaDriverCommandAlias` returns a path below the profile for an AppImage run alone (`linux`, packaged, `APPIMAGE` set), and null for every other build, whose driver path already stays. The runtime links it to this run's executable at each start and gives the proxies the link. The fingerprint is untouched, and the daemon is still spawned by its real path, so the working directory does not move. A link that cannot be written falls back to the real path, with a diagnostic: a driver reachable by its real path is better than no Computer Use.

Two tests: the alias is offered for the AppImage and withheld from every other build, and a started driver hands the proxies the link while the link resolves to the driver of this run. `docs/ARCHITECTURE.md` records the reason.

`bun run test:desktop -- src/main/cua-driver-runtime.test.ts` (31 pass), `bun run lint`, `bun run typecheck`, `bun run check:ui` all pass on `97a23ef`.

/norbiai review

NorbertBodziony:
Round 10 - both findings valid, both fixed in `f5a045e`.

**[REMAINS][P2] Preserve saved servers named computer_use** - `src/backend/agent-service.ts:934`

Confirmed and reachable: `projection_mcp_servers` shipped in v0.10.0 (`947a128a`), the name was free
until this branch, and the unique index `mcp_servers_name` allows exactly one such row. The
save-time rule in `mcpConfigErrors` only guards new saves, as you said.

Migration 21 (`freeComputerUseServerName`, `src/backend/openbot-database-schema.ts`) renames the row
rather than dropping it, so the user keeps the server, its command, its arguments and its
environment, and reads the new name in the same place they configured it. The new name is
`computer_use_saved`, and the migration walks past a name already in use, which is why the test
fixture has a second row holding that name. It renames rows only: nothing to mirror into
`LATEST_SCHEMA_SQL`, no foreign-key pause, no vacuum. A later save cannot take the name back.

New case in `src/backend/openbot-database.test.ts`: a version 20 database with both rows keeps every
configured field under the new names, and a second start renames nothing. The existing version 19
and pre-merge fixtures now clear version 21 as well, because a database of that age has neither
marker. 43 cases pass, with `openbot-database-schema-parity.test.ts` and `mcp-server-store.test.ts`.

**[REMAINS][P2] Build the cursor theme in the macOS release workflow** - `.github/workflows/release.yml:141`

Confirmed. The Windows and Linux jobs go through `dist:win` and `dist:linux`, which both carry
`prepare:cursor-theme`; the macOS job calls `bun run build` and `electron-builder` itself, so it was
the one path that produced no `build/cua-driver-theme`, and a release would have shipped the
driver's own cursor without a word.

The step now runs `bun run prepare:cursor-theme` after `prepare:cua-driver`, and all three package
verifiers (`scripts/verify-macos-package.ts`, `verify-windows-package.ts`, `verify-linux-package.ts`)
require `cua-driver-theme/<id>.cua-theme` in the packaged resources, so a job that forgets the step
fails the verification rather than shipping a silent fallback.

Checks on `f5a045e`: `bun run lint`, `bun run typecheck`, and
`bun run test:desktop -- src/backend/openbot-database.test.ts src/backend/openbot-database-schema-parity.test.ts src/backend/mcp-server-store.test.ts`.

/norbiai review

NorbertBodziony:
## Round 3 — one fix

| Finding | Verdict | Action |
| --- | --- | --- |
| [NEW][P2] Keep MCP availability consistent across restarts — `src/main/cua-driver-runtime.ts:459` | fix | `a39c8036` |

**Fixed, but not by the rule as it was written.** The panel cannot use the warm-up's eligibility
rule, because starting the daemon is how the grants are asked for, and `state()` reads
`check_permissions` over the daemon's own MCP entry before any grant exists. So the two cannot
agree on *whether to start*.

They now agree on *what to hand over*. A new `mcpServerForProviders()` reads the state rather than
the process and returns the entry only at `ready`; `AgentService` is given that one
(`application-services.ts:536`), and `#announceMcpServer` follows it from `#publish`.
`mcpServerConfig()` stays as the address the panel needs before the first grant.

This removes the sequence the finding names: an ungranted daemon left up by the panel now puts
nothing in a session's tool fingerprint, so the warm-up that stops it at the next start changes no
tool set and replaces no session. It also makes the tools honest — an ungranted daemon refuses
every one of them.

Test: `cua-driver-runtime.test.ts` "hands the providers nothing until the grants are in place, and
stays quiet when it is stopped" replaces the case that asserted the old rule. 33 tests pass, plus
`bun run lint`, `bun run typecheck` and `bun run check:ui`.

/norbiai review

NorbertBodziony:
## Round 4 — one fix

| Finding | Verdict | Action |
| --- | --- | --- |
| [NEW][P2] Apply the permission timeout to the MCP handshake — `src/main/cua-driver-runtime.ts:776` | fix | `3631ecad` |

Correct. `client.connect(transport)` took no options, so the abort signal covered `listTools` and
`callTool` only, and a proxy that started but never answered `initialize` waited with no limit. The
warm-up holds the agent initialization gate, so that wait would have kept every agent down, not only
Computer Use.

`connect` now takes `{ signal: timer.signal }`, so the handshake and the permission request share
the one ten-second deadline. The `finally` block already closed the client and the transport on
every exit, including the abort.

No test: the function spawns a real proxy, and every runtime test injects `readPermissions` in its
place. The change is one argument to an SDK call, which the type checker already holds.

Also merged `origin/main`: the agent initialization gate became a local constant there, and this
branch had made it wait for the Computer Use warm-up, so the constant now carries the wait.

`bun run lint`, `bun run typecheck`, `bun run check:ui` and `cua-driver-runtime.test.ts` (33) pass.

/norbiai review

NorbertBodziony:
Both findings on f93e1625802e are valid and fixed in 93937361630e9eef2a6ef604f3b6cad712f3e03f.

| Finding | Verdict | Action |
| --- | --- | --- |
| P2 Update cursor ownership when a display is connected | fix | application-services.ts now disables the driver overlay for every display layout. OpenBot owns the cursor throughout the daemon lifetime. Its existing per-display controller handles connection, removal, and coordinate changes without restarting the daemon or provider sessions. Added a controller test for connecting a display, moving the cursor to it, removing it, and returning to the first display. |
| P2 Replace the ineffective driver recovery instructions | fix | TROUBLESHOOTING.md tells release users to reinstall a complete OpenBot package and keep their data folder. Development checkouts keep prepare:cua-driver and the panel's Try again action. |

Merged current main (abcfba3f), which includes the browser smoke fix that waits for the native click handler before checking the result.

Checks: computer-use-highlight-window.test.ts (27 passed), cua-driver-runtime.test.ts (37 passed), each with one worker; Biome on the three changed TypeScript files; git diff --check. The commit hook also ran UI foundation and aggregate type checks successfully. This broad hook run was unintended under the current repository instructions; further local checks will remain focused. Physical display connection/removal was not tested.

Surfaces: desktop main process and documentation. No renderer, mobile, public web, Signal, IPC, database, or Team API changes in the fix commit.

Model: GPT-6. Harness: Codex desktop.

/norbiai review

NorbertBodziony:
Round 2: both findings on 93937361630e are valid and fixed in 777ba5f4.

| Finding | Verdict | Action |
| --- | --- | --- |
| P2 Wait for startup before accepting another state probe | fix | start() now awaits #starting before checking running(). Added a deferred socket-readiness test with two state() calls; neither may probe permissions before readiness. |
| P3 Remove the unused driver cursor theme pipeline | fix | Removed theme source/tests/build script, brand export, compiler allowlist and lock entries, all theme build/package/verification steps, artifact resolver, and runtime options. The driver always receives --no-overlay and OpenBot always receives observed pointer positions. |

Focused checks: runtime tests 35, artifact tests 23, lock tests 7; one file at a time and one worker. Changed-file Biome and diff checks passed. Broad checks remain in CI. The controls-only browser smoke scenario passed on macOS with real Electron; Chromium printed temporary cache errors during that run. Physical display connection/removal remains unverified.

Linux CI still fails at the initial browser Save click, including a rerun. Browser implementation and smoke code matched successful main before this round. Added pointer-event diagnostics to the local smoke fixture and failure output to identify whether Linux receives a click at the wrong target or no click. No browser assertion was weakened.

The cleanup also affects all three desktop packaging paths and the internal brand package; it changes no released protocol, persisted data, or renderer trust boundary.

/norbiai review

PR #633, opened 2026-09-20
Support field elicitations and HTTP MCP servers on every provider

NorbertBodziony:
PostHog hung in dev: its MCP elicitations only worked for Computer Use consent, and Codex never received HTTP MCP servers at all.

- Generalize MCP elicitation: empty schema stays a consent question for any server; a schema with properties becomes one question per field with secret redaction, typed values, and decline on skipped required fields.
- Move the shared elicitation helpers into agent/prompts.ts; acp-client.ts reuses them.
- Send HTTP MCP servers to Codex as url plus http_headers (confirmed against the pinned Codex 0.155.0 app-server: invalid credential surfaces as failed with the server 401, so the shape is honored).
- New mcp-provider-shapes.test.ts covers the Codex, Claude and ACP shapes; attention-registry tests cover the PostHog API-key flow.

Verified live in isolated dev: marketplace shows PostHog Installed, MCP Test reports Connected 1 tool, Codex lists flags end to end (0 flags), Opencode Muse Spark 1.3 Contributor completes a flags turn in 32s.

Checks: attention/probe/store suites green, providers suite green (47), lint clean, typecheck clean (pre-commit hooks ran the full parallel typecheck).


NorbertBodziony:
Both findings fixed on head b43903f0, no rebuttals.

1. Stale Codex sessions (P2, mcp-provider-shapes.ts:277): valid, reproduced locally. src/backend/agent/thread-lifecycle.ts now folds CODEX_MCP_ADAPTER_VERSION into toolFingerprint, so pre-HTTP sessions are replaced once through the existing flow with history intact. The replaced session is also released in the client now (it previously leaked its MCP servers on this path). Covered by a new stale-manifest replacement test in agent-service.providers.test.ts.

2. Option titles vs consts (P2, prompts.ts:32): valid, confirmed the card submits option.label (QuestionPromptBubble.tsx:328). elicitationValue now maps a selected oneOf label back to its const, falling back to the raw answer on no match. Covered by a titled-oneOf case in attention-registry.test.ts asserting region eu is sent for European Union.

Checks: 79 tests across the 5 touched-area suites green, lint exit 0, typecheck clean.

/norbiai review

PR #635, opened 2026-09-20
Make MCP work on a clean machine: report every drop, ship Bun, close the second config door, sign in natively

NorbertBodziony:
## What this changes

Four stages of `plans/003-mcp-works-on-a-clean-machine.md`, in order. Each is shippable alone; they
are one PR because stage 2 depends on stage 1's reporting to be debuggable, stage 3 would remove the
last working path without stage 2, and stage 4 is what removes the unpinned third-party bridge that
stage 2 would otherwise make live on many machines.

**1 — Every failure is visible.** `UsableMcpServer`'s error arm carries an `McpDropReason`, and the
Claude, ACP and Codex adapters return `McpHandoff<T>` with the servers they left out instead of a
silent `continue`. `AgentService` reports each new drop once — a `logger.warn` and one
`mcp_server_not_started` error event with no `agentId`, so the renderer shows a single deduped toast
rather than a banner per chat. The set is cleared when the MCP list changes. `resolveMcpCommand` is
now memoized on `PATH\0command`, removing one login shell per stdio server from every thread start;
`testMcpServer` clears that cache, because Test is the user asking about right now.

**2 — Bun 1.4.2 as a managed tool runtime.** A new `MANAGED_TOOL_RUNTIMES` id space and a
`toolRuntimes` field on `ProviderRuntimeSnapshot` (IPC only; no Team API adapter carries it), so
`providers` still means "a provider CLI" for every renderer reader. The runtime manager downloads and
verifies it like a provider CLI. Its bin directory is appended **last** to the lookup and launch
`PATH`, so a user's own Node, nvm or mise build always wins — that is the mitigation for Bun's
compatibility risk, and it makes the change a pure addition. `bun run pin:bun-runtime` refreshes the
pin at release prep.

**3 — The panel is authoritative.** Claude gets `strictMcpConfig: true` (verified against the pinned
`@anthropic-ai/claude-agent-sdk`: it drops project `.mcp.json`, user settings, plugins and agent
frontmatter for MCP only, and `settingSources` is unchanged, so permissions and hooks still load).
The Codex disabled-name sweep moves from profile generation into every thread. Those names join the
manifest fingerprint and `CODEX_MCP_ADAPTER_VERSION` goes to 3, so a session started before this is
replaced once instead of resuming with tools the panel does not show.

**4 — Native MCP OAuth; `mcp-remote` is gone.** `mcp-remote` held six refresh tokens outside
`safeStorage` and outside redaction, which AGENTS.md does not allow. Now: `src/main/mcp-oauth-store.ts`
(versioned zod envelope, cipher injected as two callbacks, no `electron` import), an
`OAuthClientProvider` on the pinned MCP SDK, and a third `openbot://mcp-auth` deep-link variant. The
token is injected at hand-off and **never** written to a row, and it is registered with
`McpHandoffLog`, so a CLI quoting the request it failed on cannot print it. All six bridge listings
are `transport: "http"`; Linear's URL is corrected from `/sse` (404) to `/mcp`.

## Verification

`bun run lint` (0 errors), `bun run typecheck`, `bun run check:ui`, and the test files this touches:
`agent-service.providers.test.ts` (51), `claude-client.test.ts`, `mcp-provider-shapes.test.ts`,
`mcp-oauth-provider.test.ts`, `mcp-oauth-store.test.ts`, `deep-link-router.test.ts`,
`provider-runtime-manager.test.ts`, `build-plugin-catalog.test.ts`. No wider suite was authorized, so
`bun run check` and the packaged-app checks are left to CI.

The OAuth tests run a `node:http` fake that speaks the real wire: RFC 9728 discovery, RFC 7591
registration, a PKCE `S256` authorization code and the bearer header. The redaction test was checked
both ways — with `recordSecret` disabled it fails, because the shared `redactText` masks an
`Authorization: <value>` pattern but not the bare value a CLI quotes.

## Things to know before approving

- **Upgrade effect, deliberate:** a user who configured MCP in `~/.claude/settings.json`, a project
  `.mcp.json` or `~/.codex/config.toml` loses those servers. No opt-out, per the product decision.
  `CHANGELOG.md` says so, and so does a one-time in-app notice.
- **Upgrade effect, deliberate:** the six OAuth listings need one sign-in again after upgrading,
  because their tokens were held by `mcp-remote`. Also in `CHANGELOG.md`.
- **Accepted limit:** all three adapters copy headers once, at spawn. An access token that expires
  mid-thread is not refreshed and the tools go away until the next thread. A bounded session with a
  token OpenBot can redact beats an unbounded one it cannot; this is also the measurement that
  decides whether user servers should later route through the local bridge.
- **Deviations from the plan, both smaller than planned:** the `npx` shim is a hard-linked `bunx`
  plus a lookup-time alias rather than a written shell script, which avoids shipping a generated
  executable; and the tool-runtime status is shown in the MCP panel rather than in onboarding.
- **Not measured:** the Windows `where.exe npx` / `$env:PATHEXT` / `execFileSync` check the plan asks
  for could not be run — this machine is darwin. `pickWindowsExecutable` is implemented with
  all-platform unit tests and falls back to today's behaviour when nothing matches.
- **Known risk:** the Codex `config/read` sweep awaits the provider. A build that never answers would
  stall the thread start there; the request goes through the same client timeout as the rest.
- `normalizeResource` accepts loopback `http://` as well as `https://`, so a user running a local MCP
  server can sign in to it. Nothing else may receive a grant.
- No `biome-ignore`, `@ts-expect-error`, `any` or assertion bypass was added.

## UI

No before/after screenshots: the changed surfaces need a signed-in OAuth server and a failing
hand-off to show anything, and neither can be staged from this shell. What changed visually is the
MCP panel row note for a stdio server with a working directory, the corrected panel sentence (it said
Codex cannot use an HTTP MCP server, untrue since #633), the tool-runtime status, the sign-in dialog,
and the `MCP server not started` toast. Stories are updated for each.

Model: Claude Opus 5. Harness: Claude Code in a t3 worktree.


NorbertBodziony:
All four findings were valid. Fixed in `c9ae604d`; no rebuttals this round.

1. **[P1] Serialize OAuth store changes** — `src/main/mcp-oauth-store.ts`. `write` and `clear` now run through `#enqueue`, a promise chain, so the copy a change starts from is the one the change before it committed, and two refreshes never write the same `.tmp` file at once. A failed change does not fail the change after it. Test: "keeps both tokens when two servers refresh at once" — it fails without the queue.
2. **[P1] Preserve the inherited Windows PATH** — `src/backend/mcp-provider-shapes.ts`. `appendToolRuntimes` falls back to `process.env.PATH` when there is no configured or login-shell `PATH` and there is a managed directory to append. With no managed directory the value stays `null`, so nothing that works today changes. Exported for its own test, like `pickWindowsExecutable`, because the case that matters is the one a Mac cannot reach.
3. **[P1] Pass OAuth authorization to OpenCode** — `src/backend/provider-drivers.ts:220`. `mcpAuthorization: context.mcpAuthorization` added. Test in `acp-client.test.ts` goes through the real driver and asserts the session payload carries `Authorization: Bearer …`; it fails without the line.
4. **[P1] Redact tokens obtained during the probe** — `src/backend/mcp-probe.ts`. `probeSecrets` now adds the sign-in provider's current token to the redaction list, so the token minted between the two attempts is covered as well as `server.authorization`. Test in `mcp-oauth-provider.test.ts` drives a first sign-in against a fake server that answers the authorized request by quoting the token.

Checks on `c9ae604d`: `bun run lint`, `bun run typecheck`, and `test:desktop` on `mcp-oauth-store.test.ts`, `mcp-oauth-provider.test.ts`, `mcp-provider-shapes.test.ts`, `mcp-probe.test.ts`, `acp-client.test.ts` (71 passed).

/norbiai review

NorbertBodziony:
Both P1 findings fixed in `20461ec3`. The P2 is accepted as real and is with the developer.

1. **[P1] Preserve unreadable OAuth records during removal** — `src/main/mcp-oauth-store.ts`. `write` and `clear` now call `#reloadIfUnreadable` first, so a keychain that refused once at startup is asked again at the moment something wants to change the file. A `clear` that still cannot read it returns without touching disk, which is what stops an empty map turning into `rm`. Tests: "keeps an unreadable envelope when a server row is removed" and "reads the file again when the keychain has started answering"; the first fails without the guard. The `write` half is deliberate and already tested by "replaces an unreadable envelope with the sign-in the user just finished": after the re-read fails, the only thing left is a sign-in the user has completed, and refusing to store it would leave them with nothing readable and no way to replace it.
2. **[P1] Share concurrent refreshes for the same resource** — `src/backend/mcp-oauth-provider.ts`. `#refresh` holds one in-flight exchange per normalized resource; later callers await it. The fake authorization server now rotates its refresh token, so the test "refreshes an expiring token once when two hand-offs ask together" fails with a second exchange.
3. **[P2] Convert existing catalog installations to native OAuth** — correct, and not fixed here. Rows installed before this release still hold `npx -y mcp-remote@latest …`, and Stage 2 is what makes them able to run. Rewriting stored user rows is a data change and a product decision, so it is going to the developer rather than being decided in a review round; it will land as its own change with its own test.

Also in this push: `src/preload/provider-runtime.test.ts` now sends `toolRuntimes`, which the snapshot decoder requires since Stage 2. That was a real CI failure, not a review finding.

Checks on `20461ec3`: `bun run lint`, `bun run typecheck`, and `test:desktop` on `mcp-oauth-store.test.ts`, `mcp-oauth-provider.test.ts`, `provider-runtime.test.ts`, `mcp-provider-shapes.test.ts`, `mcp-probe.test.ts`, `acp-client.test.ts`.

/norbiai review

NorbertBodziony:
P1 fixed in 78d9eeeae. No rebuttals this round.

1. **[P1] Preserve unreadable OAuth records during registration** — src/main/mcp-oauth-store.ts. The store now keeps the raw encrypted records from a failed read in #unreadableServers. When the file holds records this build cannot decrypt, write() throws instead of replacing the file, so saveClientInformation cannot delete unrelated credentials when the user starts (or cancels) a sign-in. A missing or corrupt file still takes the replacement path, since it holds nothing worth keeping. The retry in #reloadIfUnreadable still gives a temporarily locked keychain its chance first. Test: refuses a registration while encrypted records cannot be read — it fails without the guard and the file stays byte-identical.

The four P2 findings are real and stay open. Each one needs a decision outside this round:
2. Catalog migration rewrites stored user rows, so it stays with the developer as its own data change.
3. Waiting for the Bun download inside the connection test changes test UX on slow networks; the one-line fire-and-forget start does not satisfy what the finding asks for.
4. Letting the Team API route spend the host stored OAuth token separates browser permission from credential access across a trust boundary.
5. Folding Bun readiness into the Codex tool fingerprint replaces every affected session once Bun lands, and that invalidation needs a design check first.

Checks on 78d9eeeae: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on mcp-oauth-store.test.ts (14 passed) and mcp-oauth-provider.test.ts (8 passed).

/norbiai review

NorbertBodziony:
All four P2 findings fixed in e5c95d5aa. No rebuttals.

1. **[P2] Convert existing catalog installations to HTTP** — src/backend/mcp-server-store.ts, src/backend/agent-service.ts. McpServerStore.migrateCatalogBridgesToHttp converts rows that still match the six shipped mcp-remote definitions (name, stdio, npx, exact bridge args including Linear old /sse URL) to native http rows. Renamed rows and edited arguments stay untouched; id, enabled state, position, and all other columns are kept. It runs in AgentService.initialize, is a data rewrite with no DDL, and reruns convert nothing. The next Test on a converted row signs in through the browser like a new installation. Test: converts catalog bridge rows to native http and leaves edited rows alone.

2. **[P2] Start the runtime download before a plugin connection test** — src/main/provider-runtime-manager.ts, src/main/ipc/mcp-server-handlers.ts, src/main/index.ts. New ensureToolRuntimesReady awaits downloadAndWait for each managed tool runtime. The local stdio test path awaits it before probing; http tests wait for nothing. A failed download still runs the probe, which reports what the machine starts. Test: waits until the tool runtimes are ready.

3. **[P2] Use stored OAuth credentials in remote connection tests** — src/backend/agent-service.ts, src/main/team-api/route-mcp.ts. TestMcpServerOptions gains storedCredentials; the Team API route passes it. The service builds a silent authority from the host McpOAuth whose signIn stays null, so stored tokens are read and refreshed with no browser and no five-minute grant wait. Only toolCount and error travel back. Tests: spends a stored token for a silent test without opening a browser (probe level, expired token refreshes, no new registration); route test asserts storedCredentials: true reaches the service.

4. **[P2] Refresh affected Codex sessions when Bun becomes ready** — src/backend/agent/thread-lifecycle.ts. toolFingerprint folds the managed tool runtimes (bin directories and command aliases) alongside the configured set, and CODEX_MCP_ADAPTER_VERSION rides 3 to 4. A session started before Bun was ready mismatches once its servers resolve, so the next turn replaces it through the existing flow with history intact. The single-reading rule extends to the runtimes. Test: replaces a Codex session started before the tool runtime was ready — fails without the fix, passes with it; the replacement start carries the npx server.

Checks on e5c95d5aa: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on agent-service.providers (52), mcp-server-store (10), mcp-oauth-provider (9), mcp-probe, mcp-provider-shapes, mcp-oauth-store (14), provider-runtime-manager (45), team-api-server.mcp, agent-service.restart — all pass.

/norbiai review

NorbertBodziony:
Three new findings fixed in 16d1993ef. No rebuttals.

1. **[P2] Bound OAuth refresh before thread startup** — src/backend/mcp-oauth-provider.ts. A silent refresh now waits at most MCP_TOKEN_TIMEOUT_MS (10s, injectable as refreshTimeoutMs like signInTimeoutMs); the wait ends and the stored token answers instead. The exchange keeps running, so a token it eventually stores is what the next start reads, and concurrent callers still share one exchange. The token trade after a browser grant gets the same bound and reports The sign-in response did not arrive in time instead of holding the test past its deadline. Tests: answers with the stored token when the refresh hangs (expired token, hanging /token endpoint, returns the stale token, one refresh exchange, no browser — hangs without the fix); gives up the token trade when the authorization server hangs (full sign-in, error names the timeout). The fake server records /token arrivals before hanging and destroys held sockets on close, so the hang stays inside the test.

2. **[P2] Preserve the managed PATH when launching a server** — src/backend/mcp-provider-shapes.ts. mcpLaunchEnvironment applies the resolved server.path after the configured pairs. The lookup and the launch now agree: the configured PATH with the managed directories appended. User directories still come first; a configured PATH still stands when the lookup answered nothing. Tests: launches with the PATH the command was looked up in (fails without the fix); keeps a configured PATH when the lookup answered nothing.

3. **[P2] Match the text rendered by the MCP story** — src/renderer/stories/ServerMcp.stories.tsx. Both play queries now read /Claude only/, matching mcpProviderLimitNote.

Checks on 16d1993ef: bun run lint (no findings in touched files), bun run typecheck (all projects pass), bun run check:ui (pass), test:desktop on mcp-oauth-provider (11), mcp-provider-shapes, mcp-probe, mcp-server-store, agent-service.providers (52), provider-runtime-manager, team-api-server.mcp, mcp-oauth-store, agent-service.restart — 185 passed across 9 files.

/norbiai review

NorbertBodziony:
Two new findings fixed in e2487319f. No rebuttals.

1. **[P2] Do not block installed commands on a Bun download** — src/backend/mcp-provider-shapes.ts, src/main/ipc/mcp-server-handlers.ts, src/main/index.ts. New needsManagedRuntime resolves the configured command against the current runtimes first: http and already-resolvable stdio commands (installed interpreters, absolute paths, the user's own npx) probe at once. Only a command nothing names waits for the download, and that wait is bounded at sixty seconds — a stalled transfer keeps downloading for the next attempt while the probe reports what this machine starts. The handler reads the current runtimes through a new toolRuntimes dependency. Tests: ignores an http server, ignores a command this machine already resolves, waits for a command nothing names, skips the wait once the managed store names the command.

2. **[P2] Prevent pending OAuth exchanges from restoring deleted credentials** — src/backend/mcp-oauth-provider.ts. forget counts a generation per server before clearing storage, and every client provider built from then on carries the count it saw. A refresh or sign-in already running when the row is removed finishes with a refused write (The MCP sign-in was forgotten while it was running) instead of restoring the account, so removal reads complete and re-adding the URL cannot reuse it. Active flows are unaffected: clear still passes through and same-generation writes land. Test: does not restore credentials forgotten during a refresh — removal lands mid-exchange, the caller gets the stale token, storage stays empty; fails without the guard.

Checks on e2487319f: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on mcp-oauth-provider (12), mcp-provider-shapes, mcp-probe, mcp-server-store, agent-service.providers (52), agent-service.restart, provider-runtime-manager, team-api-server.mcp, mcp-oauth-store — 190 passed across 9 files.

/norbiai review

NorbertBodziony:
Three findings fixed in b8bb7a6f. No rebuttals.

1. **[P2] Refresh loaded sessions when Bun becomes ready** — src/backend/agent-service.ts, src/main/application-services.ts. New public AgentService.refreshAllAgentRuntimes passes through to the threads' deferred refresh. The tool runtime ready event subscribes in the composition root and marks every agent when a managed tool runtime becomes ready; provider CLI updates change no MCP resolution, so only tool runtimes refresh. Mid-turn sessions keep the mark until the turn ends, and public threads and histories stay. The Codex fingerprint from the last round stays as the per-turn check. Test: starts a fresh provider session after the tool runtimes become ready.

2. **[P2] Compare normalized URLs before deleting shared credentials** — src/backend/agent-service.ts. removeMcpServer forgets only when no remaining http row normalizes to the same resource via normalizeResource, the same function the store keys by. https://mcp.stripe.com and https://mcp.stripe.com/ share one credential until the last row goes; an un-normalizable URL still forgets nothing it could not have stored. Test: keeps the shared sign-in until the last row on its URL is removed — fails without the fix.

3. **[P2] Catch runtime download failures before awaiting the race** — src/main/ipc/mcp-server-handlers.ts. The download rejection is now caught before Promise.race, so failure and timeout both continue to the probe, which reports what this machine starts. The registrar dependencies narrowed to the members it reaches (McpServerService, McpRemoteServers), so the new handler test builds plain fakes with no assertions. Tests in new mcp-server-handlers.test.ts drive the real binder through a mocked Electron registration: installed command probes without waiting, http never waits, missing command waits before probing, failed download still probes.

Checks on b8bb7a6f: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on mcp-server-handlers (4), agent-service.providers (54), agent-service.restart, mcp-oauth-provider (12), mcp-provider-shapes, team-api-server.mcp — all pass.

/norbiai review

NorbertBodziony:
New finding fixed in 90d93110. No rebuttals.

**[P2] Start host runtime downloads for remote MCP requests** — src/main/ipc/mcp-server-handlers.ts, src/main/team-api/route-mcp.ts, dependencies.ts, team-api-server.ts, host-service.ts, application-services.ts. The IPC handlers' preparation travels as one shared shape, McpToolRuntimePreparation, from the composition root through the host into the Team API options. The save and enable routes fire and forget like their local branches; the test route shares prepareToolRuntimeForTest, so a remote test of a command nothing names waits bounded before probing, and installed commands probe at once. Without the preparation a host with no Node could neither save nor test its first server remotely. Test: starts the runtime download behind the remote save, enable, and test routes — fails without the route change; the existing local handler tests cover the shared wait.

Checks on 90d93110: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on team-api-server.mcp, mcp-server-handlers (4), agent-service.providers (54), agent-service.restart, mcp-oauth-provider (12), mcp-provider-shapes — all pass.

/norbiai review

NorbertBodziony:
New finding fixed in 69c56b43. No rebuttals.

**[P2] Align the remote Test timeout with runtime preparation** — src/main/ipc/mcp-server-handlers.ts. The remote Test branch now sends timeoutMs covering the bounded runtime wait plus the probe plus margin (60s + 10s + 20s), instead of falling back to the 15-second request limit. Without it a first download past 15 seconds made the caller report a timeout before the host probed anything. The registrar dependencies also narrowed to the members they reach, so the handler test builds plain fakes with no assertions. Test: gives a remote test a deadline that covers the host preparation — fails without the fix; the remote branch still skips local preparation and the local probe.

Checks on 69c56b43: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on mcp-server-handlers (5), team-api-server.mcp — all pass.

/norbiai review

NorbertBodziony:
Three findings fixed in f08795fd. No rebuttals.

1. **[P1] Restrict authorization URLs before opening them** — src/backend/mcp-oauth-provider.ts. redirectToAuthorization now requires https, with http allowed only for loopback, before openExternal. The address arrives in the server's own discovery document and the SDK accepts more than web pages, so a metadata document naming a file or an installed protocol handler no longer opens it. Test: refuses to open a sign-in page that is not on the web — without the fix the file: URL reaches the browser and the test hangs on the grant; with it nothing opens and the error names the refusal.

2. **[P2] Preserve discovery state for token exchange and refresh** — src/backend/mcp-oauth-provider.ts. The client provider implements the SDK discovery-state members per resource, held in memory beside the records: a refresh or later sign-in reuses the authorization server a previous exchange found instead of rediscovering at the default locations and falling back to the MCP origin's token endpoint. Invalidation clears the retained state with the credentials it belongs to. Test: reuses the authorization server the sign-in discovered — metadata only at the advertised URL, dead default server; the silent refresh after the sign-in succeeds only with the retained state.

3. **[P2] Prevent abandoned sign-ins from opening the browser** — src/backend/mcp-oauth-provider.ts. abandon marks the run; a discovery slow enough to outlast the probe finds no browser to open when it finishes, and complete on an abandoned run fails fast instead of waiting out a grant nobody answers. Test: never opens a browser for a sign-in the probe abandoned.

Checks on f08795fd: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on mcp-oauth-provider (15), mcp-provider-shapes, mcp-probe, mcp-server-store, agent-service.providers (54), agent-service.restart, provider-runtime-manager, team-api-server.mcp, mcp-oauth-store, mcp-server-handlers — 201 passed across 10 files.

/norbiai review

NorbertBodziony:
Two findings fixed in 4d477fc4. No rebuttals.

1. **[P2] Preserve discovery state for token exchange and refresh** — src/backend/mcp-oauth-provider.ts. The retained state now persists with the encrypted credentials: the record carries the authorization server URL and the metadata URL that named it, and the SDK re-fetches the documents themselves from those addresses. A refresh or sign-in after a restart reuses them instead of rediscovering at the default locations. Invalidation clears the retained state with the credentials it belongs to. Test: recovers the authorization server after a restart — sign-in discovers through the advertised URL, a new McpOAuth instance over the same store refreshes through it; fails without the fix.

2. **[P2] Release a stalled refresh before allowing another attempt** — src/backend/mcp-oauth-provider.ts. The exchange carries an abort tied to the same deadline as the wait: a request that stays open without completing is cancelled and its entry cleared, so the next caller starts a fresh exchange instead of joining the same stall and receiving the expired token again. Cancellation travels through the SDK fetch hook; the waiter bound from the last round stays. Test: starts a fresh exchange when the running refresh stalls — two sequential callers produce two token requests; without the fix the second joins the first.

Checks on 4d477fc4: bun run lint (no findings in touched files), bun run typecheck (all projects pass), test:desktop on mcp-oauth-provider (17), mcp-provider-shapes, mcp-probe, mcp-server-store, agent-service.providers (54), agent-service.restart, provider-runtime-manager, team-api-server.mcp, mcp-oauth-store, mcp-server-handlers — 203 passed across 10 files.

/norbiai review

NorbertBodziony:
Both P1 findings were valid. Fixed in `3ca57aa4`.

**1. [P1] Reject insecure discovered token endpoints** — `src/backend/mcp-oauth-provider.ts:240`

Confirmed: `normalizeResource` covers only the MCP server's own URL and `isAuthorizationUrlSafe` covers only the browser address, so neither reached the credential-carrying requests. Both `fetchFn` hooks handed the discovered URL straight to `fetch`.

`secureOAuthFetch` is now the single fetch for the refresh and the code exchange. It permits https or loopback http only, and follows redirects by hand — `redirect: "manual"`, up to 5 hops, each hop checked — because `fetch` would otherwise follow a redirect to a plain-text endpoint without saying where it went. 307/308 keep the method and body; 301/302/303 become a bodyless GET, per RFC 9110.

**2. [P1] Redact all OAuth secrets from probe failures** — `src/backend/mcp-probe.ts:123`

Confirmed, including the mechanism: `parseErrorResponse` in `node_modules/@modelcontextprotocol/sdk/dist/esm/client/auth.js:132` makes `error_description` the message of the thrown error, which `describeMcpError` returns to the renderer. `probeSecrets` collected only access tokens, and read them from the store *after* the failure — so a recoverable refusal that cleared the record dropped even the access token it was meant to mask.

`McpOAuthClientProvider` now keeps a ledger of every value the attempt handles as it handles it: access and refresh tokens, client secret, authorization code, and the PKCE verifier. `McpSignIn.secrets()` exposes it and `probeSecrets` uses it instead of re-reading the store.

**Tests** — two added in `src/backend/mcp-oauth-provider.test.ts`:
- a token endpoint advertised as `http://auth.example.com/token` is refused, and `tokenRequests` stays empty, so the grant left the machine nowhere;
- a refusal quoting the code and verifier back in `error_description` reaches the panel with both masked.

Verified both fail without the fix. Reverting the two call sites and the ledger fails 3 tests — the new pair plus the existing "keeps the token it just minted out of the failure it reports", which the ledger now also covers.

**Checks on `3ca57aa4`:** `bun run test:desktop -- src/backend/mcp-oauth-provider.test.ts src/backend/mcp-probe.test.ts src/main/mcp-oauth-store.test.ts` (49 passed), `bun run lint` (no errors), `bun run typecheck` (exit 0). No renderer change, so `check:ui` was not needed — the pre-commit hook ran it anyway and it passed.

/norbiai review

NorbertBodziony:
All three findings were valid. Fixed in `76153229`.

**1. [REMAINS][P1] Reject insecure discovered token endpoints** — `src/backend/mcp-probe.ts:199`

You are right and my previous round was incomplete: the first review asked for the guard on "refresh, code exchange, and probe authentication" and I covered only the first two. `createTransport` passed `authProvider` with no `fetch`, and `streamableHttp.js:43` calls `auth(..., { fetchFn: this._fetchWithInit })`, which defaults to global fetch — so the 401-triggered refresh posted the refresh token and client secret unguarded.

`createTransport` now passes `fetch: secureOAuthFetch()` alongside `authProvider`. This narrows nothing reachable: a provider is only attached when `oauth.signIn(config.url)` returned non-null, which requires `normalizeResource` to have already accepted the URL as https or loopback.

My earlier test for this did not actually cover the path — it asserted on the error text, which the *interactive* fallback produced, so it passed with the guard removed. Replaced with one that records what left the machine via a `globalThis.fetch` spy and asserts nothing was sent to the insecure endpoint. Without the fix it now fails with `expected [ 'http://auth.example.com/token' ] to deeply equal []` — which also confirms the leak was real.

**2. [NEW][P1] Remove credentials on cross-origin redirects** — `src/backend/mcp-oauth-provider.ts:563`

Valid, and a regression I introduced: following redirects by hand took over the stripping `fetch` was doing for me. Confirmed the premise — `auth.js:44` defaults to `client_secret_basic` when the metadata omits `token_endpoint_auth_methods_supported`.

The loop now rejects a cross-origin hop when the request carries a credential — an `Authorization` header or a body — which covers the 302→GET case, since the header survives that conversion. Discovery GETs carry neither, so an issuer may still redirect to the authorization server that answers for it.

**3. [NEW][P2] Guard credential deletion from stale authentication attempts** — `src/backend/mcp-oauth-provider.ts:305`

Valid. `clear` had neither the generation nor the abandonment check that `write` had. Both checks are now in one `ensureCurrent()` used by `write` and `clear`.

**Tests** — three added in `src/backend/mcp-oauth-provider.test.ts`, each verified to fail when its own fix is reverted (3 failed / 21 passed):
- a rejected-but-unexpired stored token, with the token endpoint advertised as plain http — nothing reaches it;
- a token endpoint that 307-redirects to another origin — the second origin receives no request at all;
- `invalidateCredentials("all")` from an abandoned attempt — rejected, and the newer record survives.

**Checks on `76153229`:** `bun run test:desktop` on `mcp-oauth-provider`, `mcp-probe`, `mcp-oauth-store` and `mcp-server-handlers` (57 passed), `bun run lint` (exit 0), `bun run typecheck` (exit 0), `bun run check:ui` (passed, via the pre-commit hook).

Note this branch also merged `origin/main` in `9e6e7f51`; the only conflict was an import union in `src/main/application-services.ts`.

/norbiai review

PR #637, opened 2026-09-20
Add permanent invitation links

NorbertBodziony:
Reusable, never-expiring invite links on both the local team host and D1-backed account hosts.

**Product**
- New **Perma link** tab in Server settings → Members, beside Email and Invite link.
- Permanent links never expire and accept unlimited joins until revoked. Link-only (no email binding), revocable from the pending list, capped at 5 per host (`maxPermanentInvites`); single-use budgets unchanged.
- Join dialog shows Expires: Never, including for links previewed over transports whose frozen codecs strip the flag (sentinel-deadline fallback).

**Compatibility (non-negotiables held)**
- Released Team API v1–v4 adapters untouched: frozen HTTP projections strip the new fields on the wire, so old clients/servers keep working (old host + new client degrades to single-use display; join still works).
- D1 migration `0020` is additive only (`max_uses DEFAULT 1`, `use_count DEFAULT 0`); pre-migration rows read as single-use and old-Worker INSERTs keep working through the deploy gap.
- Local team file gains optional fields only; old files read as single-use.
- No `biome-ignore`, no new channels, no dependency or license changes.

**UI before / after**
- Before: Email and Invite link tabs; single-use copy (24h expiry).
- After: third Perma link tab with full-width hint, no success banner after creating (link + Copy/QR stay), pending rows read "Permanent invitation link · Never expires · N joins".
- Model: opencode-go/muse-spark-1.3-contributor. Harness: Storybook Members story plus live Electron dev-app smoke via `dev:automation` (created a real perma link in the isolated dev profile). Screenshots omitted from the repo per UI PR rules.

**Checks run**
- `dev:verify --run` (desktop suites incl. new store/control-plane/modal tests, auth-api control-plane + retention, lint, typecheck, check:ui) — green.
- Live smoke: created perma + single-use links in the isolated dev app; both create correctly. Pending list stays empty there for both kinds due to a pre-existing dev-profile routing quirk (local create vs account list), unrelated to this change.
- Full `bun run check` left to CI.


NorbertBodziony:
Review fixes (both P2s fixed, no rebuttals):

1. **Preserve permanent-link requests across HTTP** (`src/main/team-api/route-team.ts`) — fixed by gating instead of extending the wire. Frozen adapters stay untouched per `packages/contracts/AGENTS.md`: the Perma tab now hides on legacy HTTP remotes (local IPC and account-plane hosts report no HTTP origin, so `permanentSupported` stays true there), and `RemoteTeamDirectory.createInvite` throws `This server connection does not support permanent invitation links.` on the HTTP branch rather than silently minting single-use. Covered by a modal tab-visibility test and a new `remote-team-directory.test.ts`.

2. **Enforce the permanent-link cap atomically** (`remote-control-plane.ts`) — the separate count query is gone for the permanent path; creation is now one conditional `INSERT ... SELECT ... WHERE (SELECT COUNT ...) < ?` that returns `invite_limit_reached` when no row is inserted. Existing cap tests exercise the zero-changes path.

Checks: targeted desktop suites (directory, modal, store), auth-api control-plane suite, full `typecheck`, `check:ui`, and `lint` on touched files — green (remaining repo lint errors are pre-existing in untouched files).

/norbiai review

NorbertBodziony:
Fixed the remaining P3: removed the duplicated single-use outstanding-invites count (one awaited D1 query per creation instead of two). The surviving check sits beside the single-use insert, after the permanent branch returns. Control-plane suite (12 tests) and Biome on the file are green.

/norbiai review

PR #655, opened 2026-09-21
Log in to ChatGPT with a code typed on another device

### LINKED ISSUE #648: Add code-based login support for GPT
## Problem

Users need an alternative way to authenticate their GPT account when the standard browser-based login flow is unavailable, inconvenient, or fails inside the desktop app.

## Proposed solution

Add a code-based GPT login flow:

- Provide a clear `Log in with code` option alongside the existing GPT login flow.
- Generate a short-lived, one-time authentication code and show the verification URL/instructions.
- Allow the user to complete authentication on another device or browser and then return to OpenBot.
- Poll or receive the completed authentication state without exposing tokens or sensitive credentials in the UI.
- Show clear states for waiting, success, expiration, cancellation, network errors, and rejected/invalid codes.
- Persist the authenticated session consistently with the existing GPT login flow.
- Ensure logout, account switching, reconnects, and app restarts handle the code-based session correctly.

## Acceptance criteria

- A user can authenticate GPT using the code flow without relying on an embedded browser.
- The code is single-use, expires safely, and cannot be reused after completion or expiration.
- The UI provides actionable feedback throughout the flow and supports retry/cancel.
- Tokens and sensitive authentication data are never displayed or logged.
- Existing GPT login remains unchanged.
- Automated tests cover success, invalid/expired code, cancellation, timeout, retry, logout, and account switching scenarios.

NorbertBodziony:
Closes #648.

The browser sign-in hands the account off to whatever browser this computer opens. That has nowhere to finish on a machine with no browser, in a remote session, or in a browser already signed in to the wrong account. Every ChatGPT row now offers a second way in, next to the existing one: OpenBot asks Codex for a short-lived device code, shows it with the page to type it on, and waits for the provider to report the result.

The existing browser login is unchanged.

## How it works

- `account/login/start` with `{ type: "chatgptDeviceCode" }` returns a `userCode` and a `verificationUrl`. The code is a one-time handle that expires; the token traded for it stays in the main process. Nothing about the token, and no other sensitive field, crosses into the renderer, so there is nothing here to redact. What the renderer gets is a phase, the code, the URL, and the expiry.
- Cancel calls `account/login/cancel`, so a dropped dialog does not leave a live code behind.
- The end of the login is read from the provider's own status row - the same signal the browser login reports through - so there is no second poller and no new completion channel.

## What the user sees

The dialog shows the QR code for the login page, the page address, the code in a mono face large enough to read off the screen onto a phone, and a countdown. It covers the waiting only.

**Before** (first cut, in this branch): the dialog ended on a last screen inside itself - a green "connected", an "expired", or a refusal - with `Done` / `Get a new code` / `Try again` buttons in a footer.

**After**: the dialog closes and the ending arrives as a toast. By then the user has been away on another device; what they come back to should be an app they can use, not a modal to dismiss first. The retry rides on the toast ("Get a new code", "Try again"), so "the code expired" is not a dead end.

## Two fixes worth calling out

- **The menu appears on connected rows too.** Asking for a code while signed in is how a user reaches a second account. The account in use keeps working throughout, and survives a code nobody types.
- **Completion is `connectionState` leaving `connecting`, not `state === "available"`.** A provider the user is already signed in to is `available` from the start. Reading that as the finish reported success the moment the code appeared: nobody asking for a second account ever got to type one. Covered by a renderer test that fails without the guard.

## Surfaces

IPC contracts (`packages/contracts`) and the preview mock, main-process handler, backend provider runtime, desktop renderer (settings and onboarding), i18n strings, Storybook. No migration, no schema change, no wire-protocol change. The reverse action (cancel) is implemented and tested.

Also included: a contrast fix for `.ui-alert[data-tone] .ui-alert-description` in `primitives.css`. It came out of this work but is app-wide and independent of the dialog.

## Tests

- `provider-runtime.test.ts` - start, cancel, and the connecting state around a code login.
- `App.test.tsx` - a code sign-in on a provider that is *already* signed in stays waiting until the row stops connecting, then closes and toasts the new account.
- `ProviderPicker.test.tsx`, `SettingsModal.test.tsx`, `OnboardingFlow.test.tsx` - the entry points, including on a connected row.

Run locally: those files plus `bun run lint`, `bun run typecheck`, `bun run check:ui` - all green. Wider suites left to CI.

Verified live against a real connected ChatGPT account twice: the menu appears on the connected row, a genuine code is issued at `https://auth.openai.com/codex/device` with a 10-minute countdown, and the account in use is untouched.

Model: Claude Opus 5. Harness: Claude Code (T3 Code).


NorbertBodziony:
Fixed both P2 findings in c60cffd1:

- Failed account switches report success: the code login completion handler now checks the provider error message before it reports success. A failed switch keeps the error and the retry action.
- Ignore replies from cancelled login attempts: each attempt has a generation. Cancellation, completion, and cleanup invalidate it. Replies, errors, and expiry timers check it. A retry waits for pending cancellation, and stale cancellation replies cannot replace the current status.

Validation: all 35 tests in App.test.tsx pass, including failed account switches and late success/error replies after cancellation. Full lint, typecheck, and check:ui pass (lint retains 160 existing warnings). In an isolated local dev app, I opened code login, closed it, retried, and saw the new code with the dialog still open. Completing an account switch requires user authentication and was not tested live.

No rebuttals.

/norbiai review

PR #667, opened 2026-09-21
Sign in to Canva with a loopback redirect address

### LINKED ISSUE #665: Fix Canva plugin authorization failing with Invalid redirect URI
## Problem

The Canva plugin authorization flow fails during OAuth with the error:

`Invalid redirect URI.`

The failure occurs on the Canva authorization page at `https://mcp.canva.com/authorize`, preventing the plugin from completing setup and connecting to Canva.

## Steps to reproduce

1. Start the Canva plugin authorization flow in OpenBot.
2. Continue to the Canva OAuth authorization page.
3. Observe the `Invalid redirect URI.` error.

## Actual behavior

Canva rejects the authorization request and the plugin cannot be connected.

## Expected behavior

The plugin should send a valid, registered redirect URI and complete the Canva authorization flow successfully. If configuration is invalid, OpenBot should show a clear actionable error identifying the missing or mismatched redirect URI configuration.

## Acceptance criteria

- Verify the redirect URI generated by OpenBot for the Canva plugin.
- Ensure the URI exactly matches the URI registered for the Canva OAuth/MCP client in every supported environment.
- Complete the authorization flow successfully after fixing the configuration.
- Add validation and diagnostics for missing, malformed, or mismatched redirect URIs.
- Add regression coverage for the Canva plugin authorization flow.
- Document any required redirect URI configuration for local and packaged builds.

## Reference

The screenshot below shows Canva returning `Invalid redirect URI.`.

Canva authorization error

NorbertBodziony:
Fixes #665.

## What was wrong

OpenBot registered and sent `openbot://mcp-auth` as the OAuth redirect address for every http MCP server. Canva accepts that address at `/register` and then refuses it at `/authorize` with `Invalid redirect URI.`, so the sign-in ends on Canva's own page, where OpenBot sees nothing and can explain nothing.

Checked against the live servers:

| Server | `openbot://mcp-auth` at `/authorize` | loopback at `/authorize` |
| --- | --- | --- |
| Canva | `400 Invalid redirect URI.` | `302` to the login page |
| Linear, Notion, Sentry, Stripe | accepted | accepted |

## What changed

- `src/main/mcp-oauth-redirect-server.ts` (new): a listener on `127.0.0.1`, on the port the operating system gives. It answers `GET /mcp-auth` only; it refuses a request whose `Host` header is not loopback, which is what a page that points a name it owns at this port sends; it refuses a `state` no sign-in is waiting for, exactly as the deep link does; and the page it serves never repeats the grant.
- `src/main/application-services.ts`: starts the listener, gives its address to `McpOAuth`, shows the main window when a grant lands, and stops the listener at teardown. A port that cannot be bound is logged and the `openbot://mcp-auth` deep link is used instead, so nothing that works today stops working.
- `src/backend/mcp-oauth-provider.ts`: `describeUnusableRedirectUrl` refuses a redirect address a grant cannot come back to (an `https` address, or plain text that is not on this machine) where the address is configured. `clientInformation()` answers `undefined` one time per interactive sign-in when the stored registration names another address, so a registration written by an older build is replaced instead of earning the same refusal. A silent refresh keeps the `client_id` its refresh token was issued to.

No configuration is needed for a local or a packaged build: the port changes on each start, an authorization server must ignore the port of a loopback redirect, and the registration is dynamic. `docs/main-process-design-notes.md` states this.

## Surfaces

Main process, backend OAuth, and one renderer comment. No IPC channel, database, or Team API change.

## Verification

- Live flow against `https://mcp.canva.com/mcp` with this code: the registration records the loopback address and `/authorize` answers `302` to Canva's login instead of `400`.
- `bun run test:desktop -- src/main/mcp-oauth-redirect-server.test.ts` - 6 passed (new file).
- `bun run test:desktop -- src/backend/mcp-oauth-provider.test.ts` - 28 passed, 4 new: the address registered, sent and repeated at the token endpoint; a stored registration that names another address replaced exactly once; a silent refresh that does not register; and an unusable address refused.
- `bun run test:desktop -- src/main/deep-link-router.test.ts` - 18 passed.
- Biome on the changed files, and the pre-commit type checks, are clean. Broad checks are left to CI.


NorbertBodziony:
## Fixes in `2481047f`

**[P2] Preserve the client associated with stored refresh tokens** - `src/backend/mcp-oauth-provider.ts:422` - valid, fixed.

The mechanism holds: `authInternal` reads `clientInformation()` before it tries the refresh, so an interactive attempt replaced the registration and then spent the stored refresh token against a `client_id` it was never issued to. Because the loopback port changes on each start, a stored registration is stale after every restart, so this would have cost a browser sign-in whenever a plain refresh would have worked.

The registration is now replaced only for the pass that needs the browser:

- A record that still has a refresh token keeps its client for one pass. A refresh sends no redirect address, so a stale one does not matter there.
- When that refresh is refused, `auth()` drops the tokens and re-enters `authInternal`. The record then has no refresh token, so that pass registers again and the authorization request carries this run's address.
- The exchange of a grant already in hand keeps the client too: `beginCodeExchange()` is called before `auth(provider, { authorizationCode })`, so a code is never traded against a client it was not issued to.
- A silent refresh is unchanged: it never has a `state`, so it always uses the client on file.

New regression test `refreshes against the stored registration before it registers again` in `src/backend/mcp-oauth-provider.test.ts`: a stored registration naming `openbot://mcp-auth`, an access token the server refuses, and a valid refresh token. The probe now ends with `registrations === 0`, no browser opened, and the refreshed token stored. Before this commit it registered once and opened a browser.

`docs/main-process-design-notes.md` states the rule.

Checks on `2481047f`: `bun run test:desktop -- src/backend/mcp-oauth-provider.test.ts` (29 passed), `src/main/mcp-oauth-redirect-server.test.ts` (6 passed), `src/main/deep-link-router.test.ts` (18 passed), Biome on the changed files, and the pre-commit type checks. No rebuttals this round.

/norbiai review

